PR #33, opened 2023-09-05
Bugfix/issue 31 full build failed

qijun-niu-timeplus:
PR checklist:
- Did you run ClangFormat ?


yokofly (approved):
some internal discussed info:
1. remove `dumpArray` in `src/AggregateFunctions/tests/gtest_uda.cpp`
The helper function in gtest is not been used. current encounter a build err in parse 
```
std::cout << col.getData()[i] << ",";
```
2. why switch StringRef to std::string_view
```
StringRef query_id;
if (query_id.size)  --> Here the size is a data member but will be considered as `size()` member function. thus compare a pointer/addr in if condition.
{...}
```

PR #79, opened 2023-09-12
fix: GET /streams on non-exist and empty databases.

### LINKED ISSUE #56: list streams endpoint returns 500 when there is no stream
**Describe what's wrong**

Before 1.3.4, list streams endpoint returns `{"data":[],"request_id":"XX"}` with status code `200` when there is no streams. However after 1.3.4, it would  return `{"code":36,"error_msg":"Code: 36. DB::Exception: database 'default' doesn't exit or does not have any streams. (BAD_ARGUMENTS) (version 1.3.4)","request_id":""}` with status code `500`. This is a breaking change to Neutron and more over, this doesn't follow rest api pattern.

Bonus: for get or delele stream/view endpoint, proton returns 500 with error code when stream/view doesn't exist. It should either return `404` or just `204`.

**How to reproduce**

curl "http://localhost:3218/proton/v1/ddl/streams"

**Additional context**

https://timeplus-workspace.slack.com/archives/C02LT9BMZ2R/p1694033367039409

zliang-min:
PR checklist:
- Did you run ClangFormat ?

Now, GET /streams:
* returns 404 if the target databse does not exist
* returns 404 if the target stream does not exist
* return 200 with empty array for `data` if the target database has no streams

closes #56


sunset3000 (approved):

jovezhong:

PR #136, opened 2023-10-03
add car demo

gangtao:
This PR add car demo which generate simulate car data in realtime to proton docker-compose. 
refer to https://github.com/timeplus-io/chameleon/pull/50

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -41,6 +41,22 @@ services:
     depends_on:
       - redpanda
 
+  cardemo:
+    image: timeplus/cardemo:latest
+    entrypoint: /bin/sh
+    command: -c "sleep 15 && echo \"$$CONSOLE_CONFIG_FILE\" > /timeplus/sink.yaml;  /timeplus/cardemo --config /timeplus/.cardemo.yaml -f /timeplus/sink.yaml"
+    environment:
+      CONSOLE_CONFIG_FILE: |
+        sinks:
+          - type: proton
+            properties:
+              interval: 200
+              host: proton
+    networks:
+      - demo_network
+    depends_on:
+      - proton
+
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -8,6 +8,9 @@ volumes:
 services:
   proton:
      image: ghcr.io/timeplus-io/proton:latest
+     ports:
+      - "3218:3218" # HTTP
```
jovezhong, line 12: these 2 ports are not required to get data from owl-shop or cardemo. There are only needed for go/python drivers. So I don't think we need to declare them.

docker-compose.yml
```diff
@@ -41,6 +44,22 @@ services:
     depends_on:
       - redpanda
 
+  cardemo:
```
jovezhong, line 49: the current docker-compose file will use Redpanda and related UI/datagen. The cardemo is independent and I think we should create a new docker-compose file in a new folder

jovezhong (changes requested):

docker-compose.yml
gangtao: it is not required for internal consumption, I added them just for clarity, in case user want to ingest or run any python client access

gangtao: I did not find anything from the readme to declare these port, it will be helpful to provide such sample

docker-compose.yml
gangtao: we can add profile to match different demo scenario, https://docs.docker.com/compose/profiles/

jovezhong:
Please review the new PR https://github.com/timeplus-io/proton/pull/156

jovezhong:

jovezhong:

PR #137, opened 2023-10-05
Feature/issue 67 uda support changelog

### LINKED ISSUE #67: UDA should support rows update in version_kv
**Use case**

```sql
drop stream if exists test14_multishard;
create stream if not exists test14_multishard(i float, k1 int, k2 string) primary key k2 settings mode='versioned_kv';
select sec_large(i) from test14_multishard;
insert into test14_multishard(i, k1, k2) values (6, 1, 'c');
insert into test14_multishard(i, k1, k2) values (5, 2, 'a');
insert into test14_multishard(i, k1, k2) values (4, 3, 'b');
insert into test14_multishard(i, k1, k2) values (2, 1, 'a');
insert into test14_multishard(i, k1, k2) values (3, 2, 'b');
insert into test14_multishard(i, k1, k2) values (4, 3, 'c');
```
```
Received exception from server (version 1.3.8):
Code: 48. DB::Exception: Received from localhost:8463. DB::Exception: Method negate is not supported for sec_large: While executing GlobalAggregatingTransform. (NOT_IMPLEMENTED)
```

**Additional context**

sunset3000:
PR checklist:
- Did you run ClangFormat ? Yes
- Did you separate headers to a different section in existing community code base ? Yes
- Did you surround `proton: starts/ends` for new code in existing community code base ? Yes

UDA support 'retract' on changelog stream. close #67

### DIFF
--- a/src/AggregateFunctions/AggregateFunctionJavaScriptAdapter.cpp
+++ b/src/AggregateFunctions/AggregateFunctionJavaScriptAdapter.cpp
@@ -215,24 +215,26 @@ JavaScriptAggrFunctionState::~JavaScriptAggrFunctionState()
 
 void JavaScriptAggrFunctionState::add(const IColumn ** src_columns, size_t row_num)
 {
+    assert(columns.size() >= 1);
     size_t num_of_input_columns = columns.size() - 1;
 
     for (size_t i = 0; i < num_of_input_columns; i++)
         columns[i]->insertFrom(*src_columns[i], row_num);
 
     /// _tp_delta column
-    columns[columns.size() - 1]->insert(1);
+    columns.back()->insert(1);
 }
 
 void JavaScriptAggrFunctionState::negate(const IColumn ** src_columns, size_t row_num)
 {
+    assert(columns.size() >= 1);
     size_t num_of_input_columns = columns.size() - 1;
 
     for (size_t i = 0; i < num_of_input_columns; i++)
         columns[i]->insertFrom(*src_columns[i], row_num);
 
     /// _tp_delta column
-    columns[num_of_input_columns]->insert(-1);
+    columns.back()->insert(-1);
 }
 
 void JavaScriptAggrFunctionState::reinitCache()
--- a/src/AggregateFunctions/AggregateFunctionJavaScriptAdapter.cpp
+++ b/src/AggregateFunctions/AggregateFunctionJavaScriptAdapter.cpp
@@ -86,10 +86,22 @@ JavaScriptBlueprint::JavaScriptBlueprint(const String & name, const String & sou
 
         {
             v8::Local<v8::Value> val;
-            if (!obj->Get(local_ctx, V8::to_v8(isolate_, "has_customized_emit")).ToLocal(&val) || !val->IsUndefined())
+            if (obj->Get(local_ctx, V8::to_v8(isolate_, "has_customized_emit")).ToLocal(&val) && !val->IsUndefined())
             {
-                LOG_INFO(&Poco::Logger::get("JavaScriptAggregateFunction"), "JavaScript UDA '{}' has defined its own emit strategy", name);
-                has_user_defined_emit_strategy = true;
+                has_user_defined_emit_strategy = V8::from_v8<bool>(isolate_, val);
+                if (has_user_defined_emit_strategy)
+                    LOG_INFO(
+                        &Poco::Logger::get("JavaScriptAggregateFunction"), "JavaScript UDA '{}' has defined its own emit strategy", name);
+            }
+        }
+
+        {
+            v8::Local<v8::Value> val;
+            if (obj->Get(local_ctx, V8::to_v8(isolate_, "support_changelog")).ToLocal(&val) && !val->IsUndefined())
+            {
+                support_changelog = V8::from_v8<bool>(isolate_, val);
+                if (support_changelog)
+                    LOG_INFO(&Poco::Logger::get("JavaScriptAggregateFunction"), "JavaScript UDA '{}' can work on changelog stream", name);
             }
         }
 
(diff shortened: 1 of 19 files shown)


src/Interpreters/Streaming/SubstituteStreamingFunction.cpp
```diff
@@ -68,6 +69,51 @@ std::unordered_map<String, String> StreamingFunctionData::changelog_func_map = {
...
+    /// `count_if`              => `__count_retract_if`
+    /// `count_distinct`        => `__count_retract_distinct`
+    /// `count_distinct_if`     => `__count_retract_distinct_if`
+    String combinator_suffix;
+    auto nested_func_name = function_name;
+    while (iter == changelog_func_map.end())
+    {
+        if (auto combinator = AggregateFunctionCombinatorFactory::instance().tryFindSuffix(nested_func_name))
+        {
+            const std::string & combinator_name = combinator->getName();
```
qijun-niu-timeplus: Since `getName` return a String value, there is no need to use a string reference here.

qijun-niu-timeplus (approved):

src/V8/Utils.h
```diff
@@ -64,7 +64,7 @@ inline void run(
     func(isolate, local_ctx, try_catch);
 }
 
-/// Validate UDA
```
chenziliang, line 67: Lets remove this line of comment since it doesn't give more information.

src/Interpreters/Streaming/SubstituteStreamingFunction.h
```diff
@@ -28,6 +28,11 @@ struct StreamingFunctionData
 
     /// only streaming query can use these functions
     static std::set<String> streaming_only_func;
+
+    /// check whether or not the function support 'retract' for changelog:
+    /// the 1st return value: indicates whether or not it support retract
+    /// the 2nd return value: the alias name of function should be rewritten in the query, empty if not support changelog
+    std::pair<bool, String> supportChangelog(const String & function_name);
```
chenziliang: `supportsChangelog(...)`

chenziliang: I think a `std::optional<String> supportsChange(...)` signature works better ? If the target func supports retract, a non-empty std::optional is returned and the the function alias name which shall be rewritten. Otherwise an empty std::optional is returned

src/AggregateFunctions/AggregateFunctionJavaScriptAdapter.cpp
```diff
@@ -206,11 +215,24 @@ JavaScriptAggrFunctionState::~JavaScriptAggrFunctionState()
...
-        col->insertFrom(*src_columns[i], row_num);
-        i++;
-    }
+    size_t num_of_input_columns = columns.size() - 1;
+
+    for (size_t i = 0; i < num_of_input_columns; i++)
+        columns[i]->insertFrom(*src_columns[i], row_num);
+
+    /// _tp_delta column
+    columns[columns.size() - 1]->insert(1);
```
chenziliang: Are we always expecting `_tp_delta` column even for non-changelog scenario ?

```
columns.back()->insert(1);
```

src/AggregateFunctions/AggregateFunctionJavaScriptAdapter.cpp
```diff
@@ -206,11 +215,24 @@ JavaScriptAggrFunctionState::~JavaScriptAggrFunctionState()
...
+
+void JavaScriptAggrFunctionState::negate(const IColumn ** src_columns, size_t row_num)
+{
+    size_t num_of_input_columns = columns.size() - 1;
+
+    for (size_t i = 0; i < num_of_input_columns; i++)
+        columns[i]->insertFrom(*src_columns[i], row_num);
+
+    /// _tp_delta column
+    columns[num_of_input_columns]->insert(-1);
```
chenziliang: `columns.back()->insert(-1)`

src/AggregateFunctions/AggregateFunctionJavaScriptAdapter.cpp
```diff
@@ -206,11 +215,24 @@ JavaScriptAggrFunctionState::~JavaScriptAggrFunctionState()
 
 void JavaScriptAggrFunctionState::add(const IColumn ** src_columns, size_t row_num)
 {
-    for (size_t i = 0; auto & col : columns)
-    {
-        col->insertFrom(*src_columns[i], row_num);
-        i++;
-    }
+    size_t num_of_input_columns = columns.size() - 1;
```
chenziliang, line 223: It appears `assert(columns.size() >= 1);` is a good idea?

src/Functions/UserDefined/UserDefinedFunctionFactory.cpp
```diff
@@ -207,8 +212,36 @@ bool UserDefinedFunctionFactory::registerFunction(
...
             V8::validateAggregationFunctionSource(function_name, {"initialize", "process", "finalize"}, config->get("source"));
+
+            /// add _tp_delta column as the last argument
+            assert(config->has("arguments"));
+            assert(config->isArray("arguments"));
+
+            bool has_delta_column = false;
+            /// Below is a workaround, because Poco::JSON::Object::getArray cannot work well
+            Poco::JSON::Array json_arguments = config->get("arguments").extract<Poco::JSON::Array>();
+            for (unsigned int i = 0; i < json_arguments.size(); i++)
```
chenziliang: `size_t i = 0;`

src/Functions/UserDefined/UserDefinedFunctionFactory.cpp
```diff
@@ -207,8 +212,36 @@ bool UserDefinedFunctionFactory::registerFunction(
...
+            {
+                const auto & arg = json_arguments.getObject(i);
+                if (arg->has("name") && arg->get("name") == "_tp_delta")
+                {
+                    has_delta_column = true;
+                    break;
+                }
+            }
+
+            if (!has_delta_column)
```
chenziliang, line 240: I don't quite understand why we need `_tp_delta` for all cases.

src/AggregateFunctions/AggregateFunctionJavaScriptAdapter.cpp
sunset3000: Yes, because in the UDA, _tp_delta should always be the last input argument.

src/Functions/UserDefined/UserDefinedFunctionFactory.cpp
sunset3000: Because when writing source of UDA, the user never knows whether the UDA is used on a changelog stream or not,   he has to assume 'process' function always cope with _tp_delta column.  And when register UDA with REST api, he does not need to add _tp_delta as an input argument, here proton automatically adds '_tp_delta' column as the last arguement.

chenziliang: Framework (proton) doesn't need to pass the delta column to UDA when in non-changelog scenario since this is more performant. The problem for now is in changelog scenarios, proton passes the delta column to UDA, can UDA differentiate these two cases from its inputs ?

qijun-niu-timeplus (approved):

PR #154, opened 2023-10-10
submodule change: introduce google bench

### LINKED ISSUE #147: introduce google benchmark
**Use case**
discussed internally:
Currently, we use benchmarks in unit tests, switching to Google benchmark will be more straightforward. and even easily generate graphs with different input scales.

https://github.com/timeplus-io/proton/blob/1b75d3d3f1110a39174d6f761ccbfbd56e54f2a8/src/Common/tests/gtest_threading.cpp#L389-L408

**Additional context**

yokofly:
_update v1: option title`ENABLE_BENCHMARK` -> `ENABLE_BENCHMARKS`_
_update v2: post clickhouse community pr for tracking_

disable by default, use option `-DENABLE_BENCHMARKS=ON/OFF` 

fix #147 
upstream pr https://github.com/ClickHouse/ClickHouse/pull/43779


src/Common/examples/CMakeLists.txt
```diff
@@ -85,3 +85,8 @@ endif()
 
 proton_add_executable (interval_tree interval_tree.cpp)
 target_link_libraries (interval_tree PRIVATE dbms)
+
+if (ENABLE_BENCHMARK)
+    proton_add_executable (benchmark_string_copy benchmark_string_copy.cpp)
+    target_link_libraries (benchmark_string_copy PRIVATE dbms)
```
yokofly: todo:
1. move to src/Common/benchmarks folder
2. link necessary lib instead whole dbms archive

build.sh
```diff
@@ -44,6 +44,7 @@ cmake .. \
     -DENABLE_CLANG_TIDY=OFF \
     -DENABLE_TESTS=ON \
     -DENABLE_EXAMPLES=ON \
+    -DENABLE_BENCHMARKS=ON \
```
qijun-niu-timeplus: I don't think we should recommand others to compiler benchmarks by default.
Could you set it `OFF` here?

yokofly: this is `build.sh` /

yokofly: above DENABLE_EXAMPLES and DENABLE_TESTS all turn to ON, so I do the same

yokofly: my point is:
1. this file is not obviously recommended to build/use right?
2. the example has been set on, it's straightforward i also build the benchmark right? any special reasons to set off?
even when we build the program, those are all executable binary, we still need manual `./benchmark_xxx` to call it.

yokofly: discussed internally:
> The ENABLE_EXAMPLES && ENABLE_TESTS options are useful for development,
but ENABLE_BENCHMARKS is optional.
Having it on by default enlarges the link time for **dbms**,
so disabling it is recommended.

qijun-niu-timeplus (approved):

PR #180, opened 2023-10-18
feat: using kafka external streams as sinks.

### LINKED ISSUE #81: Write to kafka external streams
**Use case**

Currently it's supported to create kafka external streams and read data directly from them. However, writing to such streams is not supported. It will be a very useful feature to send data to external streams so that users can easily use proton for real-time streaming processing, and send the data to other systems for different usage.

Proton already supports specifying a target stream for materialized views, so we can use the same syntax for writing to external streams:

```sql
CREATE EXTERNAL STREAM some_stream ( ... );

CREATE MATERIALIZED VIEW some_view INTO some_stream AS ... ;
```

**Additional context**

zliang-min:
Supports the following use cases:
* `INSERT INTO <kafka-external-stream> (<cols...>) VALUES (<values...>)`
* `INSERT INTO <kafka-external-stream> SELECT <cols...> FROM <other-stream>`
* `CREATE MATERIALIZED VIEW <view-name> INTO <kafka-external-stream> AS ...`

Messages will be encoded with the format sepcified with the `data_format` setting of the external stream. All output formats defined in `Formats/registerFormats.cpp` should be supported. However, not all supported formats are tested.

A new setting `properties` is added for creating kafka external streams. This allows users to customize the kafka client used by the kafka external stream without us adding more settings in the future. Not all properties are supported, currently supported properties are:
* enable.idempotence
* message.timeout.ms
* queue.buffering.max.messages
* queue.buffering.max.kbytes
* queue.buffering.max.ms
* message.send.max.retries
* retries
* retry.backoff.ms
* retry.backoff.max.ms
* batch.num.messages
* batch.size
* compression.codec
* compression.type
* compression.level

Note that, this `properties` setting is only used when the external stream is used as a sink. We can add support for source later. Also, currently `properties` is validated on runtime, we could improve it to validate it when the stream is being created.

This initial implementation does not support specifying message key.

Currently there is no good way to add tests for this feature yet. I will firstly utilize our integration test to add some tests from there. Later on we can enhance the test framework to support testing this feature.

closes #81 

PR checklist:
- Did you run ClangFormat ?


src/Storages/Streaming/StorageMaterializedView.cpp
```diff
@@ -28,6 +28,7 @@
 #include <Common/ProtonCommon.h>
 #include <Common/checkStackSize.h>
 #include <Common/setThreadName.h>
+#include "Storages/ExternalStream/StorageExternalStream.h"
```
yl-lisen: should use `<>`

src/Storages/Streaming/StorageMaterializedView.cpp
```diff
@@ -200,9 +201,13 @@ StorageMaterializedView::StorageMaterializedView(
                 throw Exception(ErrorCodes::INCORRECT_QUERY, "Target stream is not found", target_table_id.getFullTableName());
 
             auto * stream = target_table->as<StorageStream>();
-            if (stream == nullptr)
-                throw Exception(
-                    ErrorCodes::NOT_IMPLEMENTED, "MaterializedView doesn't support target storage is {}", target_table->getName());
+            if (!stream)
+            {
+                auto * external_stream = target_table->as<StorageExternalStream>();
+                if (!external_stream)
```
yl-lisen: we can simplify as
```c++
if (!target_table->as<StorageStream>() && !target_table->as<StorageExternalStream>())
    throw ...
```

src/Storages/ExternalStream/Log/FileLog.cpp
```diff
@@ -112,13 +112,12 @@ Pipe FileLog::read(
...
-        max_block_size,
-        saved_start_timestamp,
-        searchForCandidates(),
-        log));
+        this, std::move(header), std::move(context), max_block_size, saved_start_timestamp, searchForCandidates(), log));
+}
+
+SinkToStoragePtr FileLog::write(const ASTPtr & /*query*/, const StorageMetadataPtr & /*metadata_snapshot*/, ContextPtr /*context*/)
+{
+    throw Exception(ErrorCodes::NOT_IMPLEMENTED, "Ingesting data to external stream of 'log' type is not supported");
```
yl-lisen: A better way is that move it into `StorageExternalStreamImpl::write`
If no override, throw an error `NOT_IMPLEMENTED`

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -30,19 +31,23 @@ class Kafka final : public StorageExternalStreamImpl
...
+    SinkToStoragePtr write(const ASTPtr & query, const StorageMetadataPtr & metadata_snapshot, ContextPtr context) override;
+
     const String & brokers() const { return settings->brokers.value; }
     const String & dataFormat() const { return data_format; }
     const String & dataSchema() const { return settings->data_schema.value; }
     const String & topic() const { return settings->topic.value; }
     const String & securityProtocol() const { return settings->security_protocol.value; }
     const String & username() const { return settings->username.value; }
     const String & password() const { return settings->password.value; }
+    std::vector<std::pair<std::string, std::string>> properties() const { return parseProperties(settings->properties.value); }
```
yl-lisen: We can do parse properties ahead in ctor once and store as `properties`.
This interface just return it directly.

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -20,6 +27,64 @@ extern const int OK;
...
+        }));
+    };
+
+    static auto rtrim = [](std::string &s)
+    {
+        s.erase(std::find_if(s.rbegin(), s.rend(), [](unsigned char ch) {
+            return !std::isspace(ch);
+        }).base(), s.end());
+    };
+
```
yl-lisen: you can use `boost::trim_left` and `boost::trim_right` to instead of them

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -20,6 +27,64 @@ extern const int OK;
...
+    /// message.max.bytes=1024;max.in.flight=1000;group.id=my-group
+
+    std::vector<std::string> parts;
+    boost::split(parts, properties, boost::is_any_of(";"));
+    result.reserve(parts.size());
+
+    std::vector<std::string> kv{2};
+    for (const auto& part : parts)
+    {
+        boost::split(kv, part, boost::is_any_of("="));
```
yl-lisen: It seems just split a string to pair<string, string>, use `substr` may be easier

src/KafkaLog/KafkaWALPool.cpp
```diff
@@ -1,4 +1,5 @@
 #include "KafkaWALPool.h"
+#include <vector>
```
yl-lisen, line 2: The standard header file should always place at backend.
```c++
/// 1. include self header files

/// 2. include project's other header files

/// 3. include standard header files
```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include "Formats/FormatFactory.h"
+#include "Interpreters/Context.h"
+#include "KafkaLog/KafkaWALCommon.h"
+#include "KafkaSink.h"
+#include "Processors/Chunk.h"
+#include "Processors/Formats/IOutputFormat.h"
+#include "Processors/ProcessorID.h"
+#include "Storages/ExternalStream/ExternalStreamTypes.h"
+#include "Storages/ExternalStream/Kafka/Kafka.h"
+#include "Storages/ExternalStream/Kafka/WriteBufferFromKafka.h"
```
yl-lisen: please use `<>` instead of `""`

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+namespace ErrorCodes
+{
+extern const int MISSING_ACKNOWLEDGEMENT;
+extern const int INVALID_CONFIG_PARAMETER;
+}
+
+KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr context, const Poco::Logger * logger)
+    : SinkToStorage(header, ProcessorID::ExternalTableDataSinkID), log(logger), producer(nullptr, nullptr)
+{
+    is_streaming = true;
```
yl-lisen: It may not be a streaming ingestion.
In fact, we shall assign it automatically later during pipeline building or connecting (except Source needed), so you should remove this line

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+    if (data_format.empty())
+        data_format = "JSONEachRow";
+
+    writer = FormatFactory::instance().getOutputFormat(data_format, *wb, header, context);
+    writer->setAutoFlush();
+
+    pollingThread = std::thread([this]() {
+        while (!isFinished)
+        {
+            auto n = rd_kafka_poll(producer.get(), 500 /* ms */);
```
yl-lisen: the poll interval should be a configuration value or a const variable to avoid hard code

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -0,0 +1,48 @@
...
+private:
+    std::function<void(rd_kafka_t *, const rd_kafka_message_t * msg, void *)> onDrMsg();
+    std::function<void(rd_kafka_t *, void *, size_t, rd_kafka_resp_err_t, void *, void *)> onDr();
+
+    const Poco::Logger * log;
+    RdKafkaPtr producer;
+    std::unique_ptr<WriteBufferFromKafka> wb;
+    OutputFormatPtr writer;
+    std::thread pollingThread;
+    bool isFinished;
```
yl-lisen: Have to use an atomic_bool `is_finished` ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+    });
+}
+
+void KafkaSink::consume(Chunk chunk)
+{
+    if (!chunk.hasRows())
+        return;
+
+    auto blk = Block(getHeader());
+    blk.setColumns(chunk.getColumns());
```
yl-lisen: `auto block = getHeader().cloneWithColumns(chunk.detachColumns());`

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+{
+    onFinish();
+}
+
+void KafkaSink::checkpoint(CheckpointContextPtr context)
+{
+    do
+    {
+        if (auto err = wb->deliveryError(); err != RD_KAFKA_RESP_ERR_NO_ERROR)
+            throw Exception(std::string("Failed to send messages: ") + rd_kafka_err2str(err), klog::mapErrorCode(err));
```
yl-lisen: May you modify to 
```c++
throw Exception(klog::mapErrorCode(err), "Failed to send messages: {}", rd_kafka_err2str(err));
```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+        if (wb->fullyAcked())
+            break;
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
yl-lisen: Let's use notify to wake instead of sleep

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+size_t WriteBufferFromKafka::outstandings() const
+{
+    return state.outstandings.load();
+}
+size_t WriteBufferFromKafka::acked() const
+{
+    return state.acked.load();
+}
```
yl-lisen: need `ALWAYS_INLINE `

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -0,0 +1,80 @@
...
+        throw Exception(
+            "Cannot write to kafka topic at offset " + std::to_string(count()) + ", error: " + rd_kafka_err2str(rd_kafka_last_error()),
+            ErrorCodes::CANNOT_WRITE_TO_KAFKA);
+
+    ++state.outstandings;
+}
+
+void WriteBufferFromKafka::finalizeImpl()
+{
+    LOG_INFO(log, "finalizeImpl is called, offset={}", offset());
```
yl-lisen:
```c++
if (log)
    LOG_INFO(log, "finalizeImpl is called, offset={}", offset());
```

src/Storages/ExternalStream/Kafka/KafkaSink.h
zliang-min: There is only one place (the `onFinish` function) updates` `is_finished`, so I didn't use `atomic_bool` in this case. Do you have concerns on this?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Do we only use `<>`?

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
zliang-min: Why `if (log)` is needed in this case?

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -20,6 +27,64 @@ extern const int OK;
...
+
+    /// properties example:
+    /// message.max.bytes=1024;max.in.flight=1000;group.id=my-group
+
+    std::vector<std::string> parts;
+    boost::split(parts, properties, boost::is_any_of(";"));
+    result.reserve(parts.size());
+
+    std::vector<std::string> kv{2};
+    for (const auto& part : parts)
```
chenziliang: clang-format the code

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -20,6 +27,64 @@ extern const int OK;
 extern const int RESOURCE_NOT_FOUND;
 }
 
+std::vector<std::pair<std::string, std::string>> Kafka::parseProperties(std::string properties)
```
chenziliang: Let's refactor out this util and unit test it.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
chenziliang: Yes, moving forward only `<>`, unless unit test etc which is not in the include path and it is local. Please check the proton eng guildline

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,45 @@
+#pragma once
+
+#include <atomic>
+#include <librdkafka/rdkafka.h>
```
chenziliang: It will be good to have a single place to depend on librdkafka headers. So i am expecting most of the Kafka utils / functions will go to `KafkaLog` folder and others depend on `KafkaLog`

src/Storages/IStorage.h
```diff
@@ -221,6 +221,11 @@ class IStorage : public std::enable_shared_from_this<IStorage>, public TypePromo
         /// proton: FIXME, introduce versioning schema physically in metastore
         return metadata.get();
     }
+
+    /// Insert queries squash blocks for buffering (see InterpreterInsertQuery.cpp).
+    /// However, not all storages need this feature, for example, external streams likely have
+    /// their own ways to do buffering.
+    virtual bool insertWithoutSquash() const { return false; }
```
chenziliang: it seems better flip the **name logic** here since it will be more readable/

```
virtual bool squashInsert() const noexcept { return true; }
```

src/KafkaLog/KafkaWALPool.cpp
chenziliang: And do we need include vector here since most of the time it is already included in its dependency headers ?

src/Storages/ExternalStream/ExternalStreamSettings.h
```diff
@@ -15,6 +15,7 @@ class ASTStorage;
     M(String, security_protocol, "plaintext", "The protocol to connection external logstore", 0) \
     M(String, username, "", "The username of external logstore", 0) \
     M(String, password, "", "The password of external logstore", 0) \
+    M(String, properties, "", "A semi-colon-separated key-value pairs for configuring the kafka client used by the external stream. A key-value pair is separated by a equal sign. Example: 'client.id=my-client-id;group.id=my-group-id'. Note, not all properties are supported, please check the document for supported properties.", 0) \
```
chenziliang, line 18: +1

src/Storages/ExternalStream/Kafka/Kafka.cpp
chenziliang: It appears to me returning a `std::unordered_map<std::string, std::string>` could be cleaner ?

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,45 @@
...
+    };
+
+public:
+    WriteBufferFromKafka(const Poco::Logger * logger, RdKafkaTopicPtr topic_, size_t buffer_size = DBMS_DEFAULT_BUFFER_SIZE);
+
+    ~WriteBufferFromKafka() override = default;
+
+    void onDrMsg(const rd_kafka_message_t *);
+    bool fullyAcked() const;
+    rd_kafka_resp_err_t deliveryError() const;
```
chenziliang: We shall translate `rd_kafka_resp_err_t` to ErrorCodes to avoid the very low level error propagate the modular boundry and we have some util in `KafkaLog` already.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include "IO/BufferWithOwnMemory.h"
+#include "IO/WriteBuffer.h"
+
+namespace DB
+{
+
+using RdKafkaTopicPtr = std::unique_ptr<rd_kafka_topic_t, void (*)(rd_kafka_topic_t *)>;
+
+class WriteBufferFromKafka : public BufferWithOwnMemory<WriteBuffer>
```
chenziliang: If we don't want others to extend, it is is always a good idea to seal the class with `final`

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,45 @@
...
+    ~WriteBufferFromKafka() override = default;
+
+    void onDrMsg(const rd_kafka_message_t *);
+    bool fullyAcked() const;
+    rd_kafka_resp_err_t deliveryError() const;
+    size_t outstandings() const;
+    size_t acked() const;
+    void resetState();
+
+protected:
```
chenziliang: `private:`

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,45 @@
...
+    void nextImpl() override;
+    void finalizeImpl() override;
+
+private:
+    const Poco::Logger * log;
```
chenziliang: I usually put `log` as the last data members and `non-const`

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,45 @@
...
+private:
+    struct State
+    {
+        std::atomic_size_t outstandings;
+        std::atomic_size_t acked;
+        std::atomic_int errorCode;
+    };
+
+public:
+    WriteBufferFromKafka(const Poco::Logger * logger, RdKafkaTopicPtr topic_, size_t buffer_size = DBMS_DEFAULT_BUFFER_SIZE);
```
chenziliang: I usually move logger as the last parameter of a function / ctor.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,45 @@
...
+#include "IO/WriteBuffer.h"
+
+namespace DB
+{
+
+using RdKafkaTopicPtr = std::unique_ptr<rd_kafka_topic_t, void (*)(rd_kafka_topic_t *)>;
+
+class WriteBufferFromKafka : public BufferWithOwnMemory<WriteBuffer>
+{
+private:
```
chenziliang: private section gets moved to last section of a class

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,45 @@
...
+    struct State
+    {
+        std::atomic_size_t outstandings;
```
chenziliang: let's be explicit the width of a variable if it is need serde (serialized / deserialized) like `std::atomic_uint64_t or std::atomic<uint64_t>` etc

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,45 @@
...
+        std::atomic_size_t acked;
+        std::atomic_int errorCode;
```
chenziliang: `error_code`

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,45 @@
...
+    };
+
+public:
+    WriteBufferFromKafka(const Poco::Logger * logger, RdKafkaTopicPtr topic_, size_t buffer_size = DBMS_DEFAULT_BUFFER_SIZE);
+
+    ~WriteBufferFromKafka() override = default;
+
+    void onDrMsg(const rd_kafka_message_t *);
```
chenziliang: Let's be a bit more verbose like `onDeliveryMessage` since C++ style is different than Go.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,45 @@
...
+    bool fullyAcked() const;
```
chenziliang: `allAcked()` is a bit better ?

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -0,0 +1,80 @@
...
+    return state.outstandings.load();
+}
+size_t WriteBufferFromKafka::acked() const
+{
+    return state.acked.load();
+}
+
+void WriteBufferFromKafka::resetState()
+{
+    state.outstandings.store(0);
```
chenziliang: Give `State` a `reset()` interface and move the logic there, like

```
struct State
{
    void reset() 
    {
        ...
    }

...
};
```

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+rd_kafka_resp_err_t WriteBufferFromKafka::deliveryError() const
+{
+    return static_cast<rd_kafka_resp_err_t>(state.errorCode.load());
+}
+
+size_t WriteBufferFromKafka::outstandings() const
+{
+    return state.outstandings.load();
+}
```
chenziliang: empty line

chenziliang: This one line functions can be implemented in header file directly

chenziliang: and just return `state.outstandings;` will do the work instead of calling `load() / store()` unless you need control the consistency level. By default it is sequence consistency.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
chenziliang: It appears a debug level log instead of info  since it will be too verbose.

chenziliang: Ok, it is finalization, then we shall log info and let's log more information regarding topic name etc

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -0,0 +1,80 @@
...
+        RD_KAFKA_MSG_F_COPY | RD_KAFKA_MSG_F_BLOCK,
+        working_buffer.begin(),
+        offset(),
+        nullptr /* key */,
+        0 /* keylen */,
+        static_cast<void *>(this) /* opaque */);
+
+    if (unlikely(err))
+        throw Exception(
+            "Cannot write to kafka topic at offset " + std::to_string(count()) + ", error: " + rd_kafka_err2str(rd_kafka_last_error()),
```
chenziliang, line 35:
```
throw Exception(
    ErrorCodes::CANNOT_WRITE_TO_KAFKA, 
     "Failed to write to Kafka topic {} at offset {}, error='{}'",
     kafka_topic_name,
     count(),
     rd_kafka_err2str(rd_kafka_last_error());
```

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -0,0 +1,80 @@
...
+    : BufferWithOwnMemory<WriteBuffer>(buffer_size), log(logger), topic(std::move(topic_))
+{
+}
+
+void WriteBufferFromKafka::nextImpl()
+{
+    if (!offset())
+        return;
+
+    auto err = rd_kafka_produce(
```
chenziliang, line 23: Do we need a plain produce here ? Anything we can leverage in KafkaLog or since we only have sink here a plain Kafka produce is more lightweight / straight forward ? Asking this since we use functions in KafkaLog in other places instead of using the plain produce / consume from librdkafka.

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -0,0 +1,48 @@
...
+
+private:
+    std::function<void(rd_kafka_t *, const rd_kafka_message_t * msg, void *)> onDrMsg();
+    std::function<void(rd_kafka_t *, void *, size_t, rd_kafka_resp_err_t, void *, void *)> onDr();
+
+    const Poco::Logger * log;
+    RdKafkaPtr producer;
+    std::unique_ptr<WriteBufferFromKafka> wb;
+    OutputFormatPtr writer;
+    std::thread pollingThread;
```
chenziliang: `polling_thread` and can we use ThreadPool util instead of using our own thread since that one is audited.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -0,0 +1,80 @@
+#include "WriteBufferFromKafka.h"
+#include <cmath>
```
chenziliang: `cmath` for ?

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <utility>
```
chenziliang: `<utility>` for ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+extern const int INVALID_CONFIG_PARAMETER;
+}
+
+KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr context, const Poco::Logger * logger)
+    : SinkToStorage(header, ProcessorID::ExternalTableDataSinkID), log(logger), producer(nullptr, nullptr)
+{
+    is_streaming = true;
+
+    /// default values
+    std::vector<std::pair<std::string, std::string>> producer_params{
```
chenziliang: I feel like there are quite a few boilplate code which are dup with `KafkaLog`, can we try to modulize / extend KafkaLog to see if we can have a centralized place to maintain all of this ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+    /// customization, overrides default values
+    for (const auto & pair : kafka->properties())
+    {
+        if (allowed_properties.contains(pair.first))
+        {
+            producer_params.emplace_back(pair.first, pair.second);
+            continue;
+        }
+        throw Exception("Unsupported property: " + pair.first, ErrorCodes::INVALID_CONFIG_PARAMETER);
```
chenziliang: Let's stick to this format when throw

```
throw Exception(error_code, "..... format string....",  ... format args ...); 
```

`format string` is something using the format of `std::format` or `fmt::format`

chenziliang:
```
throw Exception(ErrorCodes::INVALID_CONFIG_PARAMETER, "Unsupported property: {}={}" + pair.first, p.second);
```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+        "retry.backoff.max.ms",
+        "batch.num.messages",
+        "batch.size",
+        "compression.codec",
+        "compression.type",
+        "compression.level",
+    };
+
+    /// customization, overrides default values
+    for (const auto & pair : kafka->properties())
```
chenziliang, line 44:
```
for (const auto & [key, value] : kafka->properties())
{
    ...
}
```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
chenziliang: We shall detach columns from chunk and move the columns over to blk.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+        return;
+
+    auto blk = Block(getHeader());
+    blk.setColumns(chunk.getColumns());
+    writer->write(blk);
+}
+
+void KafkaSink::onFinish()
+{
+    isFinished = true;
```
chenziliang: Do we need guard onFinish to make sure it is called only once ?

chenziliang: std::atomic_flag will do the work

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+    if (pollingThread.joinable())
+        pollingThread.join();
+
+    // Make sure all outstanding requests are transmitted and handled.
+    if (auto err = rd_kafka_flush(producer.get(), -1 /* blocks */); err)
+        throw Exception(std::string("Failed to flush kafka producer: ") + rd_kafka_err2str(err), klog::mapErrorCode(err));
```
chenziliang: It is a bad idea to throw in `dtor` which caused program to abort. Just log everything since fundamentally we can't do anything about the errors at this point.

src/Storages/ExternalStream/Kafka/Kafka.cpp
chenziliang: Most of the time, we have some utils there already. If we don't know, we can ask around in Slack.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+    for (const auto & pair : kafka->properties())
+    {
+        if (allowed_properties.contains(pair.first))
+        {
+            producer_params.emplace_back(pair.first, pair.second);
+            continue;
+        }
+        throw Exception("Unsupported property: " + pair.first, ErrorCodes::INVALID_CONFIG_PARAMETER);
+    }
+
```
chenziliang, line 53: Let's remove this redundant empty line

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+    // Make sure all outstanding requests are transmitted and handled.
+    if (auto err = rd_kafka_flush(producer.get(), -1 /* blocks */); err)
+        throw Exception(std::string("Failed to flush kafka producer: ") + rd_kafka_err2str(err), klog::mapErrorCode(err));
+
+    if (auto err = wb->deliveryError(); err != RD_KAFKA_RESP_ERR_NO_ERROR)
+        throw Exception(std::string("Failed to send messages: ") + rd_kafka_err2str(err), klog::mapErrorCode(err));
+
+    // if flush does not return an error, the delivery report queue should be empty
+    if (!wb->fullyAcked())
```
chenziliang: We probably need poll a bit here to wait for all outstanding batches are committed ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
chenziliang: sleep maybe is ok for now

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+    do
+    {
+        if (auto err = wb->deliveryError(); err != RD_KAFKA_RESP_ERR_NO_ERROR)
+            throw Exception(std::string("Failed to send messages: ") + rd_kafka_err2str(err), klog::mapErrorCode(err));
+
+        if (wb->fullyAcked())
+            break;
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    } while (true);
```
chenziliang: We may need check the shutdown / finished since it will be dead loop if there is something wrong in external Kafka, and we can't tear down.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+void WriteBufferFromKafka::finalizeImpl()
+{
+    LOG_INFO(log, "finalizeImpl is called, offset={}", offset());
+}
+
+void WriteBufferFromKafka::onDrMsg(const rd_kafka_message_t * msg)
+{
+    if (msg->err)
+        state.errorCode.store(msg->err);
```
chenziliang: We probably need an other state which is `state.errors`, `errors + acked` will be the total messages sent. We will need honor errors in different places as well.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Actually, using notify in this case will not only increase the complexity but also cause unneccessary cost. Because if we want to use notify, the background polling thread has to add the following logic:
* does `aked` equal to `outstandings`?
* if it does, notify
* if it does not do nothing
The problem here is that, when it notifies, there could be no one waiting for the notification at all. Because the only place needs the notification is the `checkpoint` function, if checkpointing is not in-progress, it's pointless to notify nobody ( i.e. checking if `aked == outstandings` every poll is meaningless ).

But please let me know if you have different ideas.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: Right, my first thought was to use `unordered_map` too. But the only use case for the `properties` is to loop over it, no lookup or update is needed. Thus `vector` is more efficient here. Besides, I think `vector<pair<string, string>>` is also quite readable, it's just the `std::` prefix makes it horrable (probably should add `using` to remove that).

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
zliang-min: I can't use `ErrorCodes` here because I need to use `rd_kafka_err2str` to get the error message from the error code.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
zliang-min: `State` just needs to stay in memory and no serde is required ( e.g. it's not a checkpoint state ).

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
zliang-min: Because it's used in `rd_kafka_dr_msg_cb` function call, I named it `onDrMsg` to match it (it has nothing to do with Go). But sure, we can make it more verbose.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
zliang-min: Correct me if I got this wrong, based on reading the code in `KafkaLog`, the existing implementation can't not be used in this sink directly, I will need to implement something like a `KafkaWALProducer` (just like the `KafkaWALConsumer`).

Secondly, IMO, all the `KafkaSink` and `WriteBufferForKafka` classes are already a high level concept, it's totally fine to use the librdkafka library directly in these classes (these classes are already wrappers). Adding another layer looks will unneccessarily complex the code (as you can see, the logic in `KafkaSink` and `WriteBufferForKafka` are very simple and straightfoward).

So, I don't really feel adding another layer helps in this case.

However, I do think that `KafkaSource` and `KafkaSink` should be more consistent. Probably it makes sense to do some refactor with followup PRs.

I hope this makes sense.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Yup, things can be modulized are:
* configuration handling
* error handling

I already tried use existing code in `klog` where it's possible, but there could be some improvement we can do. I think we can make such changes later when I do more work on the source side so that I will get better ideas what can be done.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: No, that's not neccessary. `rd_kafka_flush(.., -1)` will make sure *ALL* outstandings get handled. If at this point it's still not fully acked, something wrong must have happened.

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -166,4 +235,9 @@ void Kafka::validate()
 
     shards = result.partitions;
 }
+
+SinkToStoragePtr Kafka::write(const ASTPtr & /*query*/, const StorageMetadataPtr & metadata_snapshot, ContextPtr context)
+{
+    return std::make_shared<KafkaSink>(this, metadata_snapshot->getSampleBlock(), context, log);
```
yl-lisen, line 241: For multiple partitions, may need multiple `KafkaSink` likes `read method`.
We also need to consider how to partition

src/Storages/ExternalStream/Kafka/Kafka.cpp
chenziliang: `using std` or other namespace is not a good practice, always include namespace and you will find it is way more maintainable. I think std::unordered_map has benefits like dedup etc.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr context, const Poco::Logger * logger)
+    : SinkToStorage(header, ProcessorID::ExternalTableDataSinkID), log(logger), producer(nullptr, nullptr)
+{
+    is_streaming = true;
+
+    /// default values
+    std::vector<std::pair<std::string, std::string>> producer_params{
+        std::make_pair("enable.idempotence", "true"),
+        std::make_pair("message.timeout.ms", "0" /* infinite */),
```
chenziliang: you probably can try

```
std::vector<std::pair<std::string, std::string>> producer_params{
    {"enable.idempotence", "true"},
    {"message.timeout.ms", "0" /* infinite */},
};
```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
chenziliang: and in C++, let's be verbose a bit regarding variable name, like blk vs block

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
chenziliang: i don't recall if `rd_kafka_flush(...)` waits, does the `-1` mean wait indefinitely ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
chenziliang: Let's spin every `10` milliseconds.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
chenziliang: https://github.com/timeplus-io/proton/blob/develop/src/Storages/Streaming/StorageStream.cpp#L1740-L1792

chenziliang: The current implementation doesn't retry what has been failed. We can do this in next PR if necessary. The StorageStream has some example like how to track outstanding blocks via assigning each outstanding block an ID, and if one timeout / failed, retry.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: Dedup is not neccessary here, we can just let the latest takes the priority. If we get complains for such behavior, we can fix it then.

Also, this is to be consistent with what's defined in `KafkaLog`:
```cpp
using KConfParams = std::vector<std::pair<String, String>>;
```

And actually, I will just use this `KConfParams` in my code.

src/KafkaLog/KafkaWALPool.cpp
zliang-min: The vector header was added accidentially by the language server.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Made a const var.

src/Storages/ExternalStream/Kafka/KafkaSink.h
zliang-min: Made `is_finished` a `atomic_flag`.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
zliang-min: Removed, actually the log is not so useful.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
zliang-min: The vector  was added accidentially by the language server.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
zliang-min: The vector  was added accidentially by the language server.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Yes.

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -20,6 +24,42 @@ extern const int OK;
 extern const int RESOURCE_NOT_FOUND;
 }
 
+KConfParams Kafka::parseProperties(String properties)
```
chenziliang: You may like to have a `const String &` here to avoid a copy

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -52,6 +57,7 @@ class Kafka final : public StorageExternalStreamImpl
     Poco::Logger * log;
 
     NamesAndTypesList virtual_column_names_and_types;
+    klog::KConfParams parsed_properties;
```
chenziliang: nit : Rename as `properties`

src/Storages/ExternalStream/Kafka/KafkaSink.h
chenziliang: Using ThreadPool(1) could be a good idea here instead of a raw std::thread which has some defects and we like audit all threads if we like

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,50 @@
...
+{
+public:
+    WriteBufferFromKafka(klog::KTopicPtr topic_, Poco::Logger * logger, size_t buffer_size = DBMS_DEFAULT_BUFFER_SIZE);
+
+    ~WriteBufferFromKafka() override = default;
+
+    void onMessageDelivery(const rd_kafka_message_t *);
+
+    /// the number of acknowledgement has been received so far for the current checkpoint period
+    ALWAYS_INLINE size_t acked() const { return state.acked; }
```
chenziliang: When we are in head files, not need to specify `ALWAYS_INLINE`

chenziliang: and these routines are not called in critical loop path

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -0,0 +1,50 @@
...
+    ALWAYS_INLINE bool hasNoOutstandings() const { return state.outstandings == state.acked + state.error_count; }
+    /// allows to reset the state after each checkpoint
+    void resetState() { state.reset(); }
+
+private:
+    void nextImpl() override;
+
+    struct State final
+    {
+        std::atomic_size_t outstandings;
```
chenziliang: We probably need explicitly init the atomics

```
std::atomic_size_t outstanding = 0;
std::atomic_size_t acked = 0; 
...

```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -0,0 +1,167 @@
...
+namespace DB
+{
+
+namespace ErrorCodes
+{
+extern const int MISSING_ACKNOWLEDGEMENT;
+extern const int INVALID_CONFIG_PARAMETER;
+}
+
+KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr context, Poco::Logger * logger)
```
chenziliang: nit : `logger` -> `log_` to be consistent

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
chenziliang: This one needs resolved

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
zliang-min: These function implementations have been moved the header file since their bodies have just one line.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
zliang-min: Good to know.

src/Storages/ExternalStream/Kafka/Kafka.h
zliang-min: There is a public method call `properties`. I can name this field `properties_` instead. Does it sound right?

chenziliang: Then we are good to go without changes, or rename it to kafka_properties

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
zliang-min: It does retry. However, in theory, it's still possible it fails to send messages. In such cases, it will fail the checkpointing thus cause possible data duplication. Another consideration is to use transaction. But even with current implementation, based on my tests, failures like broker failures, network issues, etc. won't be an issue for us.

jovezhong:

jovezhong:

PR #233, opened 2023-10-30
fix: cannot create MV with kafka external stream

### LINKED ISSUE #229: Cannot create materialized view on 1.3.17 with external stream, DB::ErrnoException: Cannot open file /var/lib/proton/checkpoint/.inner.query-id.../delete, errno: 2, strerror: No such file or directory. (FILE_DOESNT_EXIST)
**Describe what's wrong**

I cannot create mv using this build

using the docker-compose in the root folder

```sql
CREATE EXTERNAL STREAM frontend_events(raw string)
                SETTINGS type='kafka',
                         brokers='redpanda:9092',
                         topic='owlshop-frontend-events';
create materialized view mv as (select now64() as _tp_time, raw:requestedUrl as url, raw:method as method, lower(hex(md5(raw:ipAddress))) as ip from frontend_events);

Received exception from server (version 1.3.17):
Code: 107. DB::Exception: Received from localhost:8463. DB::ErrnoException. DB::ErrnoException: Cannot open file /var/lib/proton/checkpoint/.inner.query-id.from-a693bf30-9210-494f-8d44-cc7c0cb734db/delete, errno: 2, strerror: No such file or directory. (FILE_DOESNT_EXIST)

```

**How to reproduce**

**Additional context**

zliang-min:
`is_streaming = true` was deleted from `KafkaSource.cpp` by accident, which caused the issue reported in #229.

I tested the change with:
* creating an MV based on kafka external stream
* gracefully stopped proton, and started it again, make sure that the MV can pick up from where it left
* deleting the MV

Closes #229

PR checklist:
- Did you run ClangFormat ?

### DIFF
--- a/src/Storages/ExternalStream/Kafka/KafkaSource.cpp
+++ b/src/Storages/ExternalStream/Kafka/KafkaSource.cpp
@@ -49,6 +49,8 @@ KafkaSource::KafkaSource(
     , virtual_col_types(header.columns(), nullptr)
     , ckpt_data(consume_ctx)
 {
+    is_streaming = true;
+
     calculateColumnPositions();
     initConsumer(kafka);
     initFormatExecutor(kafka);


src/Storages/ExternalStream/Kafka/KafkaSource.cpp
```diff
@@ -49,6 +49,8 @@ KafkaSource::KafkaSource(
     , virtual_col_types(header.columns(), nullptr)
     , ckpt_data(consume_ctx)
 {
+    is_streaming = true;
```
chenziliang, line 52: Look like something is missing, this flag shall be setup by framework atomically.  @yl-lisen

yl-lisen: We need to specify whether the current source is streaming, then other downstream pipes will automatically be set according to their input

yl-lisen (approved):

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
chenziliang: 1. hmm, i think we may need comment this protocol in the code in the base class, like 
```
/// If derived children is a streaming source, derived children need explicitly set is_streaming = true
/// to enable streaming processing.
```
2. I don't get it why we will need explicitly set this up since the storage shall indicate if it is streaming or not and the corresponding source will inherit this property from the underlying storage atomically.

yl-lisen: Agree it, the better way is that set it in `Storage::read(...)`, some sources such as `stream` or `external stream` need set the streaming flag of their built Source.

yl-lisen: Or define a super class `StreamingSourceBase`, which only set `is_streaming = true` in ctor. and some sources enabled streaming processing can inherit it

chenziliang: Let's rethink about it in another PR. merge this one first.

jovezhong:

jovezhong:

PR #245, opened 2023-11-02
Fix append time regression

### LINKED ISSUE #223: _tp_append_time column is 0
```sql
drop stream if exists test1;
create stream if not exists test1(id int, value int);
insert into test1(id, value) values(1, 2);
insert into test1(id, value) values(3, 2);
insert into test1(id, value) values(4, 2);
```
If you use `select * , _tp_append_time from test1 where _tp_time > earliest_timestamp()
`,the _tp_append_time column has no value.
image
But if you specify the column you want select, it will work well.

image

chenziliang:
PR checklist:
- Did you run ClangFormat ?

Close #223 

### DIFF
--- a/src/Core/Block.cpp
+++ b/src/Core/Block.cpp
@@ -880,6 +880,7 @@ void Block::reorderColumnsInplace(const std::vector<UInt16> & positions)
     }
 
     swap(result);
+    info = result.info;
 }
 
 void Block::renameColumn(String new_name, size_t column_pos)
--- a/src/Core/Block.cpp
+++ b/src/Core/Block.cpp
@@ -879,8 +879,9 @@ void Block::reorderColumnsInplace(const std::vector<UInt16> & positions)
         result.insert(std::move(target_col));
     }
 
-    swap(result);
-    info = result.info;
+    /// Don't swap block.info
+    data.swap(result.data);
+    index_by_name.swap(result.index_by_name);
 }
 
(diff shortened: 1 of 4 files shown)


src/Core/Block.cpp
```diff
@@ -879,7 +879,9 @@ void Block::reorderColumnsInplace(const std::vector<UInt16> & positions)
         result.insert(std::move(target_col));
     }
 
-    swap(result);
+    /// Don't swap block.info
+    data.swap(result.data);
+    index_by_name.swap(result.index_by_name);
 }
```
yl-lisen, line 885: need apply same changes in another overload function `reorderColumnsInplace`

src/NativeLog/Record/Record.cpp
```diff
@@ -135,13 +135,13 @@ RecordPtr Record::deserialize(const char * data, size_t size, const SchemaContex
         if (likely(record->codec() == DB::CompressionMethodByte::NONE))
         {
             DB::NativeReader reader(rb, DBMS_TCP_PROTOCOL_VERSION);
-            record->setBlock(reader.read());
+            record->setBlock(reader.read(), /*override_block_info=*/false);
```
yl-lisen, line 138: it seems that deserialized block will lose block info. Probably needs to fix. in future

chenziliang: So far we don't carry any information in block.info in a Record when serialization.

PR #252, opened 2023-11-06
enable backfill from historical store

### LINKED ISSUE #162: Backfill revisit
**Use case**

In Proton, data retention policy is different in streaming store and historical data store. When we rewind data (`_tp_time > now - 7d` for example), there is possibility that the data is already gone in streaming store. In such case, we will need backfill the data from the historical data store and when the backfill is done, re-connects it with the live stream events.

Backfilling data from the historical data store may require `re-sort` the data depending on the query semantics. For example, in streaming tumble / hop / session etc window aggregation, we will need resort the data according to `_tp_time` or the logic time sequence number `_tp_sn`. 

`_tp_sn` is not indexed in the historical data store by default but can enable it automatically when a stream has `_tp_sn` column during provision. Sorting by `_tp_sn` could be very expensive since data in the historical data store is usually not sorted by sequence number. 

`_tp_time` is indexed in the historical data store by default for every event in every stream and data is sorted by `_tp_time` in coarse granularity there. So re-sort data by `_tp_time` is usually way cheaper than `_tp_sn`.

From most application business point of view, when the data is already past / historical, backfill in `_tp_time` (event time) order seems making lots of sense instead of in the logic `_tp_sn` sequence number order (ingest time order).

So this issue tries to make backfill smarter and reasonable for query like below. Once this enhancement is done, it will automatically backfill events from the historical data store with necessary `sorting`.

```
SELEC max(cpu), min(cpu), device 
FROM tumble(device_utils, 5m)
WHERE _tp_time > now() - 7d;
```

For query like below, we don't need resort historical data during backfill since global aggregation doesn't care time / temporal. 
```
SELEC max(cpu), min(cpu), device 
FROM device_utils
WHERE _tp_time > now() - 7d;
```

**Additional context**

yl-lisen:
PR checklist:
- Did you run ClangFormat ?

This close #162 , the change log:
- Enable backfill from historical store by setting `enable_backfill_from_historical_store` (by default is true)
- Add a setting `emit_aggregated_during_backfill` to control whether need emit aggregated result during backfill (by default is true)
- Optimization: when `emit_aggregated_during_backfill=false`, global aggregation don't require backfilled data in order

### DIFF
--- a/src/Interpreters/InterpreterSelectQuery.cpp
+++ b/src/Interpreters/InterpreterSelectQuery.cpp
@@ -805,10 +805,6 @@ InterpreterSelectQuery::InterpreterSelectQuery(
         /// and create `SeekToInfo` objects to represent these predicates for streaming store rewinding in a streaming query.
         analyzeEventPredicateAsSeekTo(joined_tables);
 
-        /// Optimization: no requires backfill data in order for global aggregation with settings `emit_aggregated_during_backfill = false`.
-        if (!context->getSettingsRef().emit_aggregated_during_backfill.value && hasGlobalAggregation())
-            query_info.requires_backfill_input_in_order = false;
-
         /// Calculate structure of the result.
         result_header = getSampleBlockImpl();
         /// proton, FIXME. For distributed streaming query in future, we may need conditionally remove __tp_ts from result header
@@ -894,8 +890,7 @@ InterpreterSelectQuery::InterpreterSelectQuery(
     }
 
     /// proton: starts
-    checkForStreamingQuery();
-    checkUDA();
+    finalCheckAndOptimizeForStreamingQuery();
 
     if (query_info.projection)
         storage_snapshot->addProjection(query_info.projection->desc);
@@ -2343,11 +2338,6 @@ void InterpreterSelectQuery::executeFetchColumns(QueryProcessingStage::Enum proc
         /// need replay operation
         if (settings.replay_speed > 0)
         {
-            if (settings.enable_backfill_from_historical_store.value)
-                throw Exception(
-                    ErrorCodes::NOT_IMPLEMENTED,
-                    "Only support replay from streaming store, but setting 'enable_backfill_from_historical_store=true'");
-
             /// So far, only support append-only stream (or proxyed)
             StorageStream * storagestream = nullptr;
             if (Streaming::isAppendStorage(storage->dataStreamSemantic()))
@@ -3372,7 +3362,7 @@ bool InterpreterSelectQuery::shouldKeepState() const
     return false;
 }
 
-void InterpreterSelectQuery::checkForStreamingQuery() const
+void InterpreterSelectQuery::finalCheckAndOptimizeForStreamingQuery()
 {
     if (isStreaming())
     {
@@ -3384,6 +3374,27 @@ void InterpreterSelectQuery::checkForStreamingQuery() const
                 if (proxy->hasGlobalAggregation())
                     throw Exception("Streaming query doesn't support window func over a global aggregation", ErrorCodes::NOT_IMPLEMENTED);
         }
+
+        /// We will only be able to read data from streaming store in the following scenarios:
+        /// 1) required virtual column streaming only, such as `_tp_sn`
+        /// 2) seek to by absolute `_tp_sn`
+        /// 3) replay stream
+        const auto & settings = context->getSettingsRef();
+        if (settings.enable_backfill_from_historical_store.value)
+        {
+            bool required_virtual_col_streaming_only = std::ranges::any_of(required_columns, [](const auto & name) {
+                return name == ProtonConsts::RESERVED_EVENT_SEQUENCE_ID || name == ProtonConsts::RESERVED_APPEND_TIME
+                    || name == ProtonConsts::RESERVED_INGEST_TIME || name == ProtonConsts::RESERVED_PROCESS_TIME;
+            });
+            bool seek_to_by_absolute_sn = !query_info.seek_to_info->getSeekTo().empty() && !query_info.seek_to_info->isTimeBased()
+                && query_info.seek_to_info->getSeekTo() != "earliest";
+            if (required_virtual_col_streaming_only || seek_to_by_absolute_sn || settings.replay_speed > 0)
+                context->setSetting("enable_backfill_from_historical_store", false);
+        }
+
+        /// Optimization: no requires backfill data in order for global aggregation with settings `emit_aggregated_during_backfill = false`.
+        if (!settings.emit_aggregated_during_backfill.value && hasGlobalAggregation())
+            query_info.require_in_order_backfill = false;
     }
     else
     {
@@ -3408,6 +3419,8 @@ void InterpreterSelectQuery::checkForStreamingQuery() const
                 "Neither window_start nor window_end is referenced in the query, but streaming window function is used",
                 ErrorCodes::WINDOW_COLUMN_NOT_REFERENCED);
     }
+
+    checkUDA();
 }
 
 void InterpreterSelectQuery::buildShufflingQueryPlan(QueryPlan & query_plan)
--- a/src/Interpreters/InterpreterSelectQuery.cpp
+++ b/src/Interpreters/InterpreterSelectQuery.cpp
@@ -805,6 +805,10 @@ InterpreterSelectQuery::InterpreterSelectQuery(
         /// and create `SeekToInfo` objects to represent these predicates for streaming store rewinding in a streaming query.
         analyzeEventPredicateAsSeekTo(joined_tables);
 
+        /// Optimization: no requires backfill data in order for global aggregation with settings `emit_aggregated_during_backfill = false`.
+        if (!context->getSettingsRef().emit_aggregated_during_backfill.value && hasGlobalAggregation())
+            query_info.requires_backfill_input_in_order = false;
+
         /// Calculate structure of the result.
         result_header = getSampleBlockImpl();
         /// proton, FIXME. For distributed streaming query in future, we may need conditionally remove __tp_ts from result header
@@ -3432,12 +3436,16 @@ void InterpreterSelectQuery::buildWatermarkQueryPlan(QueryPlan & query_plan) con
 {
     assert(isStreaming());
     auto params = std::make_shared<Streaming::WatermarkStamperParams>(
-                query_info.query, query_info.syntax_analyzer_result, query_info.streaming_window_params);
+        query_info.query, query_info.syntax_analyzer_result, query_info.streaming_window_params);
+
+    bool skip_stamping_for_backfill_data = !context->getSettingsRef().emit_aggregated_during_backfill.value;
+
     if (query_info.hasPartitionByKeys())
-        query_plan.addStep(
-            std::make_unique<Streaming::WatermarkStepWithSubstream>(query_plan.getCurrentDataStream(), std::move(params), log));
+        query_plan.addStep(std::make_unique<Streaming::WatermarkStepWithSubstream>(
+            query_plan.getCurrentDataStream(), std::move(params), skip_stamping_for_backfill_data, log));
     else
-        query_plan.addStep(std::make_unique<Streaming::WatermarkStep>(query_plan.getCurrentDataStream(), std::move(params), log));
+        query_plan.addStep(std::make_unique<Streaming::WatermarkStep>(
+            query_plan.getCurrentDataStream(), std::move(params), skip_stamping_for_backfill_data, log));
 }
 
 void InterpreterSelectQuery::buildStreamingProcessingQueryPlanBeforeJoin(QueryPlan & query_plan)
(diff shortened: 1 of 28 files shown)


src/Interpreters/InterpreterSelectQuery.cpp
```diff
@@ -2339,6 +2343,11 @@ void InterpreterSelectQuery::executeFetchColumns(QueryProcessingStage::Enum proc
         /// need replay operation
         if (settings.replay_speed > 0)
         {
+            if (settings.enable_backfill_from_historical_store.value)
```
chenziliang: since backfill flag is true by default, let's reset it to `false` if `replay_speed` is on, otherwise user will need do `select * from my_stream settings replay_speed = 1, enable_backfill_from_historical_store=false`

tests/stream/test_stream_smoke/0099_fixed_issues.json
```diff
@@ -149,8 +149,8 @@
             {"client":"python", "query_type": "table", "depends_on_stream":"test_mv_1934", "wait":2, "query": "insert into test_stream_1934(value) values(1)"},
             {"client":"python", "query_type": "table", "wait":3, "query": "insert into test_stream_1934(value) values(2)"},
             {"client":"python", "query_type": "table", "wait":3, "query": "insert into test_stream_1934(value) values(3)"},
-            {"client":"python", "query_type": "stream", "query_id":"9900", "query_end_timer":3, "query":"select `count()`, _tp_shard, _tp_sn from test_mv_1934 where _tp_sn >= 0"},
-            {"client":"python", "query_type": "stream", "query_id":"9901", "query_end_timer":3,"drop_view":"test_mv_1934", "drop_view_wait":2, "query":"select `count()`, _tp_shard, _tp_sn from test_mv_1934 where _tp_time >= now() - 1m"}
+            {"client":"python", "query_type": "stream", "query_id":"9900", "query_end_timer":3, "query":"select `count()`, _tp_shard, _tp_sn from test_mv_1934 where _tp_sn >= 0 settings enable_backfill_from_historical_store=false"},
```
chenziliang: For `_tp_sn` seek to, always backfill from streaming store, we don't even need explicit reset backfill flag to false. We can resolve this in a separate PR though.

tests/stream/test_stream_smoke/0099_fixed_issues.json
```diff
@@ -149,8 +149,8 @@
...
+            {"client":"python", "query_type": "stream", "query_id":"9901", "query_end_timer":3,"drop_view":"test_mv_1934", "drop_view_wait":2, "query":"select `count()`, _tp_shard, _tp_sn from test_mv_1934 where _tp_time >= now() - 1m settings enable_backfill_from_historical_store=false"}
```
chenziliang: When virtual column `_tp_sn` is selected, we can't backfill from historical store by default.

tests/stream/test_stream_smoke/0028_replay_stream.yaml
```diff
@@ -53,12 +53,12 @@ tests:
...
             query_type: stream
             terminate: auto
-            query_end_timer: 20
+            query_end_timer: 10
             query_id: '2900'
             wait: 1
             query: with cte as (select id, value, date_diff('millisecond', to_datetime64(lag(_tp_append_time, 1)/1000, 3, 'UTC'), to_datetime64(_tp_append_time/1000, 3, 'UTC')) as insert_interval,
-                    date_diff('millisecond',lag(now64(), 1), now64()) as output_time_diff, _tp_time from test29_stream where _tp_time > earliest_timestamp() settings replay_speed=1)
-                    select id, value, abs(insert_interval-2000) < 200, abs(insert_interval-output_time_diff) < 10 from cte;
+                    date_diff('millisecond',lag(now64(), 1), now64()) as output_time_diff, _tp_time from test29_stream where _tp_time > earliest_timestamp() settings replay_speed=1, enable_backfill_from_historical_store=false)
```
chenziliang: when we reset backfill flag to false for replay in the code, we don't need explicitly set backfill flag to false here.

tests/stream/test_stream_smoke/0013_changelog_stream3.json
```diff
@@ -144,26 +144,22 @@
...
             ],
             "expected_results": [
                 {
                     "query_id":"1455",
                     "expected_results":[
-                        [2, 1, 0.0, 33.29999923706055],
-                        [2, 2, 0.0, 33.29999923706055],
-                        [2, 3, 0.0, 33.29999923706055],
-                        [2, 3, 0.0, 44.400001525878906]
+                        [1, 1, 33.3, 33.3],
```
chenziliang, line 159: What cause the expected results changes

src/Storages/SelectQueryInfo.h
```diff
@@ -183,6 +183,9 @@ struct SelectQueryInfo
     SeekToInfoPtr seek_to_info; /// Rewind info for left streaming store in streaming query
     SeekToInfoPtr seek_to_info_of_right_stream; /// Rewind info for right streaming store in streaming query
 
+    /// if is true, means need sort backfill input by event time (ascending)
+    bool requires_backfill_input_in_order = true;
```
chenziliang: `require_in_order_backfill`

src/Storages/Streaming/StorageStream.cpp
```diff
@@ -991,9 +997,23 @@ StorageStream::ShardsToRead StorageStream::getRequiredShardsToRead(ContextPtr co
...
+        /// We allow backfill from historical store the following scenarios:
+        /// 1) For non-inmemory keyed storage stream, we always back fill from historical store (e.g. VersionedKV, ChangelogKV)
+        /// 2) Do time travel with settings `enable_backfill_from_historical_store = true`
+        bool require_back_fill_from_historical = false;
+        if (!isInmemory() && Streaming::isKeyedStorage(dataStreamSemantic()))
+            require_back_fill_from_historical = true;
+        else if (!query_info.seek_to_info->getSeekTo().empty() && settings_ref.enable_backfill_from_historical_store.value)
+        {
+            if (!query_info.seek_to_info->isTimeBased() && query_info.seek_to_info->getSeekTo() != "earliest")
+                throw Exception(
```
chenziliang: As mentioned, to provide best user experience, when seek to with sn, disable backfill from historical store automatically.

src/Processors/Transforms/Streaming/WatermarkTransformWithSubstream.h
```diff
@@ -39,6 +40,9 @@ class WatermarkTransformWithSubstream final : public IProcessor
     WatermarkStamperPtr watermark_template;
     SERDE SubstreamHashMap<WatermarkStamperPtr> substream_watermarks;
 
+    const bool skip_stamping_for_backfill_data;
```
chenziliang: I think just `bool` will do its work.

src/Processors/Transforms/Streaming/WatermarkTransformWithSubstream.h
```diff
@@ -39,6 +40,9 @@ class WatermarkTransformWithSubstream final : public IProcessor
...
+    bool is_backfill_data = false;
```
chenziliang: `bool is_backfilling_data` or `is_in_backfilling`

src/Processors/Transforms/Streaming/WatermarkTransform.h
```diff
@@ -31,6 +31,9 @@ class WatermarkTransform final : public ISimpleTransform
 private:
     WatermarkStamperParamsPtr params;
     SERDE WatermarkStamperPtr watermark;
+
+    const bool skip_stamping_for_backfill_data;
+    bool is_backfill_data = false;
```
chenziliang: Same comments as in SubStream watermark

src/Processors/Transforms/Streaming/GlobalAggregatingTransform.cpp
```diff
@@ -125,11 +125,21 @@ void GlobalAggregatingTransform::finalize(const ChunkContextPtr & chunk_ctx)
         many_data->finalized_watermark.store(chunk_ctx->getWatermark(), std::memory_order_relaxed);
     });
 
-    auto first = many_data->variants.at(0);
-    if (unlikely(first->isTwoLevel()))
+    AggregatedDataVariantsPtr non_empty_one;
+    for (const auto & data : many_data->variants)
```
chenziliang: Try `std::ranges::find_first_of(...)` ?

src/Processors/Transforms/Streaming/GlobalAggregatingTransform.cpp
```diff
@@ -125,11 +125,21 @@ void GlobalAggregatingTransform::finalize(const ChunkContextPtr & chunk_ctx)
         many_data->finalized_watermark.store(chunk_ctx->getWatermark(), std::memory_order_relaxed);
     });
 
-    auto first = many_data->variants.at(0);
-    if (unlikely(first->isTwoLevel()))
+    AggregatedDataVariantsPtr non_empty_one;
```
chenziliang: `first_data`

src/Storages/Streaming/StorageStream.cpp
```diff
@@ -991,9 +997,23 @@ StorageStream::ShardsToRead StorageStream::getRequiredShardsToRead(ContextPtr co
...
+                    ErrorCodes::UNSUPPORTED,
+                    "Seek to by absolute sequence number is not supported when set 'enable_backfill_from_historical_store=true'");
+
+            require_back_fill_from_historical = true;
+        }
+
         result.mode = require_back_fill_from_historical ? QueryMode::STREAMING_CONCAT : QueryMode::STREAMING;
```
chenziliang, line 1017: Right now, the historical part of `STREAMING_CONCAT` is single thread, let's fix that part as well

src/Processors/Transforms/Streaming/WatermarkTransformWithSubstream.h
chenziliang: but don't have strong opinion since const most of the time prevent mistakes

tests/stream/test_stream_smoke/0013_changelog_stream3.json
yl-lisen: This test case is unstable and counts a deleted group (i.e. empty default value), so I modified it.

yl-lisen: BTW, global aggregation emit the deleted empty group is an issue to fix in future

src/Interpreters/InterpreterSelectQuery.cpp
```diff
@@ -3384,6 +3374,27 @@ void InterpreterSelectQuery::checkForStreamingQuery() const
...
+        /// 2) seek to by absolute `_tp_sn`
+        /// 3) replay stream
+        const auto & settings = context->getSettingsRef();
+        if (settings.enable_backfill_from_historical_store.value)
+        {
+            bool required_virtual_col_streaming_only = std::ranges::any_of(required_columns, [](const auto & name) {
+                return name == ProtonConsts::RESERVED_EVENT_SEQUENCE_ID || name == ProtonConsts::RESERVED_APPEND_TIME
+                    || name == ProtonConsts::RESERVED_INGEST_TIME || name == ProtonConsts::RESERVED_PROCESS_TIME;
+            });
+            bool seek_to_by_absolute_sn = !query_info.seek_to_info->getSeekTo().empty() && !query_info.seek_to_info->isTimeBased()
```
chenziliang: `seek_by_sn`

src/Interpreters/InterpreterSelectQuery.cpp
```diff
@@ -3384,6 +3374,27 @@ void InterpreterSelectQuery::checkForStreamingQuery() const
                 if (proxy->hasGlobalAggregation())
                     throw Exception("Streaming query doesn't support window func over a global aggregation", ErrorCodes::NOT_IMPLEMENTED);
         }
+
+        /// We will only be able to read data from streaming store in the following scenarios:
```
chenziliang:
```
/// For now, for the following scenarios, we disable backfill from historic data store
/// 1) User select some virtual columns which is only available in streaming store, like `_tp_sn`, `_tp_index_time`
/// 2) Seek by streaming store sequence number
/// 3) Replaying a stream. 
/// TODO, ideally we shall check if historical data store has `_tp_sn` etc columns, if they have, we can backfill from
/// the historical data store as well technically. This will be a future enhancement.
```

src/Interpreters/InterpreterSelectQuery.cpp
```diff
@@ -3384,6 +3374,27 @@ void InterpreterSelectQuery::checkForStreamingQuery() const
...
+        /// 1) required virtual column streaming only, such as `_tp_sn`
+        /// 2) seek to by absolute `_tp_sn`
+        /// 3) replay stream
+        const auto & settings = context->getSettingsRef();
+        if (settings.enable_backfill_from_historical_store.value)
+        {
+            bool required_virtual_col_streaming_only = std::ranges::any_of(required_columns, [](const auto & name) {
```
chenziliang: `has_streaming_only_virtual_columns = ...`

chenziliang (approved):

Issue #198, opened 2023-10-20
config-less Proton


yokofly:
**Describe what's wrong**
I manually built ClickHouse and the Proton binary locally. To start ClickHouse directly, I use:

```sh
./clickhouse-server
```

However, Proton fails to start. This means we must manually specify the path in proton/programs/server/config.yaml when launching the Proton server.

Note: Whether using the downloaded binary or the built binary, ClickHouse operates smoothly without the need to manually edit the path in config.xml.

**Error message and/or stacktrace**
```
Process 490393 launched: '/proton/build_debug_buildsh/programs/proton-server' (x86_64)
Processing configuration file 'config.xml'.
Code: 107. DB::Exception: Configuration file config.xml doesn't exist and there is no embedded config. (FILE_DOESNT_EXIST), Stack trace (when copying this message, always include the lines below):

0. /proton/contrib/llvm-project/libcxx/include/exception:134: std::exception::capture() @ 0xe85c762 in /proton/build_debug_buildsh/programs/proton
1. /proton/contrib/llvm-project/libcxx/include/exception:112: std::exception::exception[abi:v15000]() @ 0xe85c71b in /proton/build_debug_buildsh/programs/proton
2. /proton/base/poco/Foundation/src/Exception.cpp:27: Poco::Exception::Exception(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, int) @ 0x4acc1ac0 in /proton/build_debug_buildsh/programs/proton
3. /proton/src/Common/Exception.cpp:67: DB::Exception::Exception(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, int, bool) @ 0x24be1d1e in /proton/build_debug_buildsh/programs/proton
4. /proton/src/Common/Exception.h:43: DB::Exception::Exception<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&>(int, fmt::v8::basic_format_string<char, fmt::v8::type_identity<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&>::type>, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) @ 0xfc7f498 in /proton/build_debug_buildsh/programs/proton
[...]

PR #254, opened 2023-11-07
Open additional ports for embedded config and add folder for path

yokofly:

If the server is launched directly, it will access the current folder(generating below file/folder) instead of a dedicated folder. 
To avoid this, change `path: ./` to `path: ./proton-data/`.
```
checkpoint		data			flags			metadata		nativelog		store			user_defined		user_scripts
coordination		dictionaries_lib	format_schemas		metadata_dropped	preprocessed_configs	tmp			user_files		uuid
```

### DIFF
--- a/programs/server/embedded.xml
+++ b/programs/server/embedded.xml
@@ -7,10 +7,16 @@
     </logger>
 
     <http_port>3218</http_port>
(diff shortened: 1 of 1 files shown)


yokofly:
continue https://github.com/timeplus-io/proton/issues/198

PR #258, opened 2023-11-07
shuffle by

### LINKED ISSUE #256: Light shuffle

For large scale unique key aggregation etc scenario, we would like to shuffle the data first to minimize the memory footprint and make the down stream aggregator more efficient.

chenziliang:
PR checklist:
- Did you run ClangFormat ?

Closed #256.

Streaming part will be done in a separate PR

### DIFF
--- a/src/Interpreters/InterpreterSelectQuery.cpp
+++ b/src/Interpreters/InterpreterSelectQuery.cpp
@@ -98,6 +98,7 @@
 #include <Interpreters/Streaming/TableFunctionDescription.h>
 #include <Parsers/ASTWindowDefinition.h>
 #include <Parsers/Streaming/ASTEmitQuery.h>
+#include <Processors/QueryPlan/LightShufflingStep.h>
 #include <Processors/QueryPlan/Streaming/AggregatingStep.h>
 #include <Processors/QueryPlan/Streaming/AggregatingStepWithSubstream.h>
 #include <Processors/QueryPlan/Streaming/JoinStep.h>
@@ -165,13 +166,13 @@ void addEventTimePredicate(ASTSelectQuery & select, Int64 utc_ms)
         select.setExpression(ASTSelectQuery::Expression::WHERE, greater);
 }
 
-std::vector<size_t> keyPositionsForSubstreams(const Block & header, const SelectQueryInfo & query_info)
+std::vector<size_t> keyPositions(const Block & header, const Names & key_columns)
 {
-    std::vector<size_t> substream_key_positions;
-    substream_key_positions.reserve(query_info.partition_by_keys.size());
-    for (const auto & key : query_info.partition_by_keys)
-        substream_key_positions.emplace_back(header.getPositionByName(key));
-    return substream_key_positions;
+    std::vector<size_t> key_positions;
+    key_positions.reserve(key_columns.size());
+    for (const auto & key : key_columns)
+        key_positions.emplace_back(header.getPositionByName(key));
+    return key_positions;
 }
 
 /// Requires: 1) no window function 2) has aggregation
@@ -188,7 +189,7 @@ bool hasGlobalAggregationInQuery(const ASTPtr & query, const ASTSelectQuery & se
     return !data.aggregates.empty() || select_query.groupBy() != nullptr;
 }
 
-[[maybe_unused]] Names getShuffleByColumns(const ASTSelectQuery & query)
+Names getShuffleByColumns(const ASTSelectQuery & query)
 {
     Names shuffle_by_columns;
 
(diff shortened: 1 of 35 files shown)


src/Interpreters/InterpreterSelectQuery.cpp
```diff
@@ -187,6 +188,23 @@ bool hasGlobalAggregationInQuery(const ASTPtr & query, const ASTSelectQuery & se
...
+    Names shuffle_by_columns;
+
+    if (!query.shuffleBy())
+        return shuffle_by_columns;
+
+    shuffle_by_columns.reserve(query.shuffleBy()->children.size());
+    for (const auto & elem : query.shuffleBy()->children)
+    {
+        if (elem->as<ASTIdentifier>())
+            shuffle_by_columns.push_back(elem->getColumnName());
```
yl-lisen, line 203: If shuffle by keys are an expression or constant, ignore them ?
I guess the shuffle by keys should be calculated before light shuffling, so we can always use `shuffle_by_columns.push_back(elem->getColumnName())` even if it is not an identifier

src/Parsers/ParserSelectQuery.cpp
```diff
@@ -213,6 +217,24 @@ bool ParserSelectQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & expected,
...
+
+    if (partition_by_expression_list && shuffle_by_expression_list)
+        throw Exception(ErrorCodes::SYNTAX_ERROR, "'PARTITION BY' and 'SHUFFLE BY' shall not be used together");
+
+    /// For now, only support `SHUFFLE BY col1, col2, ...`, so expecting column identifier
+    /// `SHUFFLE BY const_literal` seems doesn't make sense
+    if (shuffle_by_expression_list)
+    {
+        for (const auto & elem : shuffle_by_expression_list->children)
+            if (!elem->as<ASTIdentifier>())
```
yl-lisen, line 235: The Identifier may be an alias of expression/function.
If want to validate whether is column, we can call `IdentifierSemantic::getMembership` after `TranslateQualifiedNamesVisitor` in TreeRewriter,

src/Processors/LightChunkSplitter.cpp
```diff
@@ -0,0 +1,89 @@
...
+    assert(num_rows > 0);
+
+    const auto & columns = chunk.getColumns();
+
+    WeakHash32 hash(num_rows);
+    for (const auto & key_column_position : key_column_positions)
+    {
+        assert(columns.size() > key_column_position);
+        const auto & key_col = columns[key_column_position]->convertToFullColumnIfConst();
+        const auto & key_col_no_lc = recursiveRemoveLowCardinality(recursiveRemoveSparse(key_col));
```
yl-lisen, line 45: Don't need this code, since `ColumnConst/ColumnLowCardinality/ColumnSparse` have been implemented this method `updateWeakHash32`

src/Processors/Transforms/AggregatingTransform.cpp
```diff
@@ -280,16 +352,67 @@ class ConvertingAggregatedToChunksTransform final : public IProcessor
...
+    IProcessor::Status prepareTwoLevelShuffled()
+    {
+        auto & output = outputs.front();
+
+        size_t new_processed = 0;
+        for (auto & input : inputs)
+        {
+            if (!input.isFinished() && input.hasData())
+            {
+                shuffled_chunks.push_back(input.pull());
```
yl-lisen, line 366: NOTE, for streaming aggregating, we need to use `pull(/*set_not_needed*/true)`, since the inputs can receive heartbeat chunk, then trigger upstream again, next loop above actions, which will cause `pipeline stuck`, there are some reasons:
1) The upstream is triggered first, resulting in no thread calling the downstream processor.
2) Missing propagate heartbeat chunk can also cause pipeline stuck.

Of course, it works in historical aggregating.

src/Parsers/ParserSelectQuery.cpp
chenziliang: yeah, thought about this. For now, probably it is fine and we may extend this soon to support expression for shuffle by.

src/Processors/Transforms/AggregatingTransform.cpp
chenziliang: This is pure historical data. Streaming/AggregatingTransform.cpp is not covered yet

src/Interpreters/InterpreterSelectQuery.cpp
chenziliang: we will still do the work i think and it shall be fine just like regular columns.

chenziliang: Meaning if end user do something like aliasing a constant or an expression / function as an identifier, we will do the shuffle work as expected.

When the logic reaches here, the ast of the shuffle by can be asserted to ASTIdentifier since it is already validated in the AST parser.

src/Processors/LightChunkSplitter.cpp
chenziliang: I just checked the code. I am not quite sure if they will produce same hash results. Have you tested it out ?

chenziliang: Testing mean, 2 or more different low cardinality blocks for example, they contain some overlapping lower cardinality rows but with other rows different, when we do the hash, we will need make sure same LC rows across different blocks will shuffled to same shard (having the same hash).  Same for sparse column.

PR #263, opened 2023-11-09
Bugfix/issue 239 fix client cancel

### LINKED ISSUE #239: proton client ignore ctrl+c on select from system.numbers
1. open proton-client
2. query without limit `select number from system.numbers where number % 10 = 0` 
3. try pass ctrl+c

In very rare cases ctrl+c work and cause stop but usually client just ignore it.
I think that cause is that select from numbers without limit is very intensive for CPU and client ignore input stream.

So need make to react to ctrl+c as stop output action even on intensive queries

yokofly:

compared with clickhouse-client, proton-client has many flaws since we do not take too much focus previous.
and this PR port some bugfix PR from clickhouse community, also fix #239 (locally test good)


yokofly:
_edit v1: update commit detail:_
this PR key commits i think is 4a98cc875626cfaf7f9d479b91e98e6bf76c9a25
this mutex guard also be helpful  587d81aa61e62ecd1d727ac778f4a06394d05e9b

even with this, there are still more necessary PR I have not covered, maybe later we can backport again.
https://github.com/ClickHouse/ClickHouse/pull/42362/files
https://github.com/ClickHouse/ClickHouse/pull/46681/files


one more thing:
the query_log does not record the insert by default, need manually add set `log_insert_query`
like `insert into v(id) settings log_insert_query=true values(2)`

yokofly:
_edit v1: update with internal discussion:_
@qijun-niu-timeplus  this pr introduces a test on proton-local, I will enable it on the docker/packager/packager build script.

- enable proton-local on CI build / single binary release
- disable proton-local on the docker image  release

later we can try clickhouse self-extract (when first started up will decompress whole bundle)

chenziliang (approved):

yokofly:
#265 #266

PR #273, opened 2023-11-13
feature: kafka external stream partitioning

### LINKED ISSUE #192: Shard strategy for Kafka Sink
**Use case**

As a user, i would like to shard the data to my Kafka topic according to some sharding expression when sink to it.

```
CREATE EXTERNAL STREAM sharded_external_stream 
(raw string, id string)
SETTINGS sharding_expr=hash(id);
```

https://github.com/timeplus-io/proton/blob/develop/src/Storages/MergeTree/MergeTreeSettings.h#L171
https://github.com/timeplus-io/proton/blob/develop/src/Storages/Streaming/StreamSink.cpp#L55-L84

**Additional context**

zliang-min:
PR checklist:
- Did you run ClangFormat ?

A new feature that adds a new setting name `sharding_expr` for creating kafka external stream. `sharding_expr` accepts a SQL expression which is evaluated into an integer, and the value of that expression is used to decide to which partition the message is sent. For example:

```SQL
CREATE EXTERNAL STREAM foo (
a string,
b int32,
) type='kafka',brokers='localhost:9092',topic='foo',sharding_expr='to_int(b)'
```

The `KafkaSink` will use the expression to divide each in-coming `Chunk` into `Block`s by the sharding key value and then send the `Block`s to the Kafka topic.

closes #192

### DIFF
--- a/src/Storages/ExternalStream/Kafka/KafkaSink.cpp
+++ b/src/Storages/ExternalStream/Kafka/KafkaSink.cpp
@@ -1,7 +1,5 @@
 #include "KafkaSink.h"
-#include <rdkafka.h>
 
-#include <Common/logger_useful.h>
 #include <DataTypes/DataTypeLowCardinality.h>
 #include <DataTypes/DataTypesNumber.h>
 #include <Interpreters/ExpressionAnalyzer.h>
@@ -9,6 +7,7 @@
 #include <Interpreters/createBlockSelector.h>
 #include <Parsers/ASTFunction.h>
 #include <Processors/Formats/IOutputFormat.h>
+#include <Common/logger_useful.h>
 
 #include <boost/algorithm/string/predicate.hpp>
 
@@ -25,7 +24,8 @@ extern const int TYPE_MISMATCH;
 
 namespace KafkaStream
 {
-ChunkPartitioner::ChunkPartitioner(ContextPtr context, const Block & header, const ASTPtr & partitioning_expr_ast) {
+ChunkPartitioner::ChunkPartitioner(ContextPtr context, const Block & header, const ASTPtr & partitioning_expr_ast)
+{
     /// `InterpreterCreateQuery::handleExternalStreamCreation` ensures this
     assert(partitioning_expr_ast);
 
@@ -36,20 +36,16 @@ ChunkPartitioner::ChunkPartitioner(ContextPtr context, const Block & header, con
     partitioning_key_column_name = partitioning_expr_ast->getColumnName();
 
     if (auto * shard_func = partitioning_expr_ast->as<ASTFunction>())
+    {
         if (shard_func->name == "rand" || shard_func->name == "RAND")
-        {
             random_partitioning = true;
-            std::random_device r;
-            rand = std::minstd_rand(r());
-        }
+    }
 }
 
 BlocksWithShard ChunkPartitioner::partition(Block block, Int32 partition_cnt) const
 {
     /// no topics have zero partitions
     assert(partition_cnt > 0);
-    /// check the ctor for how this is guaruanteed
-    assert(partitioning_expr);
 
     if (partition_cnt == 1)
         return {BlockWithShard{Block(std::move(block)), 0}};
@@ -88,15 +84,14 @@ BlocksWithShard ChunkPartitioner::doParition(Block block, Int32 partition_cnt) c
     return blocks_with_shard;
 }
 
-IColumn::Selector ChunkPartitioner::createSelector(const Block & block, Int32 partition_cnt) const
+IColumn::Selector ChunkPartitioner::createSelector(Block block, Int32 partition_cnt) const
 {
     std::vector<UInt64> slot_to_shard(partition_cnt);
     std::iota(slot_to_shard.begin(), slot_to_shard.end(), 0);
 
-    Block current_block = block;
-    partitioning_expr->execute(current_block);
+    partitioning_expr->execute(block);
 
-    const auto & key_column = current_block.getByName(partitioning_key_column_name);
+    const auto & key_column = block.getByName(partitioning_key_column_name);
 
 /// If key_column.type is DataTypeLowCardinality, do shard according to its dictionaryType
 #define CREATE_FOR_TYPE(TYPE) \
@@ -122,7 +117,12 @@ IColumn::Selector ChunkPartitioner::createSelector(const Block & block, Int32 pa
 }
 
 KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr context, Int32 initial_partition_cnt, Poco::Logger * log_)
-    : SinkToStorage(header, ProcessorID::ExternalTableDataSinkID), producer(nullptr, nullptr), topic(nullptr, nullptr), polling_threads(1), partition_cnt(initial_partition_cnt), log(log_)
+    : SinkToStorage(header, ProcessorID::ExternalTableDataSinkID)
+    , producer(nullptr, nullptr)
+    , topic(nullptr, nullptr)
+    , polling_threads(1)
+    , partition_cnt(initial_partition_cnt)
+    , log(log_)
 {
     /// default values
     std::vector<std::pair<String, String>> producer_params{
@@ -198,30 +198,9 @@ KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr conte
     rd_kafka_topic_conf_set_opaque(topic_conf, this);
 
     /// With partitioner callback, we can get the up-to-date partition count w/o additional effort.
-    rd_kafka_topic_conf_set_partitioner_cb(topic_conf, [](const rd_kafka_topic_t * /*rkt*/,
-						const void * /*keydata*/,
-						size_t /*keylen*/,
-						int32_t partition_count,
-						void * rkt_opaque,
-						void * msg_opaque) -> int32_t
-    {
-        /// update partition count
-        auto * sink = static_cast<KafkaSink *>(rkt_opaque);
-        sink->partition_cnt = partition_count;
+    rd_kafka_topic_conf_set_partitioner_cb(topic_conf, &KafkaSink::onPartitioning);
 
-        auto * partition_id_ptr = static_cast<Int32 *>(msg_opaque);
-        auto parition_id =  *partition_id_ptr;
-        delete partition_id_ptr;
-        /// This should not really happen because Kafka does not support reducing partitions.
-        /// However, KIP-694 is currently under discussion, so this might heppen in the future.
-        if (parition_id >= partition_count)
-            parition_id = partition_count - 1;
-        return parition_id;
-    });
-
-    rd_kafka_conf_set_dr_msg_cb(conf, [](rd_kafka_t * /* producer */, const rd_kafka_message_t * msg, void * opaque) {
-        static_cast<KafkaSink *>(opaque)->wb->onMessageDelivery(msg);
-    });
+    rd_kafka_conf_set_dr_msg_cb(conf, &KafkaSink::onMessageDelivery);
 
     producer = klog::KafkaPtr(rd_kafka_new(RD_KAFKA_PRODUCER, conf, errstr, sizeof(errstr)), rd_kafka_destroy);
     if (!producer)
@@ -264,7 +243,7 @@ void KafkaSink::consume(Chunk chunk)
         /// Since we set AutoFlush on writer, it makes sure that `writer->write` will call
         /// `wb->nextImpl` and waits for it finishes, it's safe to call `wb->write_to_partition`
         /// before `writer->write`.
-        wb->write_to_partition(blockWithShard.shard);
+        wb->setTargetPartition(blockWithShard.shard);
         writer->write(blockWithShard.block);
     }
 }
@@ -314,17 +293,21 @@ void KafkaSink::checkpoint(CheckpointContextPtr context)
         {
             /// for a final check, it should not wait for too long
             if (auto err = rd_kafka_flush(producer.get(), 15000 /* time_ms */); err)
-                throw Exception(
-                    klog::mapErrorCode(err), "Failed to flush kafka producer, error={}", rd_kafka_err2str(err));
+                throw Exception(klog::mapErrorCode(err), "Failed to flush kafka producer, error={}", rd_kafka_err2str(err));
 
             if (auto err = wb->lastSeenError(); err != RD_KAFKA_RESP_ERR_NO_ERROR)
                 throw Exception(
-                    klog::mapErrorCode(err), "Failed to send messages, error_cout={} last_error={}", wb->error_count(), rd_kafka_err2str(err));
-
+                    klog::mapErrorCode(err),
+                    "Failed to send messages, error_cout={} last_error={}",
+                    wb->error_count(),
+                    rd_kafka_err2str(err));
 
             if (!wb->hasNoOutstandings())
                 throw Exception(
-                    ErrorCodes::CANNOT_WRITE_TO_KAFKA, "Not all messsages are sent successfully, expected={} actual={}", wb->outstandings(), wb->acked());
+                    ErrorCodes::CANNOT_WRITE_TO_KAFKA,
+                    "Not all messsages are sent successfully, expected={} actual={}",
+                    wb->outstandings(),
+                    wb->acked());
 
             break;
         }
@@ -336,4 +319,3 @@ void KafkaSink::checkpoint(CheckpointContextPtr context)
     IProcessor::checkpoint(context);
 }
 }
-
--- a/src/Storages/ExternalStream/Kafka/KafkaSink.cpp
+++ b/src/Storages/ExternalStream/Kafka/KafkaSink.cpp
@@ -38,7 +38,11 @@ ChunkPartitioner::ChunkPartitioner(ContextPtr context, const Block & header, con
     if (auto * shard_func = partitioning_expr_ast->as<ASTFunction>())
     {
         if (shard_func->name == "rand" || shard_func->name == "RAND")
+        {
             random_partitioning = true;
+            std::random_device r;
+            rand = std::minstd_rand(r());
+        }
     }
 }
 
--- a/src/Storages/ExternalStream/Kafka/KafkaSink.cpp
+++ b/src/Storages/ExternalStream/Kafka/KafkaSink.cpp
@@ -1,7 +1,14 @@
 #include "KafkaSink.h"
+#include <rdkafka.h>
 
-#include <Processors/Formats/IOutputFormat.h>
 #include <Common/logger_useful.h>
+#include <DataTypes/DataTypeLowCardinality.h>
+#include <DataTypes/DataTypesNumber.h>
+#include <Interpreters/ExpressionAnalyzer.h>
+#include <Interpreters/TreeRewriter.h>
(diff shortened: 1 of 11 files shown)


src/Storages/ExternalStream/StorageExternalStream.cpp
```diff
@@ -25,18 +25,42 @@ namespace ErrorCodes
...
 extern const int NOT_IMPLEMENTED;
 extern const int BAD_ARGUMENTS;
+extern const int INCORRECT_NUMBER_OF_COLUMNS;
+extern const int TYPE_MISMATCH;
 }
 
 namespace
 {
+ExpressionActionsPtr
+buildShardingKeyExpression(const ASTPtr & sharding_key, ContextPtr context, const NamesAndTypesList & columns)
```
chenziliang: `ASTPtr sharding_expr`

src/Storages/ExternalStream/StorageExternalStream.cpp
```diff
@@ -25,18 +25,42 @@ namespace ErrorCodes
...
+{
+    ASTPtr query = sharding_key;
+    auto syntax_result = TreeRewriter(context).analyze(query, columns);
+    return ExpressionAnalyzer(query, syntax_result, context).getActions(true);
+}
+
+void validateEngineArgs(ContextPtr context, ASTs & engine_args, const ColumnsDescription & columns) {
+    const auto & sharding_expr_arg = engine_args[0];
```
chenziliang: We can eliminate this temp variable by inlining it

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.h
```diff
@@ -30,6 +31,9 @@ class WriteBufferFromKafka final : public BufferWithOwnMemory<WriteBuffer>
     /// allows to reset the state after each checkpoint
     void resetState() { state.reset(); }
 
+    /// Set the partition ID the buffer will write data to.
+    /// This makes it possible write data to different paritions.
+    void write_to_partition(Int32 id) { partition_id = id; }
```
chenziliang: `setTargetPartition(Int32 id)` may fit the convention better.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -20,15 +21,21 @@ void WriteBufferFromKafka::nextImpl()
     if (!offset())
         return;
 
+    /// `write_to_topic` should have been called before the buffer starts to work.
```
chenziliang: Do we have `write_to_topic` ?

chenziliang: We can assert in ctor ?

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -1,5 +1,8 @@
 #pragma once
 
+#include <random>
```
chenziliang: Let's include this header as the last segment of the include headers

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -1,7 +1,14 @@
 #include "KafkaSink.h"
+#include <rdkafka.h>
```
chenziliang: let's include this along with 3rd part header section like boost

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -10,12 +17,112 @@ namespace DB
...
+    /// `InterpreterCreateQuery::handleExternalStreamCreation` ensures this
+    assert(partitioning_expr_ast);
+
+    ASTPtr query = partitioning_expr_ast;
+    auto syntax_result = TreeRewriter(context).analyze(query, header.getNamesAndTypesList());
+    partitioning_expr = ExpressionAnalyzer(query, syntax_result, context).getActions(false);
+
+    partitioning_key_column_name = partitioning_expr_ast->getColumnName();
+
+    if (auto * shard_func = partitioning_expr_ast->as<ASTFunction>())
```
chenziliang, line 38: Let's put a outer curly braces here
```
{
}
```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -10,12 +17,112 @@ namespace DB
...
+        if (shard_func->name == "rand" || shard_func->name == "RAND")
+        {
+            random_partitioning = true;
+            std::random_device r;
```
chenziliang, line 43: Is there dangling object here ? `rand` object ref a object `r()` which has been destructed ? We may need consider reusing pseudo generator as well meaning reuse at least at Stream level if not at global level.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -10,12 +17,112 @@ namespace DB
...
+            rand = std::minstd_rand(r());
+        }
+}
+
+BlocksWithShard ChunkPartitioner::partition(Block block, Int32 partition_cnt) const
+{
+    /// no topics have zero partitions
+    assert(partition_cnt > 0);
+    /// check the ctor for how this is guaruanteed
+    assert(partitioning_expr);
```
chenziliang: Once we assert in ctor, we are good. So no need to assert here since it won't regret ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -10,12 +17,112 @@ namespace DB
...
+        partitioned_blocks[i] = block.cloneEmpty();
+
+    for (size_t pos = 0; pos < block.columns(); ++pos)
+    {
+        MutableColumns partitioned_columns = block.getByPosition(pos).column->scatter(partition_cnt, selector);
+        for (Int32 i = 0; i < partition_cnt; ++i)
+            partitioned_blocks[i].getByPosition(pos).column = std::move(partitioned_columns[i]);
+    }
+
+    BlocksWithShard blocks_with_shard;
```
chenziliang, line 79: We may need reserve here : blocks_with_shard.reserve(partitioned_blocks.size())

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -10,12 +17,112 @@ namespace DB
...
+    for (size_t i = 0; i < partitioned_blocks.size(); ++i)
+    {
+        if (partitioned_blocks[i].rows())
+            blocks_with_shard.emplace_back(std::move(partitioned_blocks[i]), i);
+    }
+
+    return blocks_with_shard;
+}
+
+IColumn::Selector ChunkPartitioner::createSelector(const Block & block, Int32 partition_cnt) const
```
chenziliang: Let's do the following, and remove `current_block`

```
ChunkPartitioner::createSelector(Block block, Int32 partition_cnt)
```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -10,12 +17,112 @@ namespace DB
...
+    return doParition(std::move(block), partition_cnt);
 }
 
-KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr context, Poco::Logger * log_)
-    : SinkToStorage(header, ProcessorID::ExternalTableDataSinkID), producer(nullptr, nullptr), polling_threads(1), log(log_)
+BlocksWithShard ChunkPartitioner::doParition(Block block, Int32 partition_cnt) const
+{
+    auto selector = createSelector(block, partition_cnt);
+
+    Blocks partitioned_blocks{static_cast<size_t>(partition_cnt)};
```
chenziliang, line 67:
```
Blocks partitioned_blocks;
partitioned_blocks.reserve(partition_cnt);

for (...)
    partitioned_blocks.emplace_back(block.cloneEmpty());
```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -157,10 +310,30 @@ void KafkaSink::checkpoint(CheckpointContextPtr context)
...
+            /// for a final check, it should not wait for too long
+            if (auto err = rd_kafka_flush(producer.get(), 15000 /* time_ms */); err)
+                throw Exception(
+                    klog::mapErrorCode(err), "Failed to flush kafka producer, error={}", rd_kafka_err2str(err));
+
+            if (auto err = wb->lastSeenError(); err != RD_KAFKA_RESP_ERR_NO_ERROR)
+                throw Exception(
+                    klog::mapErrorCode(err), "Failed to send messages, error_cout={} last_error={}", wb->error_count(), rd_kafka_err2str(err));
+
+
```
chenziliang, line 309: extra empty line

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -78,8 +186,41 @@ KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr conte
...
+        topic_conf = rd_kafka_topic_conf_new();
+        /// conf will take the ownership of topic_conf
+        rd_kafka_conf_set_default_topic_conf(conf, topic_conf);
+    }
+
+    rd_kafka_conf_set_opaque(conf, this);
+    rd_kafka_topic_conf_set_opaque(topic_conf, this);
+
+    /// With partitioner callback, we can get the up-to-date partition count w/o additional effort.
+    rd_kafka_topic_conf_set_partitioner_cb(topic_conf, [](const rd_kafka_topic_t * /*rkt*/,
```
chenziliang: This look another dangling lambda object which get dtor-ed and still referenced in C lib. We will need hold on to the lambda callback or refactor it out as a free function just as we have done in `KafkaLog` for all callbacks.

Cross language / lib  memory passing deserves great attentions.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -78,8 +186,41 @@ KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr conte
...
+						void * msg_opaque) -> int32_t
+    {
+        /// update partition count
+        auto * sink = static_cast<KafkaSink *>(rkt_opaque);
+        sink->partition_cnt = partition_count;
+
+        auto * partition_id_ptr = static_cast<Int32 *>(msg_opaque);
+        auto parition_id =  *partition_id_ptr;
+        delete partition_id_ptr;
+        /// This should not really happen because Kafka does not support reducing partitions.
```
chenziliang: If partitioned gets changed in the middle, this partition assignment make not make sense anymore, but in general, the sharding strategy might not make sense any more.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -20,15 +21,21 @@ void WriteBufferFromKafka::nextImpl()
...
+        topic,
+        /// we want to trigger the partitioner function, check KafkaSink.cpp
         RD_KAFKA_PARTITION_UA,
         RD_KAFKA_MSG_F_COPY | RD_KAFKA_MSG_F_BLOCK,
         working_buffer.begin(),
         offset(),
-        nullptr /* key */,
-        0 /* keylen */,
-        static_cast<void *>(this) /* opaque */);
+        "unused" /* key, even though we don't use the key, but it's needed to trigger the partitioner_cb */,
```
chenziliang: Can we just use empty string like "" and keylen = 0 ? If no, use one byte string.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -20,15 +21,21 @@ void WriteBufferFromKafka::nextImpl()
     if (!offset())
         return;
 
+    /// `write_to_topic` should have been called before the buffer starts to work.
+    assert(topic);
+
+    auto * part_id = new Int32(partition_id);
```
chenziliang: We don't need allocate, just convert partition_id to a `void *` , like 

```
auto err = rd_kafka_produce(
    ...
    reinterpret_cast<void *>(partition_id) /* opaque */);
```

And then convert it back in KafkaSink

```
auto partition_id = reinterpret_cast<Int32>(opaque)
```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Actually this is not needed at all, it's added by the language server, need to check how to fix this annoying behavior.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: I was thinking this method has no knowledge what ctor does, thus added this assertion here as well. But as this is initialized by ctor and it's a private field, it should be safe to remove the assertion here.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: First, we want to make sure things won't break, for example, MVs won't suddently stop working when parittions change. I think the current implement should be a good default behavior. We can propably give users more control on the behavior in the future when it's requested.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
zliang-min: This comment and assert is not valid any more, I refactored the code, but I didn't remove the comment.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
zliang-min: It complains `Cast from pointer to smaller type 'Int32' (ask 'int') loses information [bad_reinterpret_cast_small_int]`.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: No, there is nothing to leak, `r()` just returns an integer (unsigned to be precise).

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
zliang-min: If we do this
```c++
auto parition_id_from_ptr = reinterpret_cast<std::uintptr_t>(msg_opaque);
auto partition_id = static_cast<Int32>(partition_id_from_ptr);
```

We can avoid the error.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
zliang-min: Yes, empty string actually works, updated.

src/Storages/ExternalStream/StorageExternalStream.cpp
zliang-min: I used `ASTPtr` at the first place. But I found that there are a lot places use `const ASTPtr &`, in fact, `StorageStream.cpp` uses `const ASTPtr &`, thus I changed it. When it should use `const ASTPtr &`? In this same file, line 132 below, in the `StorageExternalStream` ctor, I also used `const ASTPtr &`, should I change it too?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: I am still getting used to c++ :)

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -31,12 +31,13 @@ class ChunkPartitioner
 
     BlocksWithShard doParition(Block block, Int32 partition_cnt) const;
 
-    IColumn::Selector createSelector(const Block & block, Int32 partition_cnt) const;
+    IColumn::Selector createSelector(Block block, Int32 partition_cnt) const;
+
+    static inline std::minstd_rand rand{std::random_device()()};
```
chenziliang: std::minstd_rand is usually not thread-safe, static object will be shared across different KafkaSink (hence likely multiple threads)

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -53,6 +54,32 @@ class KafkaSink final : public SinkToStorage
...
+        int32_t partition_count,
+        void * rkt_opaque,
+        void * msg_opaque)
+    {
+        /// update partition count
+        auto * sink = static_cast<KafkaSink *>(rkt_opaque);
+        sink->partition_cnt = partition_count;
+
+        auto partition_id_ptr = reinterpret_cast<std::uintptr_t>(msg_opaque);
+        auto parition_id = static_cast<Int32>(partition_id_ptr);
```
chenziliang: nit : `auto partition_id = static_cast<Int32>(reinterpret_cast<std::uintptr_t>(msg_opaque));`

src/Storages/ExternalStream/Kafka/WriteBufferFromKafka.cpp
```diff
@@ -21,21 +20,16 @@ void WriteBufferFromKafka::nextImpl()
...
         topic,
         /// we want to trigger the partitioner function, check KafkaSink.cpp
         RD_KAFKA_PARTITION_UA,
         RD_KAFKA_MSG_F_COPY | RD_KAFKA_MSG_F_BLOCK,
         working_buffer.begin(),
         offset(),
-        "unused" /* key, even though we don't use the key, but it's needed to trigger the partitioner_cb */,
-        6 /* keylen */,
-        static_cast<void *>(part_id) /* opaque */);
+        "" /* key, even though we don't use the key, but it's needed to trigger the partitioner_cb */,
```
chenziliang: The convention is

```
/*key=*/ "", /// Even though we don't use the key, but it's needed to trigger the partitioner_cb
/*opaque=*/ reinterpret_cast<void *>(partition_id));
```

src/Storages/ExternalStream/StorageExternalStream.cpp
chenziliang: This is a v good question. First of all, always stick to `const ASTPtr &` whenever possible since it doesn't need to bump up the atomic ref count (which is always not good across the CPU cores). In this case, we shall stick to `const ASTPtr & ` if the `TreeRewrtier` / `ExpressionAnalzyer` accept a const ref. If they don't, for this `buildShardingKeyExpression(...)` specifically, it is good to just accept `ASTPtr` like we did here after change, it simplify caller and as well as the callee.

chenziliang: This is a general advice / guideline for `ContextPtr` etc as well.

src/Storages/ExternalStream/Kafka/KafkaSink.h
zliang-min: As discussed, reverted it to per KafkaSink for now, and will introduce a global random generator later.

chenziliang (approved):

yokofly:
follow up: https://github.com/timeplus-io/proton/pull/290

jovezhong:

jovezhong:

PR #277, opened 2023-11-14
make random stream eps be query time settings

### LINKED ISSUE #260: Random stream EPS etc could be query time settings instead of bounding to storage
**Use case**

For now eps etc is bound to a storage definition, if i need different eps, i will need create different random streams. It will be way convenient to make them query time. 

```
select * from my_random_stream settings eps=1000000;
```

**Additional context**

lizhou1111:
this close #260 
```sql
create random stream test22_create_random(id int default rand()%4) engine Random() settings eps=1000;
select count(1) from test22_create_random settings generate_eps=100;
```
If you set `generate_eps` in query time, it will use `generate_eps` as eps to generate data.
PR checklist:
- Did you run ClangFormat ?

### DIFF
--- a/tests/stream/test_stream_smoke/0021_random_stream.json
+++ b/tests/stream/test_stream_smoke/0021_random_stream.json
@@ -374,6 +374,50 @@
                 ]
               }
             ]
+          },
+          {
+            "id": 17,
+            "tags": ["random stream table query"],
+            "name": "test random stream table query, set max_events=1000",
+            "description": "test random stream table query",
+            "steps":[
+              {
+                "statements": [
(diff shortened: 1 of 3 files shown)


tests/stream/test_stream_smoke/0021_random_stream.json
yl-lisen, line 1: missing cases covering setting `generate_eps`

lizhou1111: ok i will add it.

yl-lisen (approved):

PR #302, opened 2023-11-20
rename generate_eps to eps

### LINKED ISSUE #301: rename random stream query time eps settings and move some settings to proper place

lizhou1111:
this close #301 
rename `genarate_eps` to `eps`  and move some settings to proper place.
```sql
create random stream test22_create_random(id int default rand()%4) engine Random() settings eps=1000;
select count(1) from test22_create_random settings eps=100;
```


yl-lisen (approved):
LGTM

PR #307, opened 2023-11-20
supporting select partial columns from external kafka stream

qijun-niu-timeplus:
PR checklist:
- Did you run ClangFormat ?

### DIFF
--- a/src/Storages/ExternalStream/Kafka/KafkaSource.cpp
+++ b/src/Storages/ExternalStream/Kafka/KafkaSource.cpp
@@ -43,6 +43,7 @@ KafkaSource::KafkaSource(
     , max_block_size(max_block_size_)
     , log(log_)
     , header(header_)
+    , non_virtual_header(storage_snapshot->metadata->getSampleBlockNonMaterialized())
     , consume_ctx(kafka->topic(), shard, offset)
     , read_buffer("", 0)
     , virtual_time_columns_calc(header.columns(), nullptr)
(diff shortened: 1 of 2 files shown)


src/Storages/ExternalStream/Kafka/KafkaSource.h
```diff
@@ -62,9 +62,12 @@ class KafkaSource final : public ISource
     Poco::Logger * log;
 
     Block header;
+    const Block non_virtual_header;
```
zliang-min, line 65: This is only used for getting the input format, no need to make it a member variable?

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
```diff
@@ -272,10 +276,17 @@ void KafkaSource::initFormatExecutor(const Kafka * kafka)
...
         format_executor = std::make_unique<StreamingFormatExecutor>(
-            physical_header, std::move(input_format), [](const MutableColumns &, Exception &) -> size_t { return 0; });
+            non_virtual_header, std::move(input_format), [](const MutableColumns &, Exception &) -> size_t { return 0; });
+
+        auto converting_dag = ActionsDAG::makeConvertingActions(
+            non_virtual_header.cloneEmpty().getColumnsWithTypeAndName(),
+            physical_header.cloneEmpty().getColumnsWithTypeAndName(),
+            ActionsDAG::MatchColumnsMode::Name);
+
+        convert_non_virtual_to_physical_action = std::make_shared<ExpressionActions>(std::move(converting_dag));
```
zliang-min, line 289: This logic has nothing to do with the format executor, extract them into another method, or simply put them in ctor?

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
```diff
@@ -178,24 +179,29 @@ void KafkaSource::parseRaw(const rd_kafka_message_t * kmessage)
 void KafkaSource::parseFormat(const rd_kafka_message_t * kmessage)
 {
     assert(format_executor);
+    assert(convert_non_virtual_to_physical_action);
```
zliang-min, line 182: Since it's initialized in ctor, probably no need to assert?

src/Storages/ExternalStream/Kafka/KafkaSource.h
zliang-min: I didn't notice it's also used in `parseFormat`.

zliang-min (approved):
Some small comments, no blocking issues I could find.

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
qijun-niu-timeplus: This is binding with format. In this version, formatter will parse message to the full columns set which `create` statement specified. And this dag action will choose a subset of columns which is needed by the following processors. Thus, when we use a format like csv, it can parse the data as what we want to do.
If it's only a raw data column, we don't need to init this dag action.

PR #308, opened 2023-11-20
Update TelemetryCollector.h, 1hour to 2min

jovezhong:
sending telemetry data every 2min to get more query stats during server session.

Currently the telemetry data is sent every 1 hour, but if the user stops proton within 1 hour, we cannot get such stats, such as how long they run the server. Changing this to every 2 min and we will check whether it's too much

### DIFF
--- a/programs/server/TelemetryCollector.h
+++ b/programs/server/TelemetryCollector.h
@@ -19,7 +19,7 @@ private:
     bool new_session = true;
     Int64 started_on_in_minutes;
 
-    static constexpr auto INTERVAL_MS = 60 * 60 * 1000; /// 1 hour
+    static constexpr auto INTERVAL_MS = 2 * 60 * 1000; /// sending anonymous telemetry data every 2 minutes
 
 public:
     static TelemetryCollector & instance(ContextPtr context_)


programs/server/TelemetryCollector.h
```diff
@@ -19,7 +19,7 @@ class TelemetryCollector
     bool new_session = true;
     Int64 started_on_in_minutes;
 
-    static constexpr auto INTERVAL_MS = 60 * 60 * 1000; /// 1 hour
```
chenziliang, line 22: Let's make it configurable, otherwise we will keep updating this interval since someday we may find 1) too much volume, 2) don't worth this frequency 3) users complain the frequency

qijun-niu-timeplus (approved):

programs/server/TelemetryCollector.h
jovezhong: Let me merge this first and @qijun-niu-timeplus , please move this to config file in the other PR

jovezhong:
Create the ticket https://github.com/timeplus-io/proton/issues/315

PR #309, opened 2023-11-20
fix: time-based seek_to does not work with Kafka external streams

### LINKED ISSUE #287: using seek_to with specify timestamp in external stream  can't work
```sql
-- redpanda
-- rpk topic create test -p 3

-- proton1 
drop stream if exists test_external1;
create external stream if not exists test_external1(id int, value string) settings type = 'kafka', brokers = 'redpanda-1:9092', topic = 'test', data_format='JSONEachRow';

-- redpanda
-- rpk topic produce test
{"id":1,"value":"sometext"}
Produced to partition 0 at offset 2 with timestamp 1700015474271.
{"id":2,"value":"aaaaa"}
Produced to partition 2 at offset 1 with timestamp 1700015492170.
{"id":3,"value":"bbbbb"}
Produced to partition 1 at offset 1 with timestamp 1700015510566.
{"id":4,"value":"ccccc"}
Produced to partition 2 at offset 2 with timestamp 1700015528830.
-- proton
select * from test_external1 settings seek_to='2023-11-11 14:15:24.219';

```

zliang-min:
PR checklist:
- Did you run ClangFormat ?

Because data selected from Kafka external streams do not have `_tp_time`, the added predicate for time-based seek_to does not work. Even if we added `_tp_time` to Kafka external streams data, it would not make sense to add such predicate to the queries because Kafka already supports this feature natively, adding the predicate just wastes resources.

closes #287

### DIFF
--- a/src/Interpreters/InterpreterSelectQuery.cpp
+++ b/src/Interpreters/InterpreterSelectQuery.cpp
@@ -3604,11 +3604,11 @@ void InterpreterSelectQuery::handleSeekToSetting()
         if (seek_points.size() != 1)
             throw Exception(ErrorCodes::NOT_IMPLEMENTED, "It doesn't support time based `seek_to` settings for multiple shards");
 
-        /// If the storage does not have built-in seek_to support (for example, external streams like Kafka support seek_to, via offset, natively)
-        /// we rewrite WHERE predicates of SELECT query.
-        /// Example : SELECT * FROM stream SETTINGS seek_to='2022-01-01 00:01:01' =>
-        /// SELECT * FROM stream WHERE _tp_time >= '2022-01-01 00:01:01'
-        if (storage && !storage->supportsNativeSeekTo())
+        /// If the storage can do accurate seek_to (for example, Kafka external streams) no extra work is needed.
+        /// Otherwise, the WHERE predicates of SELECT query will be rewritten by adding a filter for filtering
+        /// records by `_tp_time`. For example: `SELECT * FROM stream SETTINGS seek_to='2022-01-01 00:01:01'` will
+        /// be rewritten to `SELECT * FROM stream WHERE _tp_time >= '2022-01-01 00:01:01'`.
+        if (storage && !storage->supportsAccurateSeekTo())
             addEventTimePredicate(getSelectQuery(), seek_points[0]);
     }
     else
--- a/src/Interpreters/InterpreterSelectQuery.cpp
+++ b/src/Interpreters/InterpreterSelectQuery.cpp
@@ -3608,7 +3608,7 @@ void InterpreterSelectQuery::handleSeekToSetting()
         /// we rewrite WHERE predicates of SELECT query.
         /// Example : SELECT * FROM stream SETTINGS seek_to='2022-01-01 00:01:01' =>
         /// SELECT * FROM stream WHERE _tp_time >= '2022-01-01 00:01:01'
-        if (storage && !storage->builtinSeekToSupport())
+        if (storage && !storage->supportsNativeSeekTo())
             addEventTimePredicate(getSelectQuery(), seek_points[0]);
     }
     else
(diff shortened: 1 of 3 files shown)


src/Storages/ExternalStream/StorageExternalStream.h
```diff
@@ -22,6 +22,7 @@ class StorageExternalStream final : public shared_ptr_helper<StorageExternalStre
     void shutdown() override;
     bool supportsSubcolumns() const override;
     bool squashInsert() const noexcept override { return false; }
+    bool builtinSeekToSupport() const noexcept override { return true; }
```
chenziliang: `supportsSeekTo` to keep the convention.

src/Interpreters/InterpreterSelectQuery.cpp
```diff
@@ -3604,10 +3604,12 @@ void InterpreterSelectQuery::handleSeekToSetting()
...
             throw Exception(ErrorCodes::NOT_IMPLEMENTED, "It doesn't support time based `seek_to` settings for multiple shards");
 
-        /// Here we rewrite WHERE predicates of SELECT query.
+        /// If the storage does not have built-in seek_to support (for example, external streams like Kafka support seek_to, via offset, natively)
+        /// we rewrite WHERE predicates of SELECT query.
         /// Example : SELECT * FROM stream SETTINGS seek_to='2022-01-01 00:01:01' =>
         /// SELECT * FROM stream WHERE _tp_time >= '2022-01-01 00:01:01'
-        addEventTimePredicate(getSelectQuery(), seek_points[0]);
+        if (storage && !storage->builtinSeekToSupport())
+            addEventTimePredicate(getSelectQuery(), seek_points[0]);
```
chenziliang, line 3612: 1) `seek_to` a timestamp is translated a seek and a filtering. We may need do same thing for external stream.
2) StorageStream support seek to by timestamp as well ?

chenziliang: @yl-lisen thoughts ?

zliang-min: For 1, it depends, if the storage does not support seek to a timestamp, it might need a filter (sometimes, it might not even possible). But for a storage like Kafka, there is no need to do another seek and filter, it's just wasting resources.

zliang-min: So, if later on, we found a external stream that needs to apply the filtering, we can't simply use `storage->builtinSeekToSupport`, we will need to defer it to the storage type level.

src/Storages/ExternalStream/StorageExternalStream.h
zliang-min: `supportsSeekTo` is confusing. For example, when `supportsSeekTo` equals `false`, it sounds it does not support seek to at all, i.e. it should throw an error if seek to is used. If we want to keep the naming convention, we should probably name it `supportsNativeSeekTo`.

zliang-min: I renamed it to `supportsNativeSeekTo`, please reopen this comment if you think that's not a proper name.

src/Interpreters/InterpreterSelectQuery.cpp
yl-lisen: For seek to by timestamp, the event time predicates is added since nativelog seek to by timestamp is not accurate.
for `kafkaSource`, we called `rd_kafka_offsets_for_times`, it also documented:
```
 * @brief Look up the offsets for the given partitions by timestamp.
 *
 * The returned offset for each partition is the earliest offset whose
 * timestamp is greater than or equal to the given timestamp in the
 * corresponding partition.
```
@zliang-min

yl-lisen: Based on above description, it may indeed be accurate for  kafka source

yl-lisen: I think the method name `supportsNativeSeekTo` is not good, we can use `supportAccurateSeekTo` or something else

zliang-min: Thanks @yl-lisen for the context of why the filter was adding at the first place. Based on that, I think `supportsAccurateSeekTo` is a better name. Updated.

yl-lisen (approved):

src/Interpreters/InterpreterSelectQuery.cpp
chenziliang: Kafka timestamp based seek is NOT accurate, it never does. So we shall filter further. It is not waste of resources.

jovezhong:

jovezhong:

PR #347, opened 2023-11-27
Feature/issue 344 asof join memory efficiency

chenziliang:
PR checklist:
- Did you run ClangFormat ?

### DIFF
--- a/src/Core/Settings.h
+++ b/src/Core/Settings.h
@@ -792,7 +792,8 @@ static constexpr UInt64 operator""_GiB(unsigned long long value)
     M(Bool, enable_backfill_from_historical_store, true, "Enable backfill data from historical data store", 0) \
     M(Bool, emit_aggregated_during_backfill, true, "Enable emit intermediate aggr result during backfill historical data", 0) \
     M(Bool, include_internal_streams, false, "Show internal streams on SHOW streams query.", 0) \
-    M(UInt64, join_max_cached_bytes, 524288000, "Max cached bytes for stream to stream join", 0) \
+    M(UInt64, join_max_buffered_bytes, 524288000, "Max buffered bytes for stream to stream join", 0) \
+    M(UInt64, join_buffered_data_block_size, 0, "For streaming join, when buffered data in memory, the data block size directs to merge small data blocks to form bigger ones to improve memory efficiency. 0 means disable merging small data blocks.", 0) \
     M(Int64, max_join_range, 300, "Max join range", 0) \
     M(Bool, compact_kv_stream, true, "Control if compact a changelog kv or versioned kv stream during query", 0) \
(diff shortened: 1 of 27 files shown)


chenziliang:
Micro-benchmark shows 10x memory reduction for single row per chunk case.

src/Core/Settings.h
```diff
@@ -792,7 +792,8 @@ static constexpr UInt64 operator""_GiB(unsigned long long value)
     M(Bool, enable_backfill_from_historical_store, true, "Enable backfill data from historical data store", 0) \
     M(Bool, emit_aggregated_during_backfill, true, "Enable emit intermediate aggr result during backfill historical data", 0) \
     M(Bool, include_internal_streams, false, "Show internal streams on SHOW streams query.", 0) \
-    M(UInt64, join_max_cached_bytes, 524288000, "Max cached bytes for stream to stream join", 0) \
+    M(UInt64, join_max_buffered_bytes, 524288000, "Max buffered bytes for stream to stream join", 0) \
+    M(UInt64, join_buffered_data_block_size, 0, "For streaming join, when buffered data in memory, the data block size directs to merge small data blocks to form bigger ones to improve memory efficiency. 0 means disable merging small data blocks.", 0) \
```
yl-lisen, line 796: I prefer setting name like this `join_max_rows_for_compact_buffered_data`

src/Interpreters/Streaming/benchmarks/data_blocks_perf.cpp
```diff
@@ -296,111 +296,134 @@ BENCHMARK_CAPTURE(
...
+    /*data_block_size=*/0)
+    ->Iterations(1);
+
+BENCHMARK_CAPTURE(
+    refCountDataBlockList,
+    Column3ChunkRows1Chunks10000000BS1000,
+    /*chunk_columns=*/3,
+    /*chunk_rows=*/1,
+    /*chunks*=*/10'000'000,
+    /*data_block_size=*/1000)
```
yl-lisen, line 345: Here we only test one `data_block_size = 1000`
For rows = 1, it seems that the larger the `data_block_size`, the smaller the memory cost.

src/Core/Settings.h
chenziliang: We follow the 'max_block_size' convention here.

src/Interpreters/Streaming/benchmarks/data_blocks_perf.cpp
chenziliang: In general, this is true. And we can emulate other combinations in this test and there are lots of combinations. 1000 is just a good starting point for this case. And this is another reason the data block size setting is exposed to end users for now.

PR #352, opened 2023-11-28
feat: support configuring message.max.bytes for kafka external streams

### LINKED ISSUE #335: when huge block size exceeds the max message size of kafka/redpanda, the huge block  can't insert into external stream
huge block can't insert into external stream.
Below is the python script to insert buge block into redpanda.
```python
from proton_driver import client

one_data = [1,2,3,4]

# proton
# drop stream if exists test_stream2;
# create external stream if not exists test_stream2(
# 	a int8,
# 	b int16,
# 	c int32,
# 	d int64
# 	)settings type = 'kafka', brokers = 'redpanda-1:9092', topic = 'test',data_format='JSONEachRow' ;

def GenerateCase(count):
    rows = []
    for i in range(count):
        
        row = one_data.copy()
        row[3] = i
        # to tuple
        row = tuple(row)
        rows.append(row)
    
    return rows

if __name__ == "__main__":
    client = client.Client(host="127.0.0.1", port=8463)
    rows = GenerateCase(100000)
    # print(rows)
    client.execute("insert into test_stream2(a, b, c, d) values", rows)
```
The error message is `proton_driver.errors.ServerException: Code: 2600.
DB::Exception: Cannot write to kafka topic at offset 2097152, error: Broker: Message size too large. Stack trace:`

zliang-min:
PR checklist:
- Did you run ClangFormat ?

Now this is possible:
```sql
CREATE EXTERNAL STREAM my_stream (
...
) SETTINGS properties='message.max.bytes=1234'
```

This allows limiting the size of messages which are sent to the topic, or makes it possible to send big messages ( by default the max bytes is 1000000 ). Valid sizes are 1000 .. 1000000000 (refs: https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md).

NOTE: the message size limit configured on the topic has higher priority, i.e. setting `message.max.bytes` to a value that is bigger than the topic limit is useless.

fixes #335

### DIFF
--- a/src/Storages/ExternalStream/Kafka/KafkaSink.cpp
+++ b/src/Storages/ExternalStream/Kafka/KafkaSink.cpp
@@ -141,6 +141,7 @@ KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr conte
         "queue.buffering.max.messages",
         "queue.buffering.max.kbytes",
         "queue.buffering.max.ms",
+        "message.max.bytes",
         "message.send.max.retries",
         "retries",
         "retry.backoff.ms",


jovezhong:
For JSONEachRow and one JSON per messsage to false, is Proton going to try to create biggest possible message in many JSON without exceeding the `message.max.bytes`? Or this setting is mainly designed for super large JSON per message?

zliang-min:
> For JSONEachRow and one JSON per messsage to false, is Proton going to try to create biggest possible message in many JSON without exceeding the `message.max.bytes`? Or this setting is mainly designed for super large JSON per message?

This property only lifts the message size limitation on the Kafka producer side, it has nothing to do with how proton buffers/batches data. We don't do anything to make sure the messages sent to Kafka are under this size limit. If a mesage (no matter how many rows that message contains) is over the limit, it won't be sent. People can always use the `one_message_per_row` setting to make sure one message contains only one row of data, however, one row of data can still be over the size limit.

qijun-niu-timeplus (approved):

jovezhong:

jovezhong:

PR #353, opened 2023-11-28
feature/support join alignment in general phase 1

yl-lisen:
- support join alignment in general phase 1
- (TODO) more separate different hash joins works.

### DIFF
--- a/src/Core/Settings.h
+++ b/src/Core/Settings.h
@@ -794,12 +794,11 @@ static constexpr UInt64 operator""_GiB(unsigned long long value)
     M(Bool, include_internal_streams, false, "Show internal streams on SHOW streams query.", 0) \
     M(UInt64, join_max_buffered_bytes, 524288000, "Max buffered bytes for stream to stream join", 0) \
     M(UInt64, join_buffered_data_block_size, 0, "For streaming join, when buffered data in memory, the data block size directs to merge small data blocks to form bigger ones to improve memory efficiency. 0 means disable merging small data blocks.", 0) \
-    M(Int64, join_quiesce_threshold_ms, 0, "For streaming join (only works on join alignment), when left or right stream is in quiesce, the maximum time to wait before the join.", 0) \
+    M(Int64, join_quiesce_threshold_ms, 0, "For streaming join, when left or right stream is in quiesce, the maximum time to wait before the join.", 0) \
     M(Int64, max_join_range, 300, "Max join range", 0) \
     M(Bool, compact_kv_stream, true, "Control if compact a changelog kv or versioned kv stream during query", 0) \
     M(UInt64, keep_versions, 3, "Control how many versions for each key kept in memory when joining. Used in versioned_kv join", 0) \
-    M(Int64, join_latency_threshold, 0, "Control when to start join left stream with right stream (only works on join alignment). Zero means automatic system picked threshold", 0) \
-    M(Bool, enable_join_alignment, false, "Enable streaming join with alignment", 0) \
+    M(Int64, join_latency_threshold, 0, "Control when to start join left stream with right stream for alignment join scenarios. Zero means system picked threshold", 0) \
     M(Bool, enforce_append_only, false, "For changelog storage, enforce query it as append only storage", 0) \
     M(UInt64, retract_max, 10000, "Control how many more values to keep around for changelog processing to workaround retract scenarios in min/max etc aggr", 0) \
     M(UInt64, retract_k_multiplier, 2, "Control how many more values to keep around for changelog processing to workaround retract scenarios in min_k/max_k etc aggr", 0) \
--- a/src/Core/Settings.h
+++ b/src/Core/Settings.h
@@ -794,11 +794,12 @@ static constexpr UInt64 operator""_GiB(unsigned long long value)
     M(Bool, include_internal_streams, false, "Show internal streams on SHOW streams query.", 0) \
     M(UInt64, join_max_buffered_bytes, 524288000, "Max buffered bytes for stream to stream join", 0) \
     M(UInt64, join_buffered_data_block_size, 0, "For streaming join, when buffered data in memory, the data block size directs to merge small data blocks to form bigger ones to improve memory efficiency. 0 means disable merging small data blocks.", 0) \
-    M(Int64, join_quiesce_threshold_ms, 0, "For streaming join, when left or right stream is in quiesce, the maximum time to wait before the join.", 0) \
+    M(Int64, join_quiesce_threshold_ms, 0, "For streaming join (only works on join alignment), when left or right stream is in quiesce, the maximum time to wait before the join.", 0) \
     M(Int64, max_join_range, 300, "Max join range", 0) \
     M(Bool, compact_kv_stream, true, "Control if compact a changelog kv or versioned kv stream during query", 0) \
     M(UInt64, keep_versions, 3, "Control how many versions for each key kept in memory when joining. Used in versioned_kv join", 0) \
-    M(Int64, join_latency_threshold, 0, "Control when to start join left stream with right stream. Zero means automatic system picked threshold", 0) \
+    M(Int64, join_latency_threshold, 0, "Control when to start join left stream with right stream (only works on join alignment). Zero means automatic system picked threshold", 0) \
+    M(Bool, enable_join_alignment, false, "Enable streaming join with alignment", 0) \
     M(Bool, enforce_append_only, false, "For changelog storage, enforce query it as append only storage", 0) \
     M(UInt64, retract_max, 10000, "Control how many more values to keep around for changelog processing to workaround retract scenarios in min/max etc aggr", 0) \
     M(UInt64, retract_k_multiplier, 2, "Control how many more values to keep around for changelog processing to workaround retract scenarios in min_k/max_k etc aggr", 0) \
--- a/src/Interpreters/CollectJoinOnKeysVisitor.cpp
+++ b/src/Interpreters/CollectJoinOnKeysVisitor.cpp
@@ -90,44 +90,46 @@ void CollectJoinOnKeysMatcher::Data::asofToJoinKeys()
 void CollectJoinOnKeysMatcher::Data::addLagBehindKeys(
     const ASTPtr & left_ast, const ASTPtr & right_ast, JoinIdentifierPosPair table_pos, Int64 lag_interval)
 {
-    addAlignedWithKeys(left_ast, right_ast, table_pos);
+    if (isLeftIdentifier(table_pos.first) && isRightIdentifier(table_pos.second))
+    {
+        left_alignment_key = left_ast->clone();
+        right_alignment_key = right_ast->clone();
+    }
+    else if (isRightIdentifier(table_pos.first) && isLeftIdentifier(table_pos.second))
+    {
+        left_alignment_key = right_ast->clone();
+        right_alignment_key = left_ast->clone();
+    }
+    else
+    {
+        throw Exception(
+            ErrorCodes::INVALID_JOIN_ON_EXPRESSION,
+            "Expressions {} and {} are from the same stream but from different arguments of equal function in `lag_behind()` function",
+            queryToString(left_ast),
+            queryToString(right_ast));
+    }
 
     /// left stream lag behind right stream
     /// lag_behind(100ms, left_stream.ts, right_stream.ts)
     analyzed_join.setLagBehindInterval(lag_interval);
 }
 
-void CollectJoinOnKeysMatcher::Data::addAlignedWithKeys(
-    const ASTPtr & left_ast, const ASTPtr & right_ast, JoinIdentifierPosPair table_pos)
+void CollectJoinOnKeysMatcher::Data::addLagBehindKeys(Int64 lag_interval)
 {
-    if (isLeftIdentifier(table_pos.first) && isRightIdentifier(table_pos.second))
-    {
-        timestamp_left_key = left_ast->clone();
-        timestamp_right_key = right_ast->clone();
-    }
-    else if (isRightIdentifier(table_pos.first) && isLeftIdentifier(table_pos.second))
-    {
-        timestamp_left_key = right_ast->clone();
-        timestamp_right_key = left_ast->clone();
-    }
-    else
-    {
-        throw Exception(
-            ErrorCodes::INVALID_JOIN_ON_EXPRESSION,
-            "Expressions {} and {} are from the same stream but from different arguments of equal function in `lag_behind()` or `aligned_with()` function",
-            queryToString(left_ast),
-            queryToString(right_ast));
-    }
+    left_alignment_key = nullptr;
+    right_alignment_key = nullptr;
 
-    analyzed_join.setRequiredJoinAlignment();
+    /// left stream lag behind right stream
+    /// lag_behind(100ms, left_stream.ts, right_stream.ts)
+    analyzed_join.setLagBehindInterval(lag_interval);
 }
 
-void CollectJoinOnKeysMatcher::Data::timestampASTToKeys()
+void CollectJoinOnKeysMatcher::Data::alignmentKeysASTToKeys()
 {
-    if (!timestamp_left_key || !timestamp_right_key)
-        throw Exception("No lag_behind(...) ON section.", ErrorCodes::INVALID_JOIN_ON_EXPRESSION);
+    if (!left_alignment_key || !right_alignment_key)
+        return;
 
-    analyzed_join.addTimestampKeys(timestamp_left_key, timestamp_right_key);
+    analyzed_join.addAlignmentKeys(left_alignment_key, right_alignment_key);
 }
 
 /// proton : ends
@@ -159,13 +161,6 @@ void CollectJoinOnKeysMatcher::visit(const ASTFunction & func, const ASTPtr & as
     if (func.name == "lag_behind")
     {
         handleLagBehind(func, ast, data);
-        return;
-    }
-
-    /// `aligned_with` is not actually a join condition but interpreted as a stream modifier`
-    if (func.name == "aligned_with")
-    {
-        handleAlignedWith(func, ast, data);
         return;
     }
     /// proton : ends
@@ -681,71 +676,67 @@ void CollectJoinOnKeysMatcher::handleLagBehind(const ASTFunction & func, const A
 {
     assert(func.name == "lag_behind");
 
-    if (func.arguments->children.size() != 3)
-        throw Exception(ErrorCodes::SYNTAX_ERROR, "Function {} takes 3 arguments, but got '{}' instead", func.name, func.formatForErrorMessage());
+    if (func.arguments->children.size() != 1 && func.arguments->children.size() != 3)
+        throw Exception(ErrorCodes::SYNTAX_ERROR, "Function {} takes 1 or 3 arguments, but got '{}' instead", func.name, func.formatForErrorMessage());
 
-    /// lag_behind(2m, col1, col2)
-    ASTPtr lag_interval_ast = func.arguments->children.at(0);
-    ASTPtr left = func.arguments->children.at(1);
-    ASTPtr right = func.arguments->children.at(2);
-
-    Int64 lag_interval_unit = 0;
     Int64 lag_interval = 0;
-
-    if (auto * interval_func = lag_interval_ast->as<ASTFunction>())
+    /// 1) lag_behind(2ms): the lag unit is based on processing time.
+    if (func.arguments->children.size() == 1)
     {
-        if (interval_func->name == "to_interval_millisecond")
-            lag_interval_unit = 1;
-        else if (interval_func->name == "to_interval_second")
-            lag_interval_unit = 1000;
-        else if (interval_func->name == "to_interval_minute")
-            lag_interval_unit = 60 * 1000;
-        else
-            throw Exception(
-                ErrorCodes::NOT_IMPLEMENTED,
-                "First argument of function {} only supports millsecond, second or minute time interval (examples, 1ms, 1s, 2m etc), but got '{}' instead",
-                func.name, func.formatForErrorMessage());
+        Int64 lag_interval_unit = 0;
+        if (auto * interval_func = func.arguments->children.at(0)->as<ASTFunction>())
+        {
+            if (interval_func->name == "to_interval_millisecond")
+                lag_interval_unit = 1;
+            else if (interval_func->name == "to_interval_second")
+                lag_interval_unit = 1000;
+            else if (interval_func->name == "to_interval_minute")
+                lag_interval_unit = 60 * 1000;
+            else
+                throw Exception(
+                    ErrorCodes::NOT_IMPLEMENTED,
+                    "First argument of function {} only supports millsecond, second or minute time interval (examples, 1ms, 1s, 2m etc), but got '{}' instead",
+                    func.name, func.formatForErrorMessage());
 
-        auto * interval = interval_func->arguments->children.at(0)->as<ASTLiteral>();
-        assert(interval);
-        lag_interval = interval->value.get<Int64>() * lag_interval_unit;
-        if (lag_interval <= 0)
-            throw Exception(
-                ErrorCodes::BAD_QUERY_PARAMETER,
-                "First argument of function {} shall be positive millisecond, second or minute time interval (examples, 1ms, 2s, 3m etc), but got '{}' instead",
-                func.name, func.formatForErrorMessage());
+            auto * interval = interval_func->arguments->children.at(0)->as<ASTLiteral>();
+            assert(interval);
+            lag_interval = interval->value.get<Int64>() * lag_interval_unit;
+            if (lag_interval <= 0)
+                throw Exception(
+                    ErrorCodes::BAD_QUERY_PARAMETER,
+                    "First argument of function {} shall be positive millisecond, second or minute time interval (examples, 1ms, 2s, 3m etc), but got '{}' instead",
+                    func.name, func.formatForErrorMessage());
+        }
+        else
+            throw Exception(ErrorCodes::SYNTAX_ERROR, "First argument of function {} shall be a time interval (examples, 1ms, 2s, 3m etc), but got '{}' instead", func.name, func.formatForErrorMessage());
+
+        data.addLagBehindKeys(lag_interval);
     }
     else
-        throw Exception(ErrorCodes::SYNTAX_ERROR, "First argument of function {} shall be a time interval (examples, 1ms, 2s, 3m etc), but got '{}' instead", func.name, func.formatForErrorMessage());
+    {
+        /// 2) lag_behind(2, col1, col2): the lag unit is based on alignment keys.
+        if (auto * interval = func.arguments->children.at(0)->as<ASTLiteral>())
+        {
+            lag_interval = interval->value.get<Int64>();
+            if (lag_interval <= 0)
+                throw Exception(
+                    ErrorCodes::BAD_QUERY_PARAMETER,
+                    "First argument of function {} shall be positive integral (examples, 1, 2, 3 etc), but got '{}' instead",
+                    func.name, func.formatForErrorMessage());
+        }
+        else
+            throw Exception(ErrorCodes::SYNTAX_ERROR, "First argument of function {} shall be a integral (examples, `lag_behind(10, t1.ts, t2.ts)` etc), but got '{}' instead", func.name, func.formatForErrorMessage());
 
-    if (data.timestamp_left_key || data.timestamp_right_key)
-        throw Exception(
-            ErrorCodes::INVALID_JOIN_ON_EXPRESSION, "Multiple `lag_behind()` or `aligned_with()` are detected in ON section. Unexpected '{}'", queryToString(ast));
+        auto left_alignment_key = func.arguments->children.at(1);
+        auto right_alignment_key = func.arguments->children.at(2);
+    
+        if (data.left_alignment_key || data.right_alignment_key)
+            throw Exception(
+                ErrorCodes::INVALID_JOIN_ON_EXPRESSION, "Multiple `lag_behind()` are detected in ON section. Unexpected '{}'", queryToString(ast));
 
-    auto table_numbers = getTableNumbers(left, right, data);
-    data.addLagBehindKeys(left, right, table_numbers, lag_interval);
-}
-
-void CollectJoinOnKeysMatcher::handleAlignedWith(const ASTFunction & func, const ASTPtr & ast, Data & data)
-{
-    assert(func.name == "aligned_with");
-
-    if (func.arguments->children.size() != 2)
-        throw Exception(
-            ErrorCodes::SYNTAX_ERROR, "Function {} takes 2 arguments, but got '{}' instead", func.name, func.formatForErrorMessage());
-
-    /// aligned_with(col1, col2)
-    ASTPtr left = func.arguments->children.at(0);
-    ASTPtr right = func.arguments->children.at(1);
-
-    if (data.timestamp_left_key || data.timestamp_right_key)
-        throw Exception(
-            ErrorCodes::INVALID_JOIN_ON_EXPRESSION,
-            "Multiple `lag_behind()` or `aligned_with()` are detected in ON section. Unexpected '{}'",
-            queryToString(ast));
-
-    auto table_numbers = getTableNumbers(left, right, data);
-    data.addAlignedWithKeys(left, right, table_numbers);
+        auto table_numbers = getTableNumbers(left_alignment_key, right_alignment_key, data);
+        data.addLagBehindKeys(left_alignment_key, right_alignment_key, table_numbers, lag_interval);
+    }
 }
 /// proton : ends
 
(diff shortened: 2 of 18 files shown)


src/Core/Settings.h
```diff
@@ -796,7 +796,9 @@ static constexpr UInt64 operator""_GiB(unsigned long long value)
     M(Int64, max_join_range, 300, "Max join range", 0) \
     M(Bool, compact_kv_stream, true, "Control if compact a changelog kv or versioned kv stream during query", 0) \
     M(UInt64, keep_versions, 3, "Control how many versions for each key kept in memory when joining. Used in versioned_kv join", 0) \
-    M(Int64, join_latency_threshold, 0, "Control when to start join left stream with right stream. Zero means automatic system picked threshold", 0) \
+    M(Int64, join_latency_threshold, 0, "Control when to start join left stream with right stream (only works on join alignment). Zero means automatic system picked threshold", 0) \
```
chenziliang: `Control when to start join left stream with right stream for alignment join scenarios. Zero means system picked threshold.`

src/Core/Settings.h
```diff
@@ -796,7 +796,9 @@ static constexpr UInt64 operator""_GiB(unsigned long long value)
...
+    M(Int64, join_quiesce_threshold, 0, "Control how long left stream or right stream is in quiesce (only works on join alignment). Zero means automatic system picked threshold", 0) \
```
chenziliang: Please rebase develop since it is added in other PR already.

src/Core/Settings.h
```diff
@@ -796,7 +796,9 @@ static constexpr UInt64 operator""_GiB(unsigned long long value)
...
+    M(Bool, enable_join_alignment, false, "Enable streaming join with alignment", 0) \
```
chenziliang: Do we need this ? I was thinking if `join_latency_threshold` is not zero, then enable alignment join automatically or alignment join is enabled by default for join scenarios where we think fit.

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.h
```diff
@@ -27,81 +30,120 @@ class JoinTransformWithAlignment final : public IProcessor
...
     {
-        return getWatermark(chunk, right_watermark_column_position, right_watermark_column_type);
-    }
+        explicit InputPortWithData(InputPort * input_port_) : input_port(input_port_) { }
 
-    inline Int64 getLeftWatermark(const Chunk & chunk) const
-    {
-        return getWatermark(chunk, left_watermark_column_position, left_watermark_column_type);
-    }
+        InputPort * input_port;
```
chenziliang, line 52: Let's move the data members to the end of the struct. `ctor / dtor, member functions, data members`

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.h
```diff
@@ -27,81 +30,120 @@ class JoinTransformWithAlignment final : public IProcessor
...
-            return max_ts;
+            ckpt_ctx = chunk.getCheckpointContext();
+
+            /// If the input needs to update data, currently the input is always two consecutive chunks with _tp_delta `-1 and +1`
+            /// So we have to process them together before processing another input
+            /// NOTE: Assume the first retracted chunk of updated data always set RetractedDataFlag.
+            if (chunk.isRetractedData())
+            {
+                input_chunks.emplace_back(std::move(chunk), LightChunkWithTimestamp{});
+                required_update_processing = true;
```
chenziliang: Look like required_update_processing mud the water a LOT. And this function is too complex (too many branches) and will be slooow.

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -27,108 +34,72 @@ JoinTransformWithAlignment::JoinTransformWithAlignment(
...
     }
-    else if (left_input.input_port->isFinished())
+    else if (input_with_data.input_port->isFinished())
     {
         /// Close the other input port
         right_input.input_port->close();
-
-        outputs.front().finish();
-
+        right_input.input_port->close();
```
chenziliang, line 124: left_input.input_port ?

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -27,108 +34,72 @@ JoinTransformWithAlignment::JoinTransformWithAlignment(
...
-        if (left_input.input_port->hasData())
+        if (input_with_data.input_port->hasData())
         {
-            left_input.input_chunks.push_back(left_input.input_port->pull(true));
-
-            if (left_input.input_chunks.back().hasRows())
-            {
-                if (auto new_watermark = getLeftWatermark(left_input.input_chunks.back()); new_watermark > left_input.watermark)
-                    left_input.watermark = new_watermark;
+            auto chunk = input_with_data.input_port->pull(true);
```
chenziliang: line 97 ~ 101 can be merged to `input_with_data.add(...)` ?

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -178,7 +147,9 @@ IProcessor::Status JoinTransformWithAlignment::prepare()
         return Status::NeedData;
 }
 
-/// Watermark alignment algorithm in general:
+/// FIXME: Separate enrichment join and biredictional join implementation
+#if 0
```
chenziliang: We don't need `if 0` ?

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -202,79 +173,138 @@ IProcessor::Status JoinTransformWithAlignment::prepare()
 /// When right stream can garbage collect its data ?
 /// For asof join, for now, we manually use keep_versions to do garbage collection. Ideally we can do better job (automatically), but seems hard since version
 /// and watermark / timestamp are different.
+#endif
+
+/// Watermark alignment algorithm in general:
+/// 1) Pull the right stream first and track its watermark / timestamp as right_stream_watermark.
+///     a. If events' `right_stream_watermark <= `,
```
chenziliang: seems this line is half-way done

src/Core/Settings.h
yl-lisen: yes, it's a temprary settings, in future enable it by default.

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -289,6 +319,158 @@ void JoinTransformWithAlignment::work()
     }
 }
 
+template <bool is_left_input>
+void JoinTransformWithAlignment::processInputData(LightChunk & chunk)
```
chenziliang: I think we shall separate this into 2 functions. Merging them together is super hard to read and there are not a lot of code shared actually.

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.h
```diff
@@ -27,81 +30,120 @@ class JoinTransformWithAlignment final : public IProcessor
...
-    }
+        InputPort * input_port;
 
-    inline Int64 getWatermark(const Chunk & chunk, std::optional<size_t> col_pos, const DataTypePtr & col_type) const
-    {
-        if (col_pos)
+        /// Input state
+        /// NOTE: Assume the input chunk is time-ordered
+        /// pair-<retracted_chunk, chunk>
+        SERDE std::list<std::pair<LightChunk, LightChunkWithTimestamp>> input_chunks;
```
chenziliang: We don't need the pair ? It seems unnecessary introduce complexity and inefficiency. Just `std::list<LightChunkWithTimestamp>` will do the work ? If first chunk is retracted, just append it and then wait for the following append chunk.

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.h
```diff
@@ -27,81 +30,120 @@ class JoinTransformWithAlignment final : public IProcessor
...
+            }
+
+            watermark = std::max(input_chunks.back().second.maxTimestamp(), watermark);
         }
-        else
-            return DB::UTCMilliseconds::now();
-    }
 
-    bool isRightInputInQuiesce() const noexcept { return DB::MonotonicMilliseconds::now() - right_input.last_data_ts >= quiesce_threshold; }
+        bool hasValidInputs() const noexcept { return !input_chunks.empty() && !(input_chunks.size() == 1 && required_update_processing); }
```
chenziliang: `hasCompleteChunks()`

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -202,79 +173,138 @@ IProcessor::Status JoinTransformWithAlignment::prepare()
...
+                unmuteInput(right_input);
+                break;
             }
         }
-        else
-        {
-            /// Left input doesn't have any data yet, we like to pull some. Mute the right input to
-            /// give left input more chance to catch up.
-            left_input.muted = false;
+    } while (left_input.hasValidInputs() && right_input.hasValidInputs());
```
chenziliang: do we need this while ?

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.h
yl-lisen: The LightChunkWithTimestamp don't contain chunk_ctx, if so we need extra flag to mark it

yl-lisen: and for retracted data, its min/max timestamp is also late and bad.

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -202,79 +173,138 @@ IProcessor::Status JoinTransformWithAlignment::prepare()
...
-                stats.right_input_muted += 1;
-            }
-        }
+    if (!left_input.hasValidInputs())
+    {
+        /// Left input has not valid chunks, unmute it
+        unmuteInput(left_input);
+        /// If right input has at least one valid chunk (another one may be an incomplete updated chunk pair <retracted_chunk, null>),
+        /// mute right input to avoid buffer too much data.
+        /// FIXME: allow buffer more ? limited by rows or bytes.
```
chenziliang: If we need alignment, we can't buffer more otherwise the alignment may never happen since we are always progressing the watermark if we buffer more ?

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -202,79 +173,138 @@ IProcessor::Status JoinTransformWithAlignment::prepare()
...
+        if (right_input.input_chunks.size() >= 2)
```
chenziliang: input.chunks.size() == 2 means at least 2 chunks or may be 4 chunks since it is a pair. I don't get the logic here.

src/Interpreters/TreeRewriter.cpp
```diff
@@ -739,10 +739,10 @@ void collectJoinedColumns(TableJoin & analyzed_join, ASTTableJoin & table_join,
...
         }
 
-        /// FIXME, validate if the combination of lag_behind(...) + strictness makes sense.
-        if (analyzed_join.lagBehindInterval() != 0)
-            /// Add lag behind columns
-            data.lagBehindASTToKeys();
+        /// FIXME, validate if the combination of lag_behind(...)/aligned_with(...) + strictness makes sense.
+        if (analyzed_join.requiredJoinAlignment())
+            /// Add alingned with timestamp columns
+            data.timestampASTToKeys();
```
chenziliang: we may rename timestamp to `alignedTimestamp`, `aligned_timestamp_key`, `aligned_timestamp_column_name` etc.

src/Interpreters/TableJoin.h
```diff
@@ -346,12 +347,14 @@ class TableJoin
     std::shared_ptr<const IKeyValueEntity> getStorageKeyValue() { return right_kv_storage; }
 
     /// proton : starts
-    void addLagBehindKeys(const ASTPtr & left_table_ast, const ASTPtr & right_table_ast);
+    void addTimestampKeys(const ASTPtr & left_table_ast, const ASTPtr & right_table_ast);
+    void setRequiredJoinAlignment(bool required_join_alignment_ = true) noexcept { required_join_alignment = required_join_alignment_; }
```
chenziliang: JoinAlignment shall be deduced automatically via lag_behind, aligned_with functions, so it seems this flag can be removed.

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -910,7 +910,7 @@ void HashJoin::init()
 
     /// FIXME, for now, only asof join need align watermark if join latency is not zero
     // For window join, we shall do same but probably in a different way
-    require_aligned_streams = (streaming_strictness == Strictness::Asof) && (left_data.join_stream_desc->latency_threshold != 0);
+    require_aligned_streams = left_data.join_stream_desc->enable_join_alignment;
```
chenziliang: Let's deduce this from join descriptions etc. For global join, it may doesn't quite matter

src/Interpreters/Streaming/JoinStreamDescription.h
```diff
@@ -21,12 +21,16 @@ struct JoinStreamDescription
         Block input_header_,
         DataStreamSemanticEx data_stream_semantic_,
         UInt64 keep_versions_,
-        Int64 latency_threshold_)
+        bool enable_join_alignment_ = false,
+        Int64 latency_threshold_ = 0,
+        Int64 quiesce_threshold_ = 0)
```
chenziliang: Please rebase develop

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
yl-lisen: `if (hasCompleteChunks()) muteInput(...)` may is better

src/Interpreters/Streaming/JoinStreamDescription.h
```diff
@@ -64,11 +68,20 @@ struct JoinStreamDescription
...
     ///    ON left.k = right.k AND left.version < right.version AND
-    ///       lag_behind(left.ts, right.ts, 20)
+    ///       lag_behind(20ms, left.ts, right.ts)
+    /// SETTINGS keep_versions=16;
+    ///
+    /// OR
+    ///
+    /// SELECT * FROM left ASOF JOIN right
+    ///    ON left.k = right.k AND left.version < right.version AND
+    ///       aligned_with(left.ts, right.ts)
```
chenziliang: why wen need another function `aligned_with(...)` ? Can we merge the 2 into one ? We will need express the `latency` parameter anyway. The current `aligned_with(left.ts, right.ts)` needs fallback to query settings to refer the join latency.

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
yl-lisen: if (hasCompleteChunks()) muteInput(...) may is better

src/Interpreters/Streaming/JoinStreamDescription.h
chenziliang: There are 2 options: create a general align function like `align(<what-to-align or align-policy-name>, <latency>, [col, ...])`. Or we can create different align functions for different purposes / scenarios.

If the former, something we will have is like 

```
align('watermark', 20ms, left.ts, right.ts);
align('interleave', 1);  <== interleaved consuming left / right inputs. Not necessary per row, but per chunk.
align(...);
```

BTW, it appears `align(...)` - the function name looks weird in a AND clause

chenziliang: or we can introduce another clause.

```
SELECT * FROM left ASOF JOIN right ON left.k = right.k AND left.version < right.version ALIGN ...
```

chenziliang: Function seems easier implementation wise.

src/Interpreters/Streaming/JoinStreamDescription.h
```diff
@@ -64,11 +68,20 @@ struct JoinStreamDescription
...
     /// SETTINGS keep_versions=16;
     UInt64 keep_versions;
+    bool enable_join_alignment;
     Int64 latency_threshold;
-    String lag_column;
+    Int64 quiesce_threshold;
+    String timestamp_column;
```
chenziliang: alignment_column

chenziliang: alignment can be timestamp or an integral column ?

chenziliang: For `interleave` alignment, we don't need an alignment column, how we are going to represent that ?

chenziliang: `lag_behind(1)` can do the work, we will need introduce a lag unit / type. It may be timestamp milliseconds, or it may be a block like in `lag_behind(1)`

yl-lisen: agree it, I will remove `aligned_with`, it's not neccessary

BTW, lag_behind(1, key1, key2), means lag unit/type is same with key1 and key2.
lag_behind(1ms) means lag 1ms based on processing time 
lag_behind(1) means by default 1 chunk ?. (Probably not needed)

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.h
chenziliang: LightChunkWithTimestamp is a super struct of LightChunk. How comes it wont work ? Need update processing bool already points out the current chunk is a retract ?

src/Interpreters/Streaming/JoinStreamDescription.h
chenziliang: yes, i was thinking similar things regarding these scenarios. 

**lag_behind(1, key1, key2)**, means lag unit/type is same with key1 and key2. 

I think we still need allow specifying the unit like 1ms ? Otherwise if key1 is nanoseconds precision, if users like to milliseconds alignment (most of the time, it will be the case), then user need specify something like **lag_behind(50000000, ts1, ts2)** which appears not quite convenient  and intuitive. As mentioned in your FIXME, we indeed need validate the key1, key2 supports the lag precision like a `date` doesn't support milliseconds precision lagging. Throw if we find this mismatch.

When key1, key2 are not timestamp or if they are timestamp but users just specify a raw number without unit, we will need handle these cases as well. For example,

lag_behind(20, integral_col1, integral_col2);
lag_behind(50000000, nanosecond_datetime64_col1, nanosecond_datetime64_col2);

If key1, key2 doesn't have same type, throws as well for now.

**lag_behind(1ms)** means lag 1ms based on processing time
yes, it is the same as just setting `join_latency_threshold`.

**lag_behind(1)** means by 1 chunk.
we still need this for interleave reading / consuming ?

chenziliang: Personally i prefer for timestamp lag behind, the first argument is always interval like 10ms instead of a raw number like 10. This will make our processing logic easier and more intuitive as well.

src/Interpreters/CollectJoinOnKeysVisitor.cpp
```diff
@@ -668,51 +676,68 @@ void CollectJoinOnKeysMatcher::handleLagBehind(const ASTFunction & func, const A
...
+                    ErrorCodes::BAD_QUERY_PARAMETER,
+                    "First argument of function {} shall be positive millisecond, second or minute time interval (examples, 1ms, 2s, 3m etc), but got '{}' instead",
+                    func.name, func.formatForErrorMessage());
+        }
         else
-            throw Exception(
-                ErrorCodes::NOT_IMPLEMENTED,
-                "First argument of function {} only supports millsecond, second or minute time interval (examples, 1ms, 1s, 2m etc), but got '{}' instead",
-                func.name, func.formatForErrorMessage());
+            throw Exception(ErrorCodes::SYNTAX_ERROR, "First argument of function {} shall be a time interval (examples, 1ms, 2s, 3m etc), but got '{}' instead", func.name, func.formatForErrorMessage());
```
chenziliang: We still need support `lag_behind(1)` to support interleave consuming ?

src/Interpreters/CollectJoinOnKeysVisitor.cpp
```diff
@@ -668,51 +676,68 @@ void CollectJoinOnKeysMatcher::handleLagBehind(const ASTFunction & func, const A
...
+        }
+        else
+            throw Exception(ErrorCodes::SYNTAX_ERROR, "First argument of function {} shall be a integral (examples, `lag_behind(10, t1.ts, t2.ts)` etc), but got '{}' instead", func.name, func.formatForErrorMessage());
 
-    if (data.lag_behind_left_key || data.lag_behind_right_key)
-        throw Exception(
-            ErrorCodes::INVALID_JOIN_ON_EXPRESSION, "Multiple `lag_behind()` are detected in ON section. Unexpected '{}'", queryToString(ast));
+        auto left_alignment_key = func.arguments->children.at(1);
+        auto right_alignment_key = func.arguments->children.at(2);
+
```
chenziliang: Let's remove this empty line

src/Interpreters/CollectJoinOnKeysVisitor.cpp
```diff
@@ -92,36 +92,44 @@ void CollectJoinOnKeysMatcher::Data::addLagBehindKeys(
...
     }
+
+    /// left stream lag behind right stream
+    /// lag_behind(100ms, left_stream.ts, right_stream.ts)
+    analyzed_join.setLagBehindInterval(lag_interval);
+}
+
+void CollectJoinOnKeysMatcher::Data::addLagBehindKeys(Int64 lag_interval)
+{
+    left_alignment_key = nullptr;
```
chenziliang: We shall throw if left_alignemnt_key or right is set.

src/Interpreters/CollectJoinOnKeysVisitor.cpp
```diff
@@ -92,36 +92,44 @@ void CollectJoinOnKeysMatcher::Data::addLagBehindKeys(
...
             ErrorCodes::INVALID_JOIN_ON_EXPRESSION,
-            "Expressions {} and {} are from the same stream but from different arguments of equal function in lag_behind() function",
+            "Expressions {} and {} are from the same stream but from different arguments of equal function in `lag_behind()` function",
             queryToString(left_ast),
             queryToString(right_ast));
     }
+
+    /// left stream lag behind right stream
+    /// lag_behind(100ms, left_stream.ts, right_stream.ts)
+    analyzed_join.setLagBehindInterval(lag_interval);
```
chenziliang: What about right stream lag behind left ?

src/Interpreters/CollectJoinOnKeysVisitor.cpp
```diff
@@ -668,51 +676,68 @@ void CollectJoinOnKeysMatcher::handleLagBehind(const ASTFunction & func, const A
...
-            throw Exception(
-                ErrorCodes::BAD_QUERY_PARAMETER,
-                "First argument of function {} shall be positive millisecond, second or minute time interval (examples, 1ms, 2s, 3m etc), but got '{}' instead",
-                func.name, func.formatForErrorMessage());
+        data.addLagBehindKeys(lag_interval);
     }
     else
-        throw Exception(ErrorCodes::SYNTAX_ERROR, "First argument of function {} shall be a time interval (examples, 1ms, 2s, 3m etc), but got '{}' instead", func.name, func.formatForErrorMessage());
+    {
+        /// 2) lag_behind(2, col1, col2): the lag unit is based on alignment keys.
```
chenziliang: We still need `lag_behind(2ms, ts1, ts2)` as mentioned in other place. And for timestamp, the first argument being always an interval seems very intuitive.

src/Interpreters/CollectJoinOnKeysVisitor.cpp
```diff
@@ -668,51 +676,68 @@ void CollectJoinOnKeysMatcher::handleLagBehind(const ASTFunction & func, const A
...
+        else
+            throw Exception(ErrorCodes::SYNTAX_ERROR, "First argument of function {} shall be a integral (examples, `lag_behind(10, t1.ts, t2.ts)` etc), but got '{}' instead", func.name, func.formatForErrorMessage());
 
-    if (data.lag_behind_left_key || data.lag_behind_right_key)
-        throw Exception(
-            ErrorCodes::INVALID_JOIN_ON_EXPRESSION, "Multiple `lag_behind()` are detected in ON section. Unexpected '{}'", queryToString(ast));
+        auto left_alignment_key = func.arguments->children.at(1);
+        auto right_alignment_key = func.arguments->children.at(2);
+    
+        if (data.left_alignment_key || data.right_alignment_key)
```
chenziliang: We can move the check to `addLagBehindKeys` consistently

src/Interpreters/CollectJoinOnKeysVisitor.h
```diff
@@ -104,6 +105,7 @@ class CollectJoinOnKeysMatcher
     static std::pair<Int64, bool> handleLeftAndRightArgumentsForRangeBetweenAsOfJoin(const ASTLiteral * literal_arg, ASTPtr non_literal_arg, Data & data);
     /// `lag_behind` is special case as `date_diff_within`
     static void handleLagBehind(const ASTFunction & func, const ASTPtr & ast, Data & data);
+    static void handleAlignedWith(const ASTFunction & func, const ASTPtr & ast, Data & data);
```
chenziliang: do we still need this ?

src/Interpreters/Streaming/JoinStreamDescription.h
```diff
@@ -21,8 +21,8 @@ struct JoinStreamDescription
         Block input_header_,
         DataStreamSemanticEx data_stream_semantic_,
         UInt64 keep_versions_,
-        Int64 latency_threshold_,
-        Int64 quiesce_threshold_ms_)
+        Int64 latency_threshold_ = 0,
```
chenziliang: Let's remove this default parameters to expose compile issue everywhere. It is a good thing to do this check.

src/Interpreters/CollectJoinOnKeysVisitor.h
```diff
@@ -58,7 +58,8 @@ class CollectJoinOnKeysMatcher
 
         /// proton : starts
         void addLagBehindKeys(const ASTPtr & left_ast, const ASTPtr & right_ast, JoinIdentifierPosPair table_pos, Int64 lag_interval);
-        void lagBehindASTToKeys();
+        void addLagBehindKeys(Int64 lag_interval);
```
chenziliang: `addLagBehindValue()` may be better.

src/Interpreters/TableJoin.h
```diff
@@ -148,8 +148,9 @@ class TableJoin
     /// If lag_interval < 0 : right stream lag behind left stream.
     Int64 lag_interval = 0;
     size_t data_block_size = 0;
-    String left_table_lag_column;
-    String right_table_lag_column;
+    String left_alignment_key_column;
+    String right_alignment_key_column;
+    bool required_join_alignment = false;
```
chenziliang: do we still need this ?

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.h
```diff
@@ -27,81 +30,84 @@ class JoinTransformWithAlignment final : public IProcessor
...
-        bool muted = false;
-        bool required_checkpoint = false;
-    };
+    void muteRightInput() noexcept
+    {
+        right_input.muted = true;
+        ++stats.right_input_muted;
+    }
+
+    void unmuteInput(InputPortWithData & input_with_data) noexcept { input_with_data.muted = false; }
```
chenziliang: `static void `

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -180,121 +150,368 @@ IProcessor::Status JoinTransformWithAlignment::prepare()
...
-                join->insertRightBlock(right_input.input_port->getHeader().cloneWithColumns(right_input.input_chunk.detachColumns()));
-            }
-            else
-            {
-                /// FIXME, watermark / heartbeat / checkpoint chunk.
-                right_input.input_chunk.clear();
-            }
+            if (retracted_chunk.rows())
+                processRightInputData(retracted_chunk);
+
```
chenziliang, line 226: remove this empty line

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -180,121 +150,368 @@ IProcessor::Status JoinTransformWithAlignment::prepare()
...
-            else
-            {
-                /// Left stream is leading ahead, mute left input to wait right input to progress
-                left_input.muted = true;
-                right_input.muted = false;
-                stats.left_input_muted += 1;
-            }
+            if (retracted_chunk.rows())
+                processLeftInputData(retracted_chunk);
+
```
chenziliang, line 258: remove empty line

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -24,113 +31,78 @@ JoinTransformWithAlignment::JoinTransformWithAlignment(
...
-    left_watermark_column_type = left_join_stream_desc->lagBehindColumnType();
+    left_input.is_changelog_input = isChangelogDataStream(left_join_stream_desc->data_stream_semantic);
+    right_input.is_changelog_input = isChangelogDataStream(right_join_stream_desc->data_stream_semantic);
 
-    right_watermark_column_position = right_join_stream_desc->lagBehindColumnPosition();
-    right_watermark_column_type = right_join_stream_desc->lagBehindColumnType();
+    /// Some join combinations have builtin buffer to do aligning, so don't build an extra external buffer to align:
+    /// 1) For asof join, the right side has builtin buffer to keep multiple versions
+    /// 2) For range join, the left and right side both have builtin buffer to keep multiple ranges
+    left_input.need_aligned_buffer = !join->leftHasBuiltInAlignedBuffer();
```
chenziliang: Here we like to say, if we like to buffer data in transform or just go ahead to push the data to hash join object.

We will need revise a few things
1. The interface names. 
2. The variable names

left_input.need_aligned_buffer => left_input.need_buffer_data_to_align
leftHasBuiltInAlignedBuffer => leftStreamRequiresBufferingDataToAlign

After we refactor out different join mode, these interface can go away probably.

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -24,113 +31,78 @@ JoinTransformWithAlignment::JoinTransformWithAlignment(
...
         quiesce_threshold_ms,
         latency_threshold,
-        left_join_stream_desc->lag_column,
-        right_join_stream_desc->lag_column);
+        left_join_stream_desc->alignment_column,
+        right_join_stream_desc->alignment_column);
 }
 
-IProcessor::Status JoinTransformWithAlignment::prepareRightInput()
+IProcessor::Status JoinTransformWithAlignment::prepareInput(InputPortWithData & input_with_data)
```
chenziliang, line 108: After we refactor out different modes, the prepareLeftInput / prepareRightInput is asymmetrix for asof join and data enrichment streaming join etc

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -180,121 +150,368 @@ IProcessor::Status JoinTransformWithAlignment::prepare()
...
 
-        /// 1) We pulled right input and it has data
-        if (right_input.input_chunk)
+        auto & [retracted_chunk, chunk] = right_input.input_chunks.front();
+
+        /// We can directly process current input data in followsing scenarios:
+        /// 1) Don't need align inputs (right side of join has builtin buffer to align)
+        /// 2) Right input chunk's max timestamp less than left stream's min timestamp
+        /// 3) Or left input is in quiesce, we still need to push execute the join probably
+        if (!right_input.need_aligned_buffer || chunk.maxTimestamp() <= left_input.minTimestamp() + latency_threshold
```
chenziliang: I think we shall compare watermark here, left watermark, right watermark and the latency threshold ?

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
```diff
@@ -180,121 +150,368 @@ IProcessor::Status JoinTransformWithAlignment::prepare()
...
+    {
+        if (isCancelled())
+            return;
+
+        /// We can directly process current input data in followsing scenarios:
+        /// 1) Don't need align inputs (left side of join has builtin buffer to align)
+        /// 2) Left input chunk's max timestamp less than right stream's min timestamp
+        /// 3) Or right input is in quiesce, we still need to push execute the join probably
+        auto & [retracted_chunk, chunk] = left_input.input_chunks.front();
+        if (!left_input.need_aligned_buffer || chunk.maxTimestamp() + latency_threshold < right_input.minTimestamp()
```
chenziliang: Same thing. I feel like the compare logic has some problem.

src/Processors/Transforms/Streaming/JoinTransformWithAlignment.cpp
yl-lisen: Compare watermark is great, it's one alignment mode, as you said, I will do more alignment modes in next ticket

PR #382, opened 2023-12-05
Bugfix/issue 381 update the ip trie dictionary key type

### LINKED ISSUE #381: Update the ip_trie dictionary key type .
**Describe what's wrong**
The previous clickhouse ip_trie dictionary is string, but it had been changed since the ipv4 and ipv4 became independent type.While I ported the `ipv4 data type` https://github.com/timeplus-io/proton/pull/264, I missed the change here.

lizhou1111:
this close #381 
The clickhouse has 7 extended dictionary types:
- flat
- hased
- range_hased
- cache
- complex_key_hased
- complex_key_cache
- ip_trie

The key type of previous ip_trie type dictionary is fixed string, So the third argument passed to `dict_get('dict_name', attr_names, id_expr)` should be fixed_string.After this update the argument type has been changed to ipv4(only ipv4 can work,clickhouse does the same).
For example:
```sql
drop stream if exists geoip_t;
CREATE stream if not exists geoip_t (
  cidr string,
  latitude float64,
  longitude float64,
  country_code string,
  state string,
  city string
) ENGINE = MergeTree ORDER BY cidr;
insert into geoip_t values ('188.166.84.125',52.3759,4.8975,'NL','North Holland','Amsterdam')

select * from geoip_t;

drop dictionary geoip;
-- ingest data from table, you can ingest data from csv(etc. source (FILE(path './test.csv' format 'CSVWithNames'))).
CREATE DICTIONARY geoip (
  cidr string,
  latitude float64,
  longitude float64,
  country_code string,
  state string,
  city string
) PRIMARY KEY cidr
SOURCE (ClickHouse (table 'geoip_t'))
LIFETIME (MIN 300 MAX 360)
LAYOUT (IP_TRIE());


select dict_get('geoip', 'latitude', to_ipv4('188.166.84.125')) as result;
┌──result─┐
│ 52.3759 │
└─────────┘
```


tests/stream/test_stream_smoke/0002_ingest_and_ddl_case.json
```diff
@@ -137,6 +137,109 @@
...
+                {
+                    "statements": [
+                        {
+                            "client": "python",
+                            "query_type": "table",
+                            "query": "drop stream if exists geoip_t"
+                        },
+                        {
+                            "client": "python",
+                            "query_type": "table",
```
yl-lisen: I prefer adding these in stateless tests

yl-lisen (approved):

Issue #175, opened 2023-10-17
Kafka external stream support csv etc input formats


chenziliang:
**Use case**

Support input formats here : https://clickhouse.com/docs/en/interfaces/formats

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

zliang-min:
Let's firstly clarify what does it mean to support CSV for kafka external stream. I suppose how this works is that each message read from the kafka topic is a line of CSV, and the number of values matches the number of fields of the stream. For example, given a kafka external stream:
```sql
CREATE EXTERNAL STREAM employees (
  id string,
  name string,
  job_title string,
  joined_on datetime64,
  level integer
) SETTINGS type='kafka'
```

It expects that each kafka message looks like this:
```csv
1001,Some Body,Software Engineer,2023-10-01T09:00:00Z,5
```

Also, we will need to convert the values to the types specified in the stream schema. That means, in the above case, it needs to convert the `joined_on` value to type `datetime64`, and `level` to `integer`.

Does this make sense?

chenziliang:
Yes, a bit more detail https://clickhouse.com/docs/en/engines/table-engines/integrations/kafka and we will need figure out how https://clickhouse.com/docs/en/interfaces/formats works.

But in general, each format expects a specific form of input and do parsing, conversion etc on the fly to match the defined schema.

chenziliang:
And this integration shall be very nature and the formats code just does the work and we will be careful about perf and need documents the input details with examples.

chenziliang:
Protobuf has highest priority.

zliang-min:
Reading from kafka external streams with CSV format actually already works today, and in fact, all supported input data format should work too ( but I didn't test all of them, I just tested CSV and JSONEachRow so far ). However, there is a major caveat ( or bug ): the external stream cannot user `_tp_time` ( probably, all T+ internal fields can't be used, I tried `_tp_sn` which will fail as well ). For example, given stream:
```sql
CREATE EXTERNAL STREAM foo (
  a string,
  b int64
) SETTINGS type='kafka',data_format='CSV',...
```
and the target Kafka topic contains messages like `some string, 10`, `select * from foo` works as expected. However, if we create a stream like
```sql
CREATE EXTERNAL STREAM foo (
  _tp_time datetime64,
  a string,
  b int64
) SETTINGS type='kafka',data_format='CSV',...
```

[...]

PR #399, opened 2023-12-11
feat: manage format schemas with SQL.

zliang-min:
PR checklist:
- Did you run ClangFormat ?

refs #175 

Introduce new SQL commands for managing format schemas (for now, only `Protobuf` schemas are supported), and make read/write Protobuf messages with Kafka external streams work.

New SQL commands:
* create a format schema
    ```sql
    CREATE FORMAT SCHEMA <name> AS <body> TYPE <type>
    ```
* show all format schemas
    ```sql
    SHOW FORMAT SCHEMAS [TYPE <type>]
    ```
* show one format schema
    ```sql
    SHOW CREATE FORMAT SCHEMA <name> [TYPE <type>]
    ```
* drop one format schema
    ```sql
    DROP FORMAT SCHEMA <name> [TYPE <type>]
    ```

`CREATE EXTERNAL STREAM` can now use the `format_schema` setting. The `format_schema` setting can also be set in `SELECT` and `INSERT` queries, and they have higher priority than the one specified with `CREATE EXTERNAL STREAM`.


chenziliang:
Does `FORMAT SCHEMA` read like dup meaning ? Does `DATA SCHEMA` reads better ?

zliang-min:
> Does `FORMAT SCHEMA` read like dup meaning ? Does `DATA SCHEMA` reads better ?

"format schema" already exists in the code base and used in all places where schemas are used (e.g. `format_schema` is an existing format setting). Also, since a schema is of a specific format, like Protobuf, using "format schema" makes sense in this context.

So, I think we should not create another term unless there is a strong reason.

src/Formats/FormatSchemaFactory.h
```diff
@@ -0,0 +1,49 @@
...
+{
+/// Manages format schemas.
+class FormatSchemaFactory final : private boost::noncopyable
+{
+public:
+    /// options for what to do if a schema alread exists
+    enum class ExistsOP {
+        noops, /// do nothing
+        replace, /// replace with the new content
+        Throw, /// throw an exception
```
yl-lisen: I prefer use `Ignore, Overwrite, Throw`

src/Formats/FormatSchemaFactory.cpp
```diff
@@ -0,0 +1,220 @@
+#include <Formats/FormatFactory.h>
+#include <Formats/FormatSchemaFactory.h>
+#include <Formats/FormatSchemaInfo.h>
+#include <google/protobuf/compiler/parser.h>
+#include <google/protobuf/descriptor.pb.h>
+#include <google/protobuf/io/tokenizer.h>
+#include <google/protobuf/io/zero_copy_stream_impl.h>
```
yl-lisen: we should have different format schema files, for example:
```c++
class IFormatSchema;

/// ProtobufFormatSchema.h
class ProtobufFormatSchema final : IFormatSchema;
```

yl-lisen: This will make it easier to extend a new format schema

src/Formats/FormatSchemaFactory.cpp
```diff
@@ -0,0 +1,220 @@
...
+{
+private:
+    std::vector<String> errors;
+
+public:
+    void AddError(int line, google::protobuf::io::ColumnNumber column, const std::string & message) override
+    {
+        std::ostringstream error;
+        error << "line: " << line << ", column: " << column << ", error: " << message;
+        errors.push_back(error.str());
```
yl-lisen: A better way
```c++
errors.push_back(fmt::format("line: {}, columns: {}, error: {}", line, column, message));
```

src/Formats/FormatSchemaFactory.cpp
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    String toString() const
+    {
+        std::ostringstream ret;
+        for (const auto & error : errors)
+            ret << error << std::endl;
+        return ret.str();
```
yl-lisen: `return fmt::join(errors, "\n")`

src/Formats/FormatSchemaFactory.cpp
```diff
@@ -0,0 +1,220 @@
...
+    std::istringstream s_payload(payload);
+    google::protobuf::io::IstreamInputStream input(&s_payload);
+    ErrorCollector error_collector;
+    google::protobuf::io::Tokenizer tokenizer(&input, &error_collector);
+    google::protobuf::FileDescriptorProto descriptor;
+    google::protobuf::compiler::Parser parser;
+
+    parser.RecordErrorsTo(&error_collector);
+    if (!parser.Parse(&tokenizer, &descriptor))
+        throw Exception(ErrorCodes::INVALID_DATA, "Invalid schema, errors: \n{}", error_collector.toString());
```
yl-lisen: we should avoid use `\n` in error message.

src/Formats/FormatSchemaFactory.cpp
```diff
@@ -0,0 +1,220 @@
...
+}
+}
+
+void FormatSchemaFactory::registerSchema(const ContextPtr & context, const String & schema_name, const String & schema_type, const String & schema_body, ExistsOP if_exists)
```
yl-lisen: I prefer put the Context as last parameter

yl-lisen: same suggests for other interfaces

src/Formats/FormatSchemaFactory.cpp
```diff
@@ -0,0 +1,220 @@
...
+            return;
+        if (if_exists == ExistsOP::Throw)
+            throw Exception(ErrorCodes::FORMAT_SCHEMA_ALREADY_EXISTS, "Format schema {} of type {} already exists", schema_name, schema_type);
+    }
+
+    if (schema_type == "Protobuf")
+        validateProtobufSchema(schema_body);
+
+    std::ofstream schema_file(fsinfo.absoluteSchemaPath(), std::ios::binary);
+    schema_file << schema_body;
```
yl-lisen: it seems we register the format schema into disk, and get it from disk, why not cache snapshot in memory ?
Lock to access the filesystem, which is not a good usage, since filesystem ops still can be execute by user at same time.
Usually, we do these things:
1) load filesystem info in memory map during proton bootstrap
2) register: write in filesystem, write in memory map
3) unregister: remove from filesytem, remove from memory map
4) get: get from memory map

src/Formats/FormatSchemaFactory.cpp
```diff
@@ -0,0 +1,220 @@
...
+    assert(!schema_name.empty());
+    assert(!schema_type.empty());
+    assert(!schema_body.empty());
+
+    checkSchemaType(schema_type);
+
+    auto format_settings = getFormatSettings(context);
+    FormatSchemaInfo fsinfo{schema_name, schema_type, false, true, format_settings.schema.format_schema_path};
+
+    std::lock_guard lock(mutex);
```
yl-lisen, line 53: We should make the lock granularity as small as possible

src/Interpreters/InterpreterCreateFormatSchemaQuery.cpp
```diff
@@ -0,0 +1,41 @@
...
+    access_rights_elements.emplace_back(AccessType::CREATE_FORMAT_SCHEMA);
+
+    auto replace_if_exists = create_format_schema_query.or_replace;
+    if (replace_if_exists)
+        access_rights_elements.emplace_back(AccessType::DROP_FORMAT_SCHEMA);
+
+    auto current_context = getContext();
+    current_context->checkAccess(access_rights_elements);
+
+    auto if_exists = FormatSchemaFactory::ExistsOP::Throw;
```
yl-lisen: rename it to `exists_op`

src/Parsers/ParserQueryWithOutput.cpp
```diff
@@ -59,6 +61,10 @@ bool ParserQueryWithOutput::parseImpl(Pos & pos, ASTPtr & node, Expected & expec
     ParserShowGrantsQuery show_grants_p;
     ParserShowPrivilegesQuery show_privileges_p;
     ParserExplainQuery explain_p(end, allow_settings_after_format_in_insert);
+    /// proton: starts
+    ParserShowFormatSchemasQuery show_format_schemas_p;
+    ParserShowCreateFormatSchemaQuery format_schema_p;
```
yl-lisen: show_create_format_schema_p

src/Storages/ExternalStream/ExternalStreamSettings.h
```diff
@@ -27,8 +27,7 @@ class ASTStorage;
     /* those are mapped to format factory settings */ \
     M(String, data_format, "", "The message format, for example JSONEachRow", 0) \
     M(String, row_delimiter, "\n", "The string to be considered as a delimiter in raw message.", 0) \
-    M(UInt64, max_row_length, 4096, "Max row length", 0) \
-    M(String, data_schema, "", "Schema identifier (used by schema-based formats)", 0)
```
yl-lisen, line 31: `data_schema` and `data_format` are very similar, i think we should use a clearer name

src/Storages/ExternalStream/StorageExternalStreamImpl.h
```diff
@@ -1,20 +1,27 @@
...
+    explicit StorageExternalStreamImpl(std::unique_ptr<ExternalStreamSettings> settings_): settings(std::move(settings_)) {}
     virtual ~StorageExternalStreamImpl() = default;
 
     virtual void startup() = 0;
     virtual void shutdown() = 0;
     virtual bool supportsSubcolumns() const { return false; }
     virtual NamesAndTypesList getVirtuals() const { return {}; }
+    /// Some implementations have its own logic to infer the format.
+    virtual const String & dataFormat() const { return settings->data_format.value; }
+    const String & formatSchema() const { return settings->format_schema.value; }
```
yl-lisen, line 25: Revisiting settings - data_format and foramt_schema, they are confusing

yl-lisen:
> PR checklist:
> 
> * Did you run ClangFormat ?
> * Did you separate headers to a different section in existing community code base ?
> * Did you surround `proton: starts/ends` for new code in existing community code base ?
> 
> Please write user-readable short description of the changes:
> 
> refs #175
> 
> Introduce new SQL commands for managing format schemas (for now, only `Protobuf` schemas are supported), and make read/write Protobuf messages with Kafka external streams work.
> 
> New SQL commands:
> 
> * create a format schema
>   ```sql
>   CREATE FORMAT SCHEMA <name> AS <body> TYPE <type>
>   ```
> * show all format schemas
>   ```sql
>   SHOW FORMAT SCHEMAS [TYPE <type>]
>   ```
> * show one format schema
>   ```sql
>   SHOW CREATE FORMAT SCHEMA <name> [TYPE <type>]
>   ```
> * drop one format schema
>   ```sql
>   DROP FORMAT SCHEMA <name> [TYPE <type>]
>   ```
> 
> `CREATE EXTERNAL STREAM` can now use the `format_schema` setting. The `format_schema` setting can also be set in `SELECT` and `INSERT` queries, and they have higher priority than the one specified with `CREATE EXTERNAL STREAM`.

Let's add some stateless test for new SQL commands

src/Formats/FormatSchemaFactory.cpp
zliang-min: Yes, agreed. I just wanted to push this later when we need to add the 2nd schema format support. The reason is when we only have one implementation, we might not have enough information to design what the interface should look like, thus we would end-up with a design that needs refactor again later when we have a new schema to support.

Let me know if you still think we should have the interface in this PR.

src/Formats/FormatSchemaFactory.cpp
zliang-min: `error_collector.toString()` already contains `"\n"`. If we don't use `"\n"` here, the issue is that, the error message could be very long and very difficult to read ( when the user uses `proton-client` ). It could be something like this `Invalid schema, errors: line: x, column: y, error: this is an error message; line: x1, column y1, error: this is another error message; line: x2, column y2, error: this is the 3rd error message` and this could go on and on. As you can see, this could be quite challenging to read.

Let me know what you think.

src/Formats/FormatSchemaFactory.cpp
zliang-min: Sure, but I found there are existing code doing it this same way, like
```cpp
template <typename T>
FormatSettings getFormatSettings(ContextPtr context, const T & settings);
```

And some functions even have `ContextPtr` in the middle, we probably should advocate this rule better.

src/Formats/FormatSchemaFactory.cpp
zliang-min: I thought about this too. There are a couple problems with such implementation:
* the filesystem and in-memory cache get out-of-sync.
* the current Protobuf input/output formats implementations rely on filesystem, if we want to use them w/o filesystem, that will require quite some refactor effort.

So, the plan is that, we will need to come back and refactor this for cluster support anyways. At that point, we will eventually move out from filesystem for good.

src/Formats/FormatSchemaFactory.h
zliang-min: I replaced `noop` with `Ignore`, and kept `Replace` to make it consistent with `CREATE OR REPLACE`.

src/Storages/ExternalStream/ExternalStreamSettings.h
zliang-min: The reason why I removed `data_schema` is that `format_schema`/`FormatSchema` already exists in the code base and `format_schema` is already part of the FORMAT_FACTORY_SETTINGS. And `data_format` is something we added later.

All schemas belong to a certain format, like it could a Protobuf scheams, a JSON schema, or an Avro schema, etc. So, calling it FormatSchema makes sense. Thus, using `data_format` and `format_schema` together is totally fine IMO. `data_format` tells which format it uses, while `format_schema` tells which schema of that format it uses.

src/Storages/ExternalStream/StorageExternalStreamImpl.h
zliang-min: Please check the other comment I made above.

src/Formats/FormatSchemaFactory.cpp
zliang-min: I think this is the smallest scope, suggestions?

src/Formats/FormatSchemaFactory.cpp
zliang-min: I refactored the code and now introduced `IExternalSchemaWriter`, which is a counter-part of the existing `IExternalSchemaReader`. This should make things consistent with the existing code and easier to extend.

src/Formats/FormatSchemaFactory.cpp
zliang-min: Removed `"\n"` from error message.

yokofly:
CI  failure with unrelated change(it's my new test..) sorry, I will post a new commit to fix.

src/Formats/FormatFactory.cpp
```diff
@@ -523,6 +552,36 @@ void FormatFactory::registerExternalSchemaReader(const String & name, ExternalSc
...
+    auto & target = dict[name].external_schema_writer_creator;
+    if (target)
+        throw Exception("FormatFactory: Schema writer " + name + " is already registered", ErrorCodes::LOGICAL_ERROR);
+    target = std::move(external_schema_writer_creator);
+}
+
+bool FormatFactory::checkIfFormatHasExternalSchemaWriter(const String & name)
+{
+    const auto & target = getCreators(name);
+    return bool(target.external_schema_writer_creator);
```
yl-lisen: `static_cast<bool>(target.external_schema_writer_creator);`

src/Formats/FormatFactory.cpp
```diff
@@ -523,6 +552,36 @@ void FormatFactory::registerExternalSchemaReader(const String & name, ExternalSc
     target = std::move(external_schema_reader_creator);
 }
 
+/// proton: starts
+void FormatFactory::registerExternalSchemaWriter(const String & name, ExternalSchemaWriterCreator external_schema_writer_creator)
+{
+    auto & target = dict[name].external_schema_writer_creator;
+    if (target)
+        throw Exception("FormatFactory: Schema writer " + name + " is already registered", ErrorCodes::LOGICAL_ERROR);
```
yl-lisen: Please use this this interface format:
```c++
throw Exception(ErrorCodes::LOGICAL_ERROR, "FormatFactory: Schema writer {} is already registered", name);
```

src/Formats/FormatFactory.cpp
```diff
@@ -523,6 +552,36 @@ void FormatFactory::registerExternalSchemaReader(const String & name, ExternalSc
...
+
+ExternalSchemaWriterPtr FormatFactory::getExternalSchemaWriter(
+    const String & name,
+    std::string_view body,
+    ContextPtr & context,
+    const std::optional<FormatSettings> & _format_settings) const
+{
+    const auto & external_schema_writer_creator = dict.at(name).external_schema_writer_creator;
+    if (!external_schema_writer_creator)
+        throw Exception("FormatFactory: Format " + name + " doesn't support schema creation.", ErrorCodes::LOGICAL_ERROR);
```
yl-lisen: same as above

src/Formats/ProtobufSchemas.cpp
```diff
@@ -13,8 +14,33 @@ namespace ErrorCodes
...
+private:
+    SchemaValidationErrors errors;
+
+public:
+    void AddError(int line, google::protobuf::io::ColumnNumber column, const std::string & message) override
+    {
+        errors.emplace_back(line, column, message);
+    }
+
+    SchemaValidationErrors getErrors() const
```
yl-lisen: A better way is that return `const SchemaValidationErrors &`

src/Formats/ProtobufSchemas.cpp
```diff
@@ -79,6 +105,21 @@ const google::protobuf::Descriptor * ProtobufSchemas::getMessageTypeForFormatSch
     return importer->import(info.schemaPath(), info.messageName());
 }
 
+/// proton: starts
+SchemaValidationErrors ProtobufSchemas::validateSchema(std::string_view & schema)
```
yl-lisen: For `string_view`, we don't need use a reference

src/Formats/ProtobufSchemas.h
```diff
@@ -37,6 +41,9 @@ class ProtobufSchemas : private boost::noncopyable
     /// The function never returns nullptr, it throws an exception if it cannot load or parse the file.
     const google::protobuf::Descriptor * getMessageTypeForFormatSchema(const FormatSchemaInfo & info);
 
+    /// proton: starts
+    SchemaValidationErrors validateSchema(std::string_view & schema);
```
yl-lisen: For `string_view`, not need a reference

src/Interpreters/InterpreterCreateFormatSchemaQuery.cpp
```diff
@@ -0,0 +1,57 @@
...
+BlockIO InterpreterCreateFormatSchemaQuery::execute()
+{
+    auto & create_format_schema_query = query_ptr->as<ASTCreateFormatSchemaQuery &>();
+
+    if (!create_format_schema_query.cluster.empty())
+        throw Exception(ErrorCodes::NOT_IMPLEMENTED, "ON CLUSTER is not implemented for CREATE FORMAT SCHEMA");
+
+    auto body = create_format_schema_query.getSchemaBody();
+    auto trimmed_body = trim_whitespaces_at_both_ends(body);
+    if (trimmed_body.empty())
```
yl-lisen: we can rename `trim_whitespaces_at_both_ends` to `trimToView`

src/Processors/Formats/ISchemaWriter.h
```diff
@@ -0,0 +1,46 @@
...
+    /// Validates the schema input.
+    virtual SchemaValidationErrors validate() = 0;
+
+    /// Persistents the schema.
+    /// If the schema already exists, and replace_if_exist is false, it returns false.
+    /// Otherwise it returns true. Throws exceptions if it fails to write.
+    virtual bool write(bool replace_if_exist) = 0;
+
+protected:
+    std::string_view schema_body;
```
yl-lisen, line 42: save a string view, it's a risk, we don't know the original string is valid or not
if we can assume it always exists, use `const String &` is a better way

src/Formats/FormatSchemaFactory.cpp
```diff
@@ -0,0 +1,181 @@
...
+    auto writer = FormatFactory::instance().getExternalSchemaWriter(format, schema_body, context, format_settings);
+    assert(writer); /* confirmed with checkSchemaType */
+    if (auto errors = writer->validate(); !errors.empty())
+        throw Exception(ErrorCodes::INVALID_DATA, "Invalid Protobuf schema, errors: {}", formatSchemaValidationErrors(errors));
+
+    auto result = writer->write(exists_op == ExistsOP::Replace);
+    if (!result && exists_op == ExistsOP::Throw)
+            throw Exception(ErrorCodes::FORMAT_SCHEMA_ALREADY_EXISTS, "Format schema {} of type {} already exists", schema_name, format);
```
yl-lisen: too many indents ?

src/Formats/FormatFactory.cpp
zliang-min: This was to keep the same style used in other function in `FormatFactory`. But we don't have to.

src/Formats/FormatFactory.cpp
zliang-min: Same as above.

src/Formats/FormatFactory.cpp
zliang-min: Same as above.

src/Interpreters/InterpreterCreateFormatSchemaQuery.cpp
zliang-min: I renamed it to just `trim`, because I think `ToView` is redundant, the function signature already says it returns a view.

src/Processors/Formats/ISchemaWriter.h
zliang-min: The Writer is deisnged to not to own the schema body, so yes, it assumes it out-lives the body string ( I belive this contract is clear because the constructor function takes a view as a parameter ).

I can switch it to a `const String &` if that's neccessary. But could you elaborate a bit on why `const String &` is better in this case, please?

yl-lisen (approved):
LGTM

src/Processors/Formats/ISchemaWriter.h
yl-lisen: Because `const String &` as a external references more intuitively,
string_view also is good.

jovezhong:

jovezhong:

PR #417, opened 2023-12-15
make function name  case insensitive

### LINKED ISSUE #383: Function name sensitivity

So far, Proton has mixed case sensitivity for its function names. What we can do is 
1) Making them all case sensitive 
2) Making them all case insensitive. 

\1 Has breaking changes.
\2 Seems making sense as there are integrations which depends on upper case COUNT aggr function for example. But there are caveats regarding these changes since sometimes we compare function names in the code to do special things. We shall have converted them to lower cases at these places before the comparison most of the time, but there may be still somewhere missing this lower case conversion, hence will have breaking changes if we are not at care.

lizhou1111:
this closes #383 
This pr contains two parts:
1. make all function case insensitve
```sql
create stream test1(id int, value int);
select id, max(value), MAX(value), min(value), Count(1) from test1 group by id;
insert into test1(id, value) values (2,2)
insert into test1(id, value) values (1,2)
insert into test1(id, value) values (1,3)
insert into test1(id, value) values (2,3)
┌─id─┬─max(value)─┬─MAX(value)─┬─min(value)─┬─Count()─┐
│  2 │          2 │          2 │          2 │       1 │
└────┴────────────┴────────────┴────────────┴─────────┘
┌─id─┬─max(value)─┬─MAX(value)─┬─min(value)─┬─Count()─┐
│  2 │          2 │          2 │          2 │       1 │
│  1 │          2 │          2 │          2 │       1 │
└────┴────────────┴────────────┴────────────┴─────────┘
┌─id─┬─max(value)─┬─MAX(value)─┬─min(value)─┬─Count()─┐
│  2 │          2 │          2 │          2 │       1 │
│  1 │          3 │          3 │          2 │       2 │
└────┴────────────┴────────────┴────────────┴─────────┘
┌─id─┬─max(value)─┬─MAX(value)─┬─min(value)─┬─Count()─┐
│  2 │          3 │          3 │          2 │       2 │
│  1 │          3 │          3 │          2 │       2 │
└────┴────────────┴────────────┴────────────┴─────────┘

```
2. make combination function also case insensitive, such as `count_if`
3. Fix a small bug about quantile.The `quantile(key, level)` can take one or two parameters, but the previous version only can take two parameters.The default value of level is 0.5, it will work as the `median` function.
```sql
select quantile(value) from test1;
insert into test1(id, value) values (1,1);
insert into test1(id, value) values (1,2);
insert into test1(id, value) values (1,3)

┌─quantile(value)─┐
│               1 │
└─────────────────┘
┌─quantile(value)─┐
│             1.5 │
└─────────────────┘
┌─quantile(value)─┐
│               2 │
└─────────────────┘

```


src/Compression/CompressionFactory.cpp
```diff
@@ -66,7 +66,7 @@ CompressionCodecPtr CompressionCodecFactory::get(
             }
             else if (const auto * ast_func = inner_codec_ast->as<ASTFunction>())
             {
-                codec_family_name = ast_func->name;
+                codec_family_name = ast_func->covered_name;
```
yl-lisen: toUpper(ast_func->name);

src/Interpreters/ExpressionAnalyzer.cpp
```diff
@@ -171,13 +171,15 @@ void tryTranslateToParametricAggregateFunction(
...
     else if (lower_name == "quantile")
     {
-        /// Translate `quantile(key, level)` to `quantile(level)(key)`
-        if (arguments.size() != 2)
-            throw Exception(ErrorCodes::NUMBER_OF_ARGUMENTS_DOESNT_MATCH, "Aggregate function {} requires two arguments.", node->name);
-
-        ASTPtr expression_list = std::make_shared<ASTExpressionList>();
-        expression_list->children.push_back(arguments[1]);
-        parameters = getAggregateFunctionParametersArray(expression_list, "", context);
+        ///Translate `quantile(key, level)` to `quantile(level)(key)`,and the default level is 0.5, median fucntion is the alias of quantile(key, 0.5)
```
yl-lisen, line 174: add whitespace behind `///`

src/Interpreters/ExpressionAnalyzer.cpp
```diff
@@ -171,13 +171,15 @@ void tryTranslateToParametricAggregateFunction(
...
+        if (arguments.size() != 2 && arguments.size() != 1)
+            throw Exception(ErrorCodes::NUMBER_OF_ARGUMENTS_DOESNT_MATCH, "Aggregate function {} requires one or two arguments", node->name);
+        if (arguments.size() == 2)
```
yl-lisen, line 177: add a empty line between these

src/Interpreters/ExpressionAnalyzer.cpp
yl-lisen: move the comments to line 179

src/Parsers/ASTFunction.h
```diff
@@ -25,6 +25,7 @@ class ASTFunction : public ASTWithAlias
     String code_name;
 
     /// If covered_name exists, we shall show 'covered_name(...)' instead of 'name(...)' (Only covered function name)
+    /// Example: In the stream query the 'now64' function will be changed to '__streaming_now64' function.So the covered_name is 'now64'
```
yl-lisen: execute query: select now64() from stream;
internal query: select __streaming_now64() from stream;
show query: select now64() from stream;

src/Parsers/ASTFunction.cpp
```diff
@@ -256,7 +257,6 @@ void ASTFunction::formatImplWithoutAlias(const FormatSettings & settings, Format
         settings.ostr << code_name;
         return;
     }
-    /// proton: ends.
```
yl-lisen: shouldn't remove this line

src/Storages/StorageFactory.cpp
```diff
@@ -232,7 +253,15 @@ AccessType StorageFactory::getSourceAccessType(const String & table_engine) cons
...
-        return AccessType::NONE;
+    {
+        /// proton: starts
+        it = case_insensitive_storages.find(Poco::toLower(table_engine));
+        /// proton: ends
+
+        if (it == case_insensitive_storages.end())
+            return AccessType::NONE;
+    }
+
```
yl-lisen: add private functions `get` and `tryGet` to simplify the logic here

src/Storages/StorageFactory.cpp
```diff
@@ -126,12 +135,17 @@ StoragePtr StorageFactory::get(
...
             {
-                auto hints = getHints(name);
                 /// proton: starts
-                if (!hints.empty())
-                    throw Exception("Unknown engine " + name + ". Maybe you meant: " + toString(hints), ErrorCodes::UNKNOWN_STORAGE);
-                else
-                    throw Exception("Unknown engine " + name, ErrorCodes::UNKNOWN_STORAGE);
+                name = Poco::toLower(name);
+                it = case_insensitive_storages.find(name);
+                if (it == case_insensitive_storages.end())
```
yl-lisen: same here

src/Storages/StorageFactory.cpp
```diff
@@ -205,7 +219,14 @@ StoragePtr StorageFactory::get(
 
     assert(arguments.getContext() == arguments.getContext()->getGlobalContext());
 
-    auto res = storages.at(name).creator_fn(arguments);
+    auto it = storages.find(name);
+
+    /// proton: starts
+    if (it == storages.end())
+        it = case_insensitive_storages.find(Poco::toLower(name));
+    /// proton: ends
```
yl-lisen: same here

src/Parsers/ExpressionElementParsers.cpp
```diff
@@ -852,6 +852,11 @@ bool ParserFunction::parseImpl(Pos & pos, ASTPtr & node, Expected & expected, [[
     auto function_node = std::make_shared<ASTFunction>();
     tryGetIdentifierNameInto(identifier, function_node->name);
 
+    /// proton: starts.
+    function_node->covered_name = function_node->name;
+    function_node->name = Poco::toLower(function_node->name);
```
yl-lisen: Only set covered_name if lower name is different with original name

yl-lisen (approved):

Issue #425, opened 2023-12-18
multishard incorrect result on aggr_distinct (with append only stream)


yokofly:
**Describe what's wrong**
```
client@1:
create stream v2(id int, val int) primary key id settings shards=3;
client@1:
select sum_distinct(val), sum(val) from v2
client@2:
insert into v2(id, val) values(3, 30), (4, 40)
client@2:
insert into v2(id, val) values(1, 30), (7, 40)
```

**Error message and/or stacktrace**

```
c1 :) select sum_distinct(val), sum(val) from v2

SELECT
  sum_distinct(val), sum(val)
FROM
  v2

Query id: 695de04b-70bf-4d6e-9497-319d339d8c43

┌─sum_distinct_streaming(val)─┬─sum(val)─┐
│                          70 │       70 │
└─────────────────────────────┴──────────┘
┌─sum_distinct_streaming(val)─┬─sum(val)─┐
│                         100 │      140 │
```
**Additional context**

the result shall be 70 right?

Issue #419, opened 2023-12-18
when do checkpoint on aggr_distinct, the distinct need trigger the nested func dockpt as well.


yokofly:
**How to reproduce**
proton@1.3.27 and below case exists in the append-only stream

**Error message and/or stack trace**
```
client@1: create stream v(id int, val float)
client@1: subscribe to select sum_distinct(val) from v SETTINGS checkpoint_interval=1

client@2: insert into v(id, val) values(2, 3)

client@1: kill query 
client@1: recover from 'query_id'

client@2: insert into v(id, val) values(1, 2)
```
below is the output in client@1
```
daa12914b7a5 :) subscribe to select sum_distinct(val) from v SETTINGS checkpoint_interval=1

SELECT
  sum_distinct(val)
FROM
  v
SETTINGS
  checkpoint_interval = 1

Query id: 634ced7d-4d08-4287-aa62-0d3945adb857

┌─sum_distinct(val)─┐
│                 3 │
└───────────────────┘
↘ Progress: 1.00 rows, 4.00 B (0.26 rows/s., 1.04 B/s.)               (0.0 CPU)^Cancelling query.
Query was cancelled.

1 row in set. Elapsed: 4.040 sec. 

daa12914b7a5 :) recover from '634ced7d-4d08-4287-aa62-0d3945adb857'

RECOVER FROM '634ced7d-4d08-4287-aa62-0d3945adb857'

Query id: 29c7c4ff-8ddc-43c3-add5-b43110eea158

┌─sum_distinct(val)─┐
│                 2 │
└───────────────────┘
↙ Progress: 1.00 rows, 4.00 B (0.10 rows/s., 0.38 B/s.)               (0.0 CPU)^Cancelling query.
Query was cancelled.

1 row in set. Elapsed: 10.711 sec. 

daa12914b7a5 :) 

```
**Additional context**

PR #426, opened 2023-12-19
Feature/issue 3303 distinct aggr support changelog

yokofly:
- make distinct this aggr combinator support changelog/retract.
- the basic design uses a hashmap(internal is using btreemap) to save the state(count value).  and pass the processed data to the next function.
- also enhance changelog function parse.
- add some helper function in CountedValueMap(internal btreemap)

test:
count_distinct_if shall be found in https://github.com/timeplus-io/proton/compare/feature/issue-3303-distinct-aggr-support-changelog?expand=1#diff-f3bf2509e030a3c87801df81d8e1b41cc2062a0bd3930eca45a8026cbb88320aR214-R228
serialize ckpt also works, see https://github.com/timeplus-io/proton/compare/feature/issue-3303-distinct-aggr-support-changelog?expand=1#diff-f3bf2509e030a3c87801df81d8e1b41cc2062a0bd3930eca45a8026cbb88320aR36-R43

and we have reported 2 issues, one is still open #425 , one(#419) will be fixed in commit 2fea686cba3de5bb6af5caaf3b9b3a040a3e0a81


src/Interpreters/Streaming/SubstituteStreamingFunction.cpp
```diff
@@ -169,6 +173,13 @@ void StreamingFunctionData::visit(DB::ASTFunction & func, DB::ASTPtr)
                 return;
             }
         }
+        else
+        {
+            /// replace `<aggr>_distinct[_combinator]` ==> `<aggr>_distinct_streaming[_combinator]` for streaming query
+            constexpr std::string_view distinct_raw{"_distinct"}, distinct_streaming_raw{"_distinct_streaming"};
+            if (size_t pos = func.name.find(distinct_raw); pos != std::string::npos)
+                func.name.replace(pos, distinct_raw.length(), distinct_streaming_raw);
+        }
```
yl-lisen: /// Always show original func name
func.covered_name = func.name;

yl-lisen: we should define a function:
```c++
void ASTFunction::substitute(String new_name)
{
    /// Always show original func name
    if (func.covered_name.empty())
        func.covered_name = func.name;

    func.name = std::move(new_name);
}
```

tests/stream/test_stream_smoke/0013_changelog_stream13.yaml
```diff
@@ -0,0 +1,263 @@
...
+      - client: python
+        query_type: table
+        wait: 1
+        query: create stream v_12181627(id int, val int) primary key id settings mode='changelog_kv';
+      - client: python
+        query_type: stream
+        depends_on_stream: v_12181627
+        query_id: '12181627213'
+        query: subscribe to select sum_distinct(val), sum(val), count_distinct(val), count(val)
+          from v_12181627 settings checkpoint_interval=1;
```
yl-lisen: Another scenario is global aggregation emit changelog, for example:
```sql
subscribe to select count_distinct(val) from v emit changelog setting checkpoint_interval = 1
```

src/Interpreters/Streaming/SubstituteStreamingFunction.cpp
```diff
@@ -77,21 +76,26 @@ std::optional<String> StreamingFunctionData::supportChangelog(const String & fun
     /// Support combinator suffix, for example:
     /// `count`                 => `__count_retract`
     /// `count_if`              => `__count_retract_if`
-    /// `count_distinct`        => `__count_retract_distinct`
-    /// `count_distinct_if`     => `__count_retract_distinct_if`
+    /// `count_distinct`        => `__count_retract_distinct_retract`
+    /// `count_distinct_if`     => `__count_retract_distinct_retract_if`
     String combinator_suffix;
+    constexpr std::string_view distinct_raw{"_distinct"}, distinct_retract_raw{"_distinct_retract"};
```
yl-lisen: directly use literal string "_distinct" is better ?

src/Interpreters/Streaming/SubstituteStreamingFunction.cpp
```diff
@@ -159,7 +163,7 @@ void StreamingFunctionData::visit(DB::ASTFunction & func, DB::ASTPtr)
                     /// Make _tp_delta as the last argument to avoid unused column elimination for query like below
                     /// SELECT count(), avg(i) FROM (SELECT i, _tp_delta FROM versioned_kv) GROUP BY i; =>
                     /// SELECT __count_retract(_tp_delta), __avg_retract(i, _tp_delta) FROM (SELECT i, _tp_delta FROM versioned_kv) GROUP BY i; =>
-                    if (func.name.starts_with("__count_retract") && delta_pos - func.arguments->children.begin() > 0)
+                    if ((func.name == "__count_retract" || func.name == "__count_retract_if") && delta_pos - func.arguments->children.begin() > 0)
                         /// Fix for nullable since this substitution is not equal
                         func.arguments->children[0] = std::make_shared<ASTIdentifier>(ProtonConsts::RESERVED_DELTA_FLAG);
                     else
```
yl-lisen, line 167: we can remove this special path

src/Interpreters/Streaming/SubstituteStreamingFunction.cpp
yl-lisen: and please use this func instead of anywhere needs substitute func ,
some one is in `StreamingNowFunctionData::visit`

src/AggregateFunctions/Streaming/CountedValueMap.h
```diff
@@ -38,6 +38,7 @@ struct CountedValueArena<StringRef>
             arena.free(const_cast<char *>(key.data), key.size);
     }
 
+    ArenaWithFreeLists * getArenaWithFreeLists() { return &arena; }
 private:
```
yl-lisen: separate with an empty line

src/AggregateFunctions/Streaming/CountedValueMap.h
```diff
@@ -98,6 +99,35 @@ class CountedValueMap
...
+        }
+        else
+        {
+            /// Didn't find v in the map
+            [[maybe_unused]] auto [_, inserted] = m.emplace(arena->emplace(std::move(v)), 1);
+            assert(inserted);
+
+            eraseExtraElements();
+            return true;
+        }
```
yl-lisen: we shouldn't copy another same code, we can refactor this interface:
```c++
/// \returns: the emplaced element iterator, if failed to emplace, return invalid iterator, `m.end()`
iter tryEmplace(T v)
{
    ...
    return <iter> or <empty_iter>;
}

/// \returns: added.
bool insert(T v)
{
    auto iter = tryEmplace(std::move(v));
    return iter != m.end();
}
```

yl-lisen: For your distinct implementation, you can use
```
auto iter = map.tryEmplace(key);
/// If added and ref count is 1, assume it's a new key
if (iter != map.end() && iter->second == 1)
{
    ....
}
```

yl-lisen: same for erase

src/AggregateFunctions/Streaming/AggregateFunctionDistinctRetract.h
```diff
@@ -0,0 +1,341 @@
...
+        }
+        else
+        {
+            for (const auto & [key, _] : map)
+            {
+                const char * begin = key.data;
+                for (size_t i = 0; i < argument_size - 1; ++i)
+                    begin = argument_columns[i]->deserializeAndInsertFromArena(begin);
+
+                argument_columns[argument_size - 1]->insert(static_cast<int8_t>(+1));
```
yl-lisen: argument_columns.back()->insert(Int8(1));

src/AggregateFunctions/Streaming/AggregateFunctionDistinctRetract.h
```diff
@@ -0,0 +1,341 @@
...
+        {
+            for (const auto & [key, delta_col] : extra_data_since_last_finalize)
+            {
+                /// serialize key
+                const char * begin = key.c_str();
+                for (size_t i = 0; i < argument_size - 1; ++i)
+                    begin = argument_columns[i]->deserializeAndInsertFromArena(begin);
+
+                /// insert delta_col
+                argument_columns[argument_size - 1]->insert(delta_col);
```
yl-lisen: argument_columns.back()->insert(delta_col);

src/Interpreters/Streaming/SubstituteStreamingFunction.cpp
```diff
@@ -177,8 +176,13 @@ void StreamingFunctionData::visit(DB::ASTFunction & func, DB::ASTPtr)
         {
             /// replace `<aggr>_distinct[_combinator]` ==> `<aggr>_distinct_streaming[_combinator]` for streaming query
             constexpr std::string_view distinct_raw{"_distinct"}, distinct_streaming_raw{"_distinct_streaming"};
-            if (size_t pos = func.name.find(distinct_raw); pos != std::string::npos)
-                func.name.replace(pos, distinct_raw.length(), distinct_streaming_raw);
+            if (size_t pos = func.name.find(distinct_raw); pos != std::string::npos) {
+                std::string new_name = func.name;
+                new_name.replace(pos, distinct_raw.length(), distinct_streaming_raw);
+
```
yokofly, line 190: this is a bit ugly, but I do not wanna introduce boost..
```cpp
if (string.contains("_distinct")) // cpp 23 , proton use cpp@20
if (str.find("_distinct") != std::string::npos) // current proton style
{
    auto new_name(func.name);
    boost::replace_first(new_name, "_distinct", "_distinct_streaming")
    func.substitute(new_name)
 }
 ```

File: src/AggregateFunctions/Streaming/CountedValueMap.h [hunk 10]
```diff
@@ -72,48 +73,28 @@ class CountedValueMap
...
     }
 
-    bool insert(T v)
+    /**
+    - If returnIfNewData is true, 
+        the function behaves returning true only if the element is new insert instead.
+    - If returnIfNewData is false (which is the default), 
+        the function behaves like the original insert,
+        returning true in all cases except when the capacity is full, and the new element is not desired.
+    */
```
yl-lisen: For a good interface, it should have a clear parameters and returns.
but now, it's complex and confusing

src/Interpreters/Streaming/SubstituteStreamingFunction.cpp
yokofly: wait, proton has using cpp 23, but the `replace` still manual in cpp standard
edit: ummm but proton uses cpp20

src/AggregateFunctions/Streaming/CountedValueMap.h
yokofly: is below works?
```cpp

    /**
    * Inserts an element into the map.
    * 
    * @param v The element to be inserted.
    * @param returnTrueIfNewData Controls the return behavior:
    *    - If true, returns true only if a new element is actually inserted.
    *    - If false (default), behaves like the standard insert: 
    *      returns true except when the capacity is full and the new element is not inserted.
    * 
    * @return True if the element is successfully processed (inserted or found), false otherwise.
    */
    bool insert(T v, bool returnTrueIfNewData = false)
    {
        if (atCapacity())
        {
            /// At capacity, this is an optimization
            /// fast ignore elements we don't want to maintain
            if (less(lastValue(), v))
                return false;
        }

        auto iter = m.find(v);
        if (iter != m.end())
        {
            ++iter->second;
            return !returnTrueIfNewData;
        }
        else
        {
            /// Didn't find v in the map
            [[maybe_unused]] auto [_, inserted] = m.emplace(arena->emplace(std::move(v)), 1);
            assert(inserted);

            eraseExtraElements();
            return true;
        }
    }

    /**
    * Erases an element from the map, if present.
    * 
    * @tparam TT The type of the element to be erased.
    * @param v The element to be erased.
    * @param returnTrueIfNewErased Controls the return behavior:
    *    - If true, returns true only if the element is found and actually erased (i.e., its count reaches zero).
    *    - If false (default), returns true if the element is found, regardless of whether it's actually erased.
    * 
    * @return True if the element is successfully processed based on the 'returnTrueIfNewErased' flag, false otherwise.
    */
    template <typename TT>
    bool erase(const TT & v, bool returnTrueIfNewErased = false) /// To enable heterogeneous erase
    {
        auto iter = m.find(v);
        if (iter != m.end())
        {
            --iter->second;
            if (iter->second == 0)
            {
                arena->free(iter->first);
                m.erase(iter);
                return true;
            }
            return !returnTrueIfNewErased;
        }
        return false;
    }

```

src/Interpreters/Streaming/SubstituteStreamingFunction.cpp
```diff
@@ -153,19 +158,25 @@ void StreamingFunctionData::visit(DB::ASTFunction & func, DB::ASTPtr)
...
 
                 return;
             }
         }
+        else
+        {
+            /// replace `<aggr>_distinct[_combinator]` ==> `<aggr>_distinct_streaming[_combinator]` for streaming query
+            if (!boost::algorithm::contains(func.name, "_distinct_streaming")) {
+                if (boost::algorithm::contains(func.name, "_distinct")) {
+                    std::string new_name = func.name;
```
yl-lisen:
```c++
if (boost::algorithm::contains(func_name_lower, "_distinct"))
{
    boost::algorithm::replace_first(func_name_lower, "_distinct", "_distinct_streaming")
    substitueFunction(func, std::move(func_name_lower));
}
```

yokofly: `distinct_if` ?

yokofly: user input: distinct_streaming_if 
we shall skip the change?
so i have to check not contain streaming ?

src/Interpreters/Streaming/SubstituteStreamingFunction.cpp
```diff
@@ -77,21 +78,25 @@ std::optional<String> StreamingFunctionData::supportChangelog(const String & fun
...
     auto nested_func_name = function_name;
     while (iter == changelog_func_map.end())
     {
         if (auto combinator = AggregateFunctionCombinatorFactory::instance().tryFindSuffix(nested_func_name))
         {
-            const std::string combinator_name = combinator->getName();
+            std::string combinator_name = combinator->getName();
             /// TODO: support more combinators
-            if (combinator_name != "_if")
+            if (combinator_name != "_if" && combinator_name != "_distinct" && combinator_name != "_distinct_retract")
```
yl-lisen, line 91: not need care `_distinct_retract` ， can't happen here

yokofly: what if user input `_distinct_retract` ? do not consider this condition?

yl-lisen: we should block them

src/Interpreters/Streaming/SubstituteStreamingFunction.cpp
yokofly: https://godbolt.org/z/xfh1PhKPx  no need to consider user input our internal function naming?
like `distinct_streaming_if`

src/Interpreters/Streaming/SubstituteStreamingFunction.cpp
yl-lisen: yes, let's keep your changes

yl-lisen (approved):

PR #434, opened 2023-12-21
feat: support message key for Kafka external stream

### LINKED ISSUE #433: Support message keys for Kafka external stream
**Use case**

As a kafka external streams user, I would like to be able to set the message keys on messages which are sent to a Kafka external stream.

Add a `message_key` external stream setting to specify an expression for generating message keys. E.g.

```sql
CREATE EXTERNAL STREAM example (
  field_one string,
  field_two int32,
  ...
) SETTINGS type='kafka',...,message_key='field_one';
```

`message_key` can be any expression that returns a string value (string or fixedstring):

```sql
CREATE EXTERNAL STREAM example (
  field_one string,
  field_two int32,
  ...
) SETTINGS type='kafka',...,message_key='split_by_string(\',\', field_one)[1]';
```

There is already a setting `sharding_expr` which allows to specify a function expression to calculate the partition ID of messages, however, message keys could be more useful than just for partitioning, also, they are also part of the messages' metadata stored in Kafka brokers.

**Additional context**

zliang-min:
PR checklist:
- Did you run ClangFormat ?

closes #433 

Added a new external stream setting `message_key` which is an expression that returns a string value, the values return by the expression will be used as the message key for each row. Examples:

```sql
-- use a column
CREATE EXTERNAL STREAM example_one (
  one string,
  two int32
) SETTINGS type='kafka',...,message_key='one'

-- use a complex expression
CREATE EXTERNAL STREAM example_two (
  one string,
  two int32
) SETTINGS type='kafka',...,message_key='split_by_string(\',\', one)[1]'
```

`message_key` can be used with `sharding_expr` together, and `sharding_expr` will take priority (exactly what a Kafka client does).

Together with this change, a couple more changes are worth mentioning:
* the random sharding logic has been simplified, it now relies on librdkafka for partitioning.
* it now uses `rd_kafka_produce_batch` to send messages to Kafka, which is expected to be more efficient ( at least, it removes the overhead of calling `rd_kafka_produce` for each message ). This should benefit `one_message_per_row` even more.


src/Storages/ExternalStream/StorageExternalStream.cpp
```diff
@@ -56,13 +56,13 @@ void validateEngineArgs(ContextPtr context, ASTs & engine_args, const ColumnsDes
...
 std::unique_ptr<StorageExternalStreamImpl> createExternalStream(
-    IStorage * storage, std::unique_ptr<ExternalStreamSettings> settings, ContextPtr & context [[maybe_unused]], const ASTs & engine_args, bool attach, ExternalStreamCounterPtr external_stream_counter)
+    IStorage * storage, std::unique_ptr<ExternalStreamSettings> settings, ContextPtr & context [[maybe_unused]], const ASTs & engine_args, bool attach, ExternalStreamCounterPtr external_stream_counter, ContextPtr context_)
 {
     if (settings->type.value.empty())
         throw Exception(ErrorCodes::BAD_ARGUMENTS, "External stream type is required in settings");
 
     if (settings->type.value == StreamTypes::KAFKA || settings->type.value == StreamTypes::REDPANDA)
-        return std::make_unique<Kafka>(storage, std::move(settings), engine_args, attach, external_stream_counter);
+        return std::make_unique<Kafka>(storage, std::move(settings), engine_args, attach, external_stream_counter, context_);
```
chenziliang: We can move over context_

src/Storages/ExternalStream/StorageExternalStream.cpp
```diff
@@ -149,7 +149,7 @@ StorageExternalStream::StorageExternalStream(
     storage_metadata.setColumns(columns_);
     setInMemoryMetadata(storage_metadata);
 
-    auto stream = createExternalStream(this, std::move(external_stream_settings_), context_, engine_args, attach, std::make_shared<ExternalStreamCounter>());
+    auto stream = createExternalStream(this, std::move(external_stream_settings_), context_, engine_args, attach, std::make_shared<ExternalStreamCounter>(), context_);
```
chenziliang: Same, we can move over context_ ?

src/Storages/ExternalStream/Kafka/WriteBufferFromKafkaSink.h
```diff
@@ -0,0 +1,31 @@
...
+    {
+        if (!offset())
+            return;
+
+        on_next(working_buffer.begin(), offset());
+    }
+
+    const std::function<void(char *, size_t)> on_next;
+};
+}
```
chenziliang, line 34: We may need an extra empty line after

src/Storages/ExternalStream/Kafka/WriteBufferFromKafkaSink.h
```diff
@@ -0,0 +1,31 @@
...
+#include <IO/WriteBuffer.h>
+#include <base/StringRef.h>
+
+namespace DB
+{
+
+class WriteBufferFromKafkaSink final : public BufferWithOwnMemory<WriteBuffer>
+{
+public:
+    explicit WriteBufferFromKafkaSink(const std::function<void(char * pos, size_t len)> callback, size_t buffer_size = DBMS_DEFAULT_BUFFER_SIZE)
```
chenziliang: const object in parameter is almost always a bad idea since we have to copy everything to the target and we probably have to copy everything to the parameter on the call site as well. We can do

```
explicit WriteBufferFromKafkaSink(std::function<...> on_next_, ...)
    : ..
    , on_next(std::move(on_next_))
{
    assert(on_next);
}
```

src/Storages/ExternalStream/Kafka/WriteBufferFromKafkaSink.h
```diff
@@ -0,0 +1,31 @@
...
+private:
+    void nextImpl() override
+    {
+        if (!offset())
+            return;
+
+        on_next(working_buffer.begin(), offset());
+    }
+
+    const std::function<void(char *, size_t)> on_next;
```
chenziliang: Don't have to be const, but it is a personal style.

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -29,8 +27,7 @@ class ChunkSharder
     BlocksWithShard shard(Block block, Int32 shard_cnt) const;
 
 private:
-    void useRandomSharding();
-    Int32 getNextShardIndex(Int32 shard_cnt) const noexcept { return static_cast<Int32>(rand()) % shard_cnt; }
+    Int32 getNextShardIndex(Int32 /*shard_cnt*/) const noexcept { /* do not specify the partition ID */ return RD_KAFKA_PARTITION_UA; }
```
chenziliang: Let's use 2 lines here one for comment and the other for return 

```
/// return RD_KAFKA_PARTITION_UA to let librdkafka to choose a target partition.
return ...
```

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -39,14 +36,13 @@ class ChunkSharder
...
     bool random_sharding = false;
-    mutable std::minstd_rand rand;
 };
 }
 
 class KafkaSink final : public SinkToStorage
 {
 public:
-    KafkaSink(const Kafka * kafka, const Block & header, ContextPtr context, Int32 initial_partition_cnt, Poco::Logger * log_);
+    KafkaSink(const Kafka * kafka, const Block & header, Int32 initial_partition_cnt, const ASTPtr & message_key, ContextPtr context, Poco::Logger * log_);
```
chenziliang: ContextPtr and logger as last parameters, upvote

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -56,42 +52,62 @@ class KafkaSink final : public SinkToStorage
...
+    /// Callback for Kafka message delivery report
+    static void onMessageDelivery(rd_kafka_t * /* producer */, const rd_kafka_message_t * msg, void *  /*opaque*/);
+    void onMessageDelivery(const rd_kafka_message_t * msg);
+
+    void addMessageToBatch(char * pos, size_t len);
+
+    /// the number of acknowledgement has been received so far for the current checkpoint period
+    size_t acked() const { return state.acked; }
+    /// the number of errors has been received so far for the current checkpoint period
+    size_t error_count() const { return state.error_count; }
```
chenziliang: `errorCount()` and i usually put `noexcept` there for most setter / getter if they are no-memory allocation and we are sure it won't throw. like `errorCount() const noexcept`

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -56,42 +52,62 @@ class KafkaSink final : public SinkToStorage
...
+    /// the number of outstanding messages for the current checkpoint period
+    size_t outstandings() const { return state.outstandings; }
+    /// the last error code received from delivery report callback
+    rd_kafka_resp_err_t lastSeenError() const { return static_cast<rd_kafka_resp_err_t>(state.last_error_code.load()); }
+    /// check if there are no more outstandings (i.e. delivery reports have been recieved
+    /// for all out-go messages, regardless if a message is successfully delivered or not)
+    bool hasNoOutstandings() const { return state.outstandings == state.acked + state.error_count; }
```
chenziliang: Can we flip the logic here to make it more readable, like 

```
bool hasOutstandingMessages(...)
{
    return state.outstandings != state.acked + state.error_count;
}
```

Nit: renaming `outstanding()` to `outstandingMessages()` is probably more readable.

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -56,42 +52,62 @@ class KafkaSink final : public SinkToStorage
...
 
-    klog::KafkaPtr producer;
-    klog::KTopicPtr topic;
-    std::unique_ptr<WriteBufferFromKafka> wb;
-    OutputFormatPtr writer;
-    ThreadPool polling_threads;
-    std::atomic_flag is_finished;
     Int32 partition_cnt;
+    bool one_message_per_row;
+    Poco::Logger * log;
```
chenziliang: can we put `Poco::Logger * log;` as the last data member and use `logger` ? It may really not matter, just a style and `log` is usually referred as nlog and klog in the other places.

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -56,42 +52,62 @@ class KafkaSink final : public SinkToStorage
...
 
-    Poco::Logger * log;
+    ExpressionActionsPtr message_key_expr;
+    bool delete_message_key_column;
+    size_t message_key_column_pos;
+
+    /// For constructing the message batch
+    std::vector<rd_kafka_message_t> current_batch;
+    std::deque<StringRef> keys_queue;
+    Int32 next_partition;
```
chenziliang: we may want inline init the primitive data members like int, bool, size_t etc here and there are quite a few of them.

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -56,42 +52,62 @@ class KafkaSink final : public SinkToStorage
...
+    mutable std::atomic_uint_fast32_t next_partition_counter = 0;
+
+    struct State final
```
chenziliang: `final` can be removed since it is private and embeded in a class and it has no virtual functions.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -20,37 +21,39 @@ extern const int CANNOT_WRITE_TO_KAFKA;
...
 extern const int INVALID_CONFIG_PARAMETER;
 extern const int TYPE_MISMATCH;
+extern const int INVALID_SETTING_VALUE;
 }
 
-namespace KafkaStream
+namespace
 {
-void ChunkSharder::useRandomSharding()
+ExpressionActionsPtr buildExpression(ContextPtr context, const Block & header, const ASTPtr & expr_ast)
```
chenziliang: nit : ContextPtr last parameter if you don't mind

chenziliang: `const ContextPtr & ` ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -20,37 +21,39 @@ extern const int CANNOT_WRITE_TO_KAFKA;
...
+
+    ASTPtr query = expr_ast;
+    auto syntax_result = TreeRewriter(context).analyze(query, header.getNamesAndTypesList());
+    return ExpressionAnalyzer(query, syntax_result, context).getActions(false);
+}
 }
 
+namespace KafkaStream
+{
 ChunkSharder::ChunkSharder(ContextPtr context, const Block & header, const ASTPtr & sharding_expr_ast)
```
chenziliang: `ContextPtr context` as last parameter

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -20,37 +21,39 @@ extern const int CANNOT_WRITE_TO_KAFKA;
...
 {
-    assert(sharding_expr_ast);
-
-    ASTPtr query = sharding_expr_ast;
-    auto syntax_result = TreeRewriter(context).analyze(query, header.getNamesAndTypesList());
-    sharding_expr = ExpressionAnalyzer(query, syntax_result, context).getActions(false);
+    sharding_expr = buildExpression(context, header, sharding_expr_ast);
```
chenziliang: Nit: since Context and Block header are used only once in ctor, we may simplify ChunkSharder to have less dependency and i think it is time to refactor it out in different source files (next PR)


```
ChunkSharder::ChunkSharder(ExpressionActionsPtr sharding_expr_, const ASTPtr & sharding_expr_ast)
    : sharding_expr(std::move(sharding_expr_))
{
  ...
}
```

On call side, we do 

```
ChunkSharder sharder(buildExpression(context, header, sharding_expr_ast), sharding_expr_ast);
```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -20,37 +21,39 @@ extern const int CANNOT_WRITE_TO_KAFKA;
...
 
     sharding_key_column_name = sharding_expr_ast->getColumnName();
 
     if (auto * shard_func = sharding_expr_ast->as<ASTFunction>())
     {
         if (shard_func->name == "rand" || shard_func->name == "RAND")
```
chenziliang: Use boost case insensitive compare ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -203,23 +204,15 @@ KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr conte
...
-
-    /// With partitioner callback, we can get the up-to-date partition count w/o additional effort.
-    rd_kafka_topic_conf_set_partitioner_cb(topic_conf, &KafkaSink::onPartitioning);
-
+    rd_kafka_conf_set_opaque(conf, this); /* needed by onMessageDelivery */
     rd_kafka_conf_set_dr_msg_cb(conf, &KafkaSink::onMessageDelivery);
 
-    producer = klog::KafkaPtr(rd_kafka_new(RD_KAFKA_PRODUCER, conf, errstr, sizeof(errstr)), rd_kafka_destroy);
+    size_t value_size = 8;
+    char topic_refresh_interval_ms_value[8]{'\0'}; /* max: 3600000 */
```
chenziliang: Lets do `/// max refresh interval: 3600000`

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -228,31 +221,104 @@ KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr conte
         throw Exception("Failed to create kafka handle", klog::mapErrorCode(rd_kafka_last_error()));
     }
 
-    topic = klog::KTopicPtr(rd_kafka_topic_new(producer.get(), kafka->topic().c_str(), nullptr), rd_kafka_topic_destroy);
-    wb = std::make_unique<WriteBufferFromKafka>(topic.get());
+    topic.reset(rd_kafka_topic_new(producer.get(), kafka->topic().c_str(), nullptr));
+    wb = std::make_unique<WriteBufferFromKafkaSink>([this](char * pos, size_t len) { addMessageToBatch(pos, len); });
 
     String data_format = kafka->dataFormat();
```
chenziliang, line 209: i saw some other place, we assert non-empty, shall do same here ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -228,31 +221,104 @@ KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr conte
...
+    {
+        message_key_expr = buildExpression(context, header, message_key_ast);
+        const auto & sample_block = message_key_expr->getSampleBlock();
+        /// The last column is the key column, the others are required columns (to be used to calculate the key value).
+        message_key_column_pos = message_key_expr->getResultPositions().back();
+        auto message_key_column_name = sample_block.getColumnsWithTypeAndName().back().name;
+        /// If the key column already exists in the head, which means the message key is
+        /// one of the column in the stream, in this case, it should not delete the column.
+        /// Otherwise, it's a computed column, it should be deleted from the block.
+        delete_message_key_column = !header.tryGetPositionByName(message_key_column_name).has_value();
```
chenziliang: `header.contains(...)` or `header.has(...)` i didn't recall which one

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -228,31 +221,104 @@ KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr conte
...
+    }
+
+    if (one_message_per_row)
+    {
+        /// The callback allows `IRowOutputFormat` based formats produce one Kafka message per row.
         writer = FormatFactory::instance().getOutputFormat(
             data_format, *wb, header, context, [this](auto & /*column*/, auto /*row*/) { wb->next(); });
+        if (!static_cast<IRowOutputFormat*>(writer.get()))
```
chenziliang: We shall use `dynamic_cast<...>`

chenziliang: Or even, we can assert `data_format` instead of this dynamic_cast ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -228,31 +221,104 @@ KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr conte
...
 
+    /// Polling message deliveries.
     polling_threads.scheduleOrThrowOnError([this]() {
         while (!is_finished.test())
             if (auto n = rd_kafka_poll(producer.get(), POLL_TIMEOUT_MS))
                 LOG_TRACE(log, "polled {} events", n);
     });
+
+    /// Monitor partition changes.
+    metadata_threads.scheduleOrThrowOnError([this, topic_refresh_interval_ms]() {
```
chenziliang: I was thinking we shall merge `metadata_threads / polling message deliveries threads` in one

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -228,31 +221,104 @@ KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr conte
...
+        auto last_refresh_at {std::chrono::steady_clock::now()};
+
+        while (!is_finished.test())
+        {
+            /// Do not sleep for topic_refresh_interval_ms, it will block cancelling the query.
+            std::this_thread::sleep_for(std::chrono::milliseconds(500));
+
+            auto now {std::chrono::steady_clock::now()};
```
chenziliang: Use `Stopwatch`

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -228,31 +221,104 @@ KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr conte
...
     });
+
+    /// Monitor partition changes.
+    metadata_threads.scheduleOrThrowOnError([this, topic_refresh_interval_ms]() {
+        auto last_refresh_at {std::chrono::steady_clock::now()};
+
+        while (!is_finished.test())
+        {
+            /// Do not sleep for topic_refresh_interval_ms, it will block cancelling the query.
+            std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
chenziliang: With merge, we don't need sleep

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -263,14 +329,73 @@ void KafkaSink::consume(Chunk chunk)
...
+    keys_queue.clear();
+
+    /// When one_message_per_row is set to true, one Kafka message will be generated for each row.
+    /// Otherwise, all rows in the same block will be in the same kafka message.
+    if (one_message_per_row)
+        current_batch.reserve(chunk.rows());
+    else
+        current_batch.reserve(blocks.size());
+
     for (auto & blockWithShard : blocks)
```
chenziliang: `block_with_shard`

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -263,14 +329,73 @@ void KafkaSink::consume(Chunk chunk)
...
     {
-        /// Since we set AutoFlush on writer, it makes sure that `writer->write` will call
-        /// `wb->nextImpl` and waits for it finishes, it's safe to call `wb->write_to_partition`
-        /// before `writer->write`.
-        wb->setTargetPartition(blockWithShard.shard);
-        writer->write(blockWithShard.block);
+        next_partition = blockWithShard.shard;
+        /// If message key is not used, and it's using random paritioning,
+        /// simply use round robin to avoid calling the paritioner function.
+        if (next_partition == RD_KAFKA_PARTITION_UA && !message_key_expr)
```
chenziliang: Actually can we just use `RD_KAFKA_PARTITION_UA` ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -263,14 +329,73 @@ void KafkaSink::consume(Chunk chunk)
...
+            next_partition = next_partition_counter++ % partition_cnt;
+
+        if (!message_key_expr)
+        {
+            writer->write(blockWithShard.block);
```
chenziliang: do we need set target partition before write otherwise next_partition calc is wasted ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -263,14 +329,73 @@ void KafkaSink::consume(Chunk chunk)
...
+
+        /// Compute and collect message keys and removed the key column from the block after executing the expression
+        message_key_expr->execute(blockWithShard.block);
+        const auto & message_key_column {blockWithShard.block.getByPosition(message_key_column_pos).column};
+
+        // Collect all the message keys for creating the messages for the block.
+        size_t rows {message_key_column->size()};
+        for (size_t i = 0; i < rows; ++i)
+        {
+            /// Since the capacity is big enough, calling push_back won't invalidate iterators.
```
chenziliang: Do we depend on iterator invalidation here ? If we do, it appears pretty error prone

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -228,31 +221,104 @@ KafkaSink::KafkaSink(const Kafka * kafka, const Block & header, ContextPtr conte
         throw Exception("Failed to create kafka handle", klog::mapErrorCode(rd_kafka_last_error()));
     }
 
-    topic = klog::KTopicPtr(rd_kafka_topic_new(producer.get(), kafka->topic().c_str(), nullptr), rd_kafka_topic_destroy);
-    wb = std::make_unique<WriteBufferFromKafka>(topic.get());
+    topic.reset(rd_kafka_topic_new(producer.get(), kafka->topic().c_str(), nullptr));
+    wb = std::make_unique<WriteBufferFromKafkaSink>([this](char * pos, size_t len) { addMessageToBatch(pos, len); });
```
chenziliang, line 207: `wb` doesn't have to be a data member ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -263,14 +329,73 @@ void KafkaSink::consume(Chunk chunk)
...
+        for (size_t i = 0; i < rows; ++i)
+        {
+            /// Since the capacity is big enough, calling push_back won't invalidate iterators.
+            keys_queue.push_back(message_key_column->getDataAt(i));
+        }
+
+        /// If the expression added a new column holds the key values, it should not be part of the message payload.
+        /// we cannot erase the column from blockWithShard.block here, because it could lead to heap-use-after-free
+        /// error when rd_kafka_produce_batch tries to read the keys.
+        if (delete_message_key_column)
```
chenziliang: There are several things we can improve probably
1. `keys_queue`, we probably can use a std::vector<...> without pop_front. Meaning, populate once and use index to access the message key for each row in `addMessageToBatch`
2. temp `Block blk` copy, we can choose remove the message key column but hold on to the key column separately instead copying the whole block

Something like this

```
...
private:
    size_t current_batch_row = 0; /// used in addMessageToBatch to access \keys
    std::vector<StringRef> keys; 
```

```
KafkaSink::consume(... chunk)
{
    current_batch_row = 0;
    {
        /// Release keys memory and reserve again to avoid big memory hanging around forever.
        std:vector<std::string_view> tmp;
        keys.swap(tmp);
        keys.reserve(chunk.size());
    }

    message_key_expr->execute(blockWithShard.block);
    ColumnPtr key_column = block_with_shard.block.getByPosition(message_key_column_pos).column;

    // Collect all the message keys for creating the messages for the block.
    size_t rows {message_key_column->size()};
    for (size_t i = 0; i < rows; ++i)
        keys_queue.push_back(key_column->getDataAt(i));

    if (delete_message_key_column)
        block_with_shard.block.erase(message_key_column_pos);

    writer->write(blockWithShard.block);

    ...
}

void KafkaSink::addMessageToBatch(char * pos, size_t len)
{
    StringRef key = keys.empty() ? "" : keys[current_batch_row++];

     ....
    void * payload = malloc(len);
    std::memcpy(payload, pos, len);

    current_batch.push_back(rd_kafka_message_t{
        .partition = next_partition,
        .payload = payload,
        .len = len,
        .key = key.data(),
        .key_len = key.size(),
    });

    ++state.outstandings;
}

```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -290,12 +415,33 @@ void KafkaSink::onFinish()
...
+
+void KafkaSink::onMessageDelivery(rd_kafka_t * /* producer */, const rd_kafka_message_t * msg, void * opaque)
+{
+    static_cast<KafkaSink *>(opaque)->onMessageDelivery(msg);
+}
+
+void KafkaSink::onMessageDelivery(const rd_kafka_message_t * msg)
+{
+    /// Can't rely on librdkafka freeing up the payloads with RD_KAFKA_MSG_F_FREE, it will throw:
+    ///   AddressSanitizer: alloc-dealloc-mismatch (operator new [] vs free)
```
chenziliang: We shall move the ownership to librdkafka by using malloc instead of new

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -290,12 +415,33 @@ void KafkaSink::onFinish()
...
+    char * payload = static_cast<char *>(msg->payload);
+    delete[] payload;
```
chenziliang: so, we don't need delete by ourselves when move the memory to librdkafka.

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -51,24 +51,22 @@ class Kafka final : public StorageExternalStreamImpl
...
     void cacheVirtualColumnNamesAndTypes();
     std::vector<Int64> getOffsets(const SeekToInfoPtr & seek_to_info, const std::vector<int32_t> & shards_to_query) const;
+    void validateMessageKey(const String & message_key, IStorage * storage, ContextPtr context);
     void validate(const std::vector<int32_t> & shards_to_query = {});
-
     static std::vector<int32_t> parseShards(const std::string & shards_setting);
 
-private:
     StorageID storage_id;
+    const ASTs engine_args;
```
chenziliang: don't have to be const

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -230,6 +241,31 @@ std::vector<int32_t> Kafka::parseShards(const std::string & shards_setting)
     return specified_shards;
 }
 
+void Kafka::validateMessageKey(const String & message_key_, IStorage * storage, ContextPtr context)
```
chenziliang: if necessary `const ContextPtr &`

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -230,6 +241,31 @@ std::vector<int32_t> Kafka::parseShards(const std::string & shards_setting)
...
+    auto syntax_result = TreeRewriter(context).analyze(message_key_ast, storage->getInMemoryMetadata().getColumns().getAllPhysical());
+    auto analyzer = ExpressionAnalyzer(message_key_ast, syntax_result, context).getActions(true);
+    const auto & block = analyzer->getSampleBlock();
+    if (block.columns() != 1)
+        throw Exception(ErrorCodes::INVALID_SETTING_VALUE, "message_key expression must return exactly one column");
+
+    auto type_id = block.getByPosition(0).type->getTypeId();
+    if (type_id != TypeIndex::String && type_id != TypeIndex::FixedString)
+        throw Exception(ErrorCodes::INVALID_SETTING_VALUE, "message_key must have type of string");
+
```
chenziliang: let's remove this empty line

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: The output format takes a buffer reference (i.e. it does not own the buffer), thus someone else must own the buffer, and `KafkaSink` should be the one.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Here is where we iniitalize `data_format` and make sure it will not be empty, so that we can `assert` is not empty in other places. But not here.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: `assert` is not used because it does not generate meaningful error messages and that would make it difficult for users to figure out what happened.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Combined into one thread.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Eventhough, we still need to free the payload memory by ourselves if the message was failed to be published. I.e. 
```c++
if (msg->err)
{
   /// free the memory here
}
```

So I think it's easier just always free the memory.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: LIke I mentioned in the below comment, since we need to free up the memory in error cases anyways, I prefer not to let librdkafka own that memory as I don't see any good reason why we should do this. Let me know if you have different opinions.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: `next_partition` is a member variable, it's used in `addMessageToBatch`.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Yes we can, that we simply rely on librdkafka to calculate the parition ID.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: No, that comment was outdated and I forgot to remove it.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
chenziliang: We shall let librdkafka to own it, otherwise too error prone

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
chenziliang: Please check exiting code in StreamShard how we manage it

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Updated, please check.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Updated, please check.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -38,17 +38,10 @@ ExpressionActionsPtr buildExpression(ContextPtr context, const Block & header, c
...
-    sharding_expr = buildExpression(context, header, sharding_expr_ast);
-
-    sharding_key_column_name = sharding_expr_ast->getColumnName();
-
-    if (auto * shard_func = sharding_expr_ast->as<ASTFunction>())
-    {
-        if (shard_func->name == "rand" || shard_func->name == "RAND")
-            random_sharding = true;
-    }
+    sharding_expr = sharding_expr_;
```
chenziliang: we could do 

```
: sharding_exp(sharding_expr_),
sharding_key_column_name(column_name)
```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -304,8 +287,8 @@ void KafkaSink::addMessageToBatch(char * pos, size_t len)
         .partition = next_partition,
         .payload = const_cast<void *>(static_cast<const void *>(payload)),
         .len = len,
-        .key = const_cast<void *>(static_cast<const void *>(key.data())),
-        .key_len = key.size(),
+        .key = const_cast<void *>(static_cast<const void *>(key.data)),
```
chenziliang: do we need all of this cast ?

chenziliang: Same for payload

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -21,13 +21,16 @@ namespace KafkaStream
...
 public:
-    ChunkSharder(ContextPtr context, const Block & header, const ASTPtr & sharding_expr_ast);
+    ChunkSharder(ExpressionActionsPtr sharding_expr_, const String & column_name);
     ChunkSharder();
 
     BlocksWithShard shard(Block block, Int32 shard_cnt) const;
 
 private:
-    Int32 getNextShardIndex(Int32 /*shard_cnt*/) const noexcept { /* do not specify the partition ID */ return RD_KAFKA_PARTITION_UA; }
+    Int32 getNextShardIndex(Int32 /*shard_cnt*/) const noexcept {
```
chenziliang: clang-format

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -59,13 +59,14 @@ class Kafka final : public StorageExternalStreamImpl
...
-    Poco::Logger * log;
 
     NamesAndTypesList virtual_column_names_and_types;
 
     std::mutex shards_mutex;
     int32_t shards = 0;
 
     ASTPtr message_key_ast;
+
+    Poco::Logger * log;
```
chenziliang: `logger`

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -302,28 +296,39 @@ void KafkaSink::consume(Chunk chunk)
...
-        keys_for_current_batch.swap(keys);
+        if (!keys_for_current_batch.empty())
+        {
+            std::vector<StringRef> keys;
+            keys_for_current_batch.swap(keys);
+        }
         keys_for_current_batch.reserve(chunk.rows());
         current_batch_row = 0;
     }
+    if (!current_batch.empty())
```
chenziliang, line 312: nit : coud we use blank line to separate these logic groups and put a comment there like

```
/// We do swap with empty std::vector here to avoid some big underlying memory hang out there forever.
/// since std::vector::clear still holds on to its allocated memory
```

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -335,35 +340,54 @@ void KafkaSink::consume(Chunk chunk)
...
-                throw Exception(klog::mapErrorCode(msg.err), rd_kafka_err2str(msg.err));
+    rd_kafka_resp_err_t err {RD_KAFKA_RESP_ERR_NO_ERROR};
+    for (size_t i = 0; i < current_batch.size(); ++i)
+        if (current_batch[i].err)
+            err = current_batch[i].err;
+        else
+            batch_payload[i].release(); /// payload of messages which are succesfully handled by rd_kafka_produce_batch will be free'ed by librdkafka
+
+    /// Clean up all the bookkeepings for the batch.
+    std::vector<rd_kafka_message_t> batch;
```
chenziliang, line 381: Since we do clean up here, do we need do again at line 309

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: The difference here is that `key.data` has type of `const char *`, if we remove the `const_cast`, it will report `Cannot initialize a member subobject of type 'void *' with an lvalue of type 'const char *'. But the `static_cast` can be removed.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Just in case the function call gets interrupted somehow and cause the cleanup skipped ( not 100% sure if that will happen or not ). As discussed, keep this as it is for now.

PR #436, opened 2023-12-21
enhancement: properly produce raw format data to Kafka external streams

### LINKED ISSUE #435: Write to Kafka external stream with raw data format should only send the raw data to Kafka

Currently, when `data_format` is not speicified in `CREATE EXTERNAL STREAM` with `type='kafka'` and there is only one string value column, it uses the Raw data format.

When reading from such a stream, the behavior is correct, it uses the message payload as-is as the column value. However, when writing to such a stream, it writes a JSON object, something like `{"data": "<the-raw-value>"}`. This is not the correct raw format behavior, this is exactly the same as using `JSONEachRow` format. The correct behavior for raw writing is to just write the string value as the Kafka message payload without adding any other thing to it. Simply put with the following SQL:

```sql
INSERT INTO my_raw_kafka_external_stream SELECT FROM my_raw_kafka_external_stream
```

should keep producing the unique message again and again to the Kafka topic.

zliang-min:
PR checklist:
- Did you run ClangFormat ?

closes #435 

Please check the issue description for the desired behavior for raw writting.

### DIFF
--- a/src/Storages/ExternalStream/Kafka/Kafka.cpp
+++ b/src/Storages/ExternalStream/Kafka/Kafka.cpp
@@ -183,24 +183,19 @@ void Kafka::calculateDataFormat(const IStorage * storage)
     if (!data_format.empty())
         return;
 
+    /// If there is only one column and its type is a string type, use RawBLOB. Use JSONEachRow otherwise.
     auto column_names_and_types{storage->getInMemoryMetadata().getColumns().getOrdinary()};
-    if (column_names_and_types.size() != 1)
-        throw Exception(
-            ErrorCodes::INVALID_SETTING_VALUE,
-            "`data_format` settings is empty but the Kafka external stream definition has multiple columns. Proton doesn't know how to "
-            "parse Kafka messages without a data format.");
-
-    auto type = column_names_and_types.begin()->type;
-    if (type->getTypeId() == TypeIndex::String || type->getTypeId() == TypeIndex::FixedString)
+    if (column_names_and_types.size() == 1)
     {
-        /// no-op
+        auto type = column_names_and_types.begin()->type;
+        if (type->getTypeId() == TypeIndex::String || type->getTypeId() == TypeIndex::FixedString)
+        {
+            data_format = "RawBLOB";
+            return;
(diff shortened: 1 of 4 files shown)


src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -180,24 +180,19 @@ void Kafka::calculateDataFormat(const IStorage * storage)
...
-            ErrorCodes::INVALID_SETTING_VALUE,
-            "`data_format` settings is empty but the Kafka external stream definition has multiple columns. Proton doesn't know how to "
-            "parse Kafka messages without a data format.");
-
-    auto type = column_names_and_types.begin()->type;
-    if (type->getTypeId() == TypeIndex::String || type->getTypeId() == TypeIndex::FixedString)
+    if (column_names_and_types.size() == 1)
     {
-        /// no-op
+        auto type = column_names_and_types.begin()->type;
```
chenziliang: auto type = [column_names_and_types.begin()->type->getTypeId();

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -180,24 +180,19 @@ void Kafka::calculateDataFormat(const IStorage * storage)
...
+        if (type->getTypeId() == TypeIndex::String || type->getTypeId() == TypeIndex::FixedString)
+        {
+            data_format = "RawBLOB";
```
chenziliang, line 193: Is it `RawBlob` ?

chenziliang:
Other changes look good

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: No, it's `RawBLOB`.

chenziliang: It is our new format or existing one since it doesn't seem follow our name convention

jovezhong:

jovezhong:

PR #449, opened 2023-12-27
change the data processing order in changelog

### LINKED ISSUE #448: change the data processing order in changelog
Consider the following scenario:
```sql
drop stream if exists test_kv;
create stream if not exists test_kv(id int, value int) primary key id settings mode='changelog_kv';
select count_distinct(value) from test_kv;
insert into test_kv(id, value, _tp_delta) values(2, 1, 1)(2,1,-1)(3,2,1)(3,2,-1);
```
I insert a piece of data with a primary key of 2 and then delete it, and then insert a piece of data with a primary key of 3 and then delete it.So in this chunk, there is no new data.The result of `count_distinct` should be 0 instead of 2(1.3.28-rc will output 2)
```
SELECT
  count_distinct(value)
FROM
  test_kv

Query id: 7e0d70e6-ec42-47cb-81bd-f650d2326cb9

┌─count_distinct(value)─┐
│                     2 │
└───────────────────────┘
```

lizhou1111:
change the data processing order in changelog.
For example: 
```sql
drop stream if exists test_kv;
create stream if not exists test_kv(id int, value int) primary key id settings mode='changelog_kv';
select count_distinct(value) from test_kv;
insert into test_kv(id, value, _tp_delta) values(2, 1, 1)(2, 1, -1)(3, 2, 1)(3, 2, -1);
```
The output  will be `2`, it should be `0`.Because a piece of data is inserted and then deleted, there should be no data.The `sum_distinct` will encountered the same problem
 
The previous processing strategy was to divide two chunks according to the delta value. The data with a delta value of -1 is placed in the first chunk, and the data with a delta value of 1 is placed in the second chunk(take the example above, it will be `[(2, 1, -1), (3, 2, -1),(2, 1, 1), (3, 2, 1)]`). Therefore, the data with a delta value of -1 will be placed in the first chunk. will be processed first, which is inconsistent with the order of data input.

The current strategy is putting consecutive data with the same _tp_delta value in a chunk.
If the input chunk delta flags are `[1, 1, 1, -1, -1, 1, 1, 1]`.We will split into 3 chunks:`[[1, 1, 1], [-1, -1], [1, 1, 1]]`.
This not only ensures that the order of data processing is consistent with the input, but also ensures that the _tp_delta values in the same chunk are the same. 

this closes #448

### DIFF
--- a/src/Processors/Transforms/Streaming/ChangelogTransform.cpp
+++ b/src/Processors/Transforms/Streaming/ChangelogTransform.cpp
@@ -145,18 +145,14 @@ void ChangelogTransform::work()
         return;
     }
 
-    IColumn::Selector selector(rows, 0);
-    for (size_t i = 0; auto delta : delta_flags)
-        selector[i++] = delta > 0;
-
     /// cut every column in chunk_columns and put them into a new chunk
-    auto cut_cols_into_chunk = [&chunk_columns, this, &selector](UInt64 & start_pos, UInt64 end_pos) {
+    auto cut_cols_into_chunk = [&chunk_columns, this, &delta_flags](UInt64 & start_pos, UInt64 end_pos) {
         Chunk chunk_output;
 
         for (const auto & col : chunk_columns)
             chunk_output.addColumn(col->cut(start_pos, end_pos - start_pos));
 
-        if (!selector[start_pos])
+        if (delta_flags[start_pos] < 0)
         {
             /// retract chunk
             chunk_output.setRetractedDataFlag();
@@ -164,6 +160,7 @@ void ChangelogTransform::work()
         }
         else
         {
+            /// update chunk
             chunk_output.setChunkContext(input_data.chunk.getChunkContext());
             this->transformChunk(chunk_output);
         }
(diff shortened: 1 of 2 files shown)


src/Processors/Transforms/Streaming/ChangelogTransform.cpp
```diff
@@ -149,26 +149,46 @@ void ChangelogTransform::work()
...
+    for (size_t end_pos = 0; end_pos < selector.size(); ++end_pos)
     {
-        chunks[0].setRetractedDataFlag();
-        transformChunk(chunks[0]);
+        if (selector[end_pos] != selector[start_pos])
+        {
+            cut_cols_into_chunk(start_pos, end_pos);
+            start_pos = end_pos;
+        }
     }
```
yl-lisen: we can only loop once to calculate the ranges for each parts. the selector is not needed now.
```c++
for (size_t i = 0; auto delta : delta_flags)
   /// ... calculate the range parts

for (auto & part : parts)
    /// cut each part.
```

yl-lisen: or  in one loop:
```c++
for (size_t i = 0; auto delta : delta_flags)
   /// calculate the range current part
   /// cut current part.
```

lizhou1111: yes, selector is not needed

yl-lisen (approved):

PR #465, opened 2024-01-02
support tracking updated aggregated states

### LINKED ISSUE #446: Support tracking whether the aggregation window or group is changed
**Use case**

Choose a good way to track whether the aggregation window or group changes, which can be used in:
- emit changelog for aggregated result (supported, but there are some bugs)
- only emit changed windows (TODO)
- only emit changed groups (TODO)
- incremental checkpointing (TODO)

**Additional context**

yl-lisen:
This close #446 

- Add updates tracking data (tracking changes/ updates .etc)
- Unified versioned serialize and deserialize of aggregator
- Remove some useless implementation in `Aggregator`


src/Common/serde.h
```diff
@@ -27,6 +27,27 @@ void ALWAYS_INLINE deserialize(S & s, RB & rb, VersionType version, Args &&... a
     s.deserialize(rb, version, std::forward<Args>(args)...);
 }
 
+/// With owned versions
+template <typename S, typename WB, typename... Args>
```
chenziliang: Let's use a readable template name like WriteBuffer, ReadBuffer etc or at least comment, otherwise it could be very hard to maintain.

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -130,6 +129,7 @@ struct AggregatedDataVariants : private boost::noncopyable
     /// Pools for states of aggregate functions. Ownership will be later transferred to ColumnAggregateFunction.
     Arenas aggregates_pools;
     Arena * aggregates_pool{};    /// The pool that is currently used for allocation.
+    std::unique_ptr<Arena> retracted_pool;  /// Use an independent pool to manage retracted data, which will be cleared after each finalization
```
chenziliang: `an independent` -> `separate`

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -371,6 +371,17 @@ struct AggregatedDataVariants : private boost::noncopyable
         /// proton: ends;
     }
 
+    void reset();
+
+    void resetAggregatesPool()
+    {
+        aggregates_pools = Arenas(1, std::make_shared<Arena>());
+        aggregates_pool = aggregates_pools.back().get();
+        aggregates_pool->enableRecycle(true);
```
chenziliang, line 380: For global aggregation, do we need enable recycle ?

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -650,6 +666,9 @@ class Aggregator final
         size_t window_keys_num;
 
         WindowParamsPtr window_params;
+
+        bool tracking_changes = false;
```
chenziliang: Let's comment what's for `tracking_changes` and `tracking_udpates`

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -787,25 +809,39 @@ class Aggregator final
       *            SELECT avg(i) AS i, sum(k) AS k FROM my_stream GROUP BY device_id <-- first level global aggr, don't prune states
       *          );
       */
-    BlocksList convertToBlocks(AggregatedDataVariants & data_variants, bool final, ConvertAction action, size_t max_threads) const;
-    BlocksList mergeAndConvertToBlocks(ManyAggregatedDataVariants & data_variants, bool final, ConvertAction action, size_t max_threads) const;
+    BlocksList convertToBlocks(AggregatedDataVariants & data_variants, bool final, bool clear_states, size_t max_threads) const;
```
chenziliang: Let's document `\param clear_states`
Moving forward, let's comment the method and method parameters as much as possible.

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -787,25 +809,39 @@ class Aggregator final
...
     ///                            =>  result block - [00:00, 00:04)
     ///   gcd_bucket2 - [00:02, 00:04)
     Block spliceAndConvertBucketsToBlock(
-        AggregatedDataVariants & variants, bool final, ConvertAction action, const std::vector<Int64> & gcd_buckets) const;
+        AggregatedDataVariants & variants, bool final, bool clear_states, const std::vector<Int64> & gcd_buckets) const;
     Block mergeAndSpliceAndConvertBucketsToBlock(
-        ManyAggregatedDataVariants & variants, bool final, ConvertAction action, const std::vector<Int64> & gcd_buckets) const;
+        ManyAggregatedDataVariants & variants, bool final, bool clear_states, const std::vector<Int64> & gcd_buckets) const;
+
+    /// Convert the `updated data` (different with `normal data`)
```
chenziliang: Let's comment / document in detail

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -787,25 +809,39 @@ class Aggregator final
...
+    BlocksList convertUpdatedToBlocks(AggregatedDataVariants & data_variants) const;
+
+    /// \return: merged updated data if exists
+    /// NOTE: The merged data is as `normal data`, which should use `convertToBlocks` to convert
```
chenziliang: Don't quite get what it actually mean for the `NOTE`, could you please elaborate ?

chenziliang: I think we probably can delete this NOTE

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -787,25 +809,39 @@ class Aggregator final
...
+    AggregatedDataVariantsPtr mergeUpdatedGroups(ManyAggregatedDataVariants & data_variants) const;
 
-    /// Used for merge changed groups and return the <retracted_state, aggregated_state> of changed groups
-    std::pair<AggregatedDataVariantsPtr, AggregatedDataVariantsPtr>
-    mergeRetractedGroups(ManyAggregatedDataVariants & aggregated_data, ManyAggregatedDataVariants & retracted_data) const;
+    /// Convert the `retracted data` (different with `normal data`)
+    BlocksList convertRetractedToBlocks(AggregatedDataVariants & data_variants) const;
+
+    /// \return: merged retracted data if exists
+    /// NOTE: The merged data is as `normal data`, which should use `convertToBlocks` to convert
```
chenziliang: We probably can delete this NOTE

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -787,25 +809,39 @@ class Aggregator final
...
+    AggregatedDataVariantsPtr mergeRetractedGroups(ManyAggregatedDataVariants & data_variants) const;
+
+    /// Used for merge legacy retracted data into result
```
chenziliang: Do we still need this legacy ?

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -821,7 +857,7 @@ class Aggregator final
     /// Precondition: for all blocks block.info.is_overflows flag must be the same.
     /// (either all blocks are from overflow data or none blocks are).
     /// The resulting block has the same value of is_overflows flag.
-    Block mergeBlocks(BlocksList & blocks, bool final, ConvertAction action);
+    Block mergeBlocks(BlocksList & blocks, bool final, bool clear_states, bool only_updated);
```
chenziliang: Let's comment / document the parameters

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -1084,65 +1088,60 @@ class Aggregator final
...
+    void mergeAggregateStates(AggregateDataPtr & dst, AggregateDataPtr & src, Arena * arena, bool clear_states) const;
+
+    void destroyAggregateStates(AggregateDataPtr & place) const;
+
+    void serializeAggregateStates(const AggregateDataPtr & place, WriteBuffer & wb) const;
+    void deserializeAggregateStates(AggregateDataPtr & place, ReadBuffer & rb, Arena * arena) const;
+
+    void clearDataVariants(AggregatedDataVariants & data_variants) const;
+
+    /// @return does need abort ?
```
chenziliang: `/// \return true means ..., false means ...`

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -1207,30 +1206,31 @@ class Aggregator final
...
-    ///   STATE VERSION 1 - Legacy version
-    ///   STATE VERSION 2 - REVISION 1 (Enable revision)
-    static constexpr UInt64 STATE_V2_MIN_REVISION = 1;
+    void checkpoint(const AggregatedDataVariants & data_variants, WriteBuffer & wb) const;
+    void recover(AggregatedDataVariants & data_variants, ReadBuffer & rb) const;
 
-    void checkpoint(const AggregatedDataVariants & data_variants, WriteBuffer & wb);
-    void recover(AggregatedDataVariants & data_variants, ReadBuffer & rb);
+private:
+    /// [Version-3]
```
chenziliang: Let's remove it

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -1207,30 +1206,31 @@ class Aggregator final
...
+    void doCheckpointV3(const AggregatedDataVariants & data_variants, WriteBuffer & wb) const;
+    void doRecoverV3(AggregatedDataVariants & data_variants, ReadBuffer & rb) const;
 
-    void doCheckpoint(const AggregatedDataVariants & data_variants, WriteBuffer & wb);
-    void doRecover(AggregatedDataVariants & data_variants, ReadBuffer & rb);
+    /// [Version-2]
```
chenziliang: Let's remove it

src/Interpreters/Streaming/Aggregator.h
chenziliang: `retracted_pool` => `retract_pool`

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -103,6 +107,73 @@ Columns materializeKeyColumns(Columns & columns, ColumnRawPtrs & key_columns, co
 
     return materialized_columns;
 }
+
+Arena * getArena(AggregatedDataVariants & variants, AggregateStateType type)
```
chenziliang: `getArena` can be a member function of `AggregatedDataVariants` and we may use `getPool()` for convention

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -103,6 +107,73 @@ Columns materializeKeyColumns(Columns & columns, ColumnRawPtrs & key_columns, co
...
+        return blocks;
+    };
+
+    size_t num_threads = thread_pool ? std::min(thread_pool->getMaxThreads(), buckets.size()) : 1;
+    if (num_threads <= 1)
+        return converter(arena, nullptr, nullptr);
+
+    /// Process in parallel
+    for (size_t i = pools.size(); i < num_threads; ++i)
+        pools.push_back(std::make_shared<Arena>());
```
chenziliang: Do we need keep the arenas around which are used to parallel converting ? Or we can GC them after we are done ?

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -103,6 +107,73 @@ Columns materializeKeyColumns(Columns & columns, ColumnRawPtrs & key_columns, co
...
+    else
+        return variants.aggregates_pool;
+}
+
+template <typename BucketConverter>
+BlocksList concurrentBucketConvert(ThreadPool * thread_pool, const std::vector<Int64> & buckets, Arena * arena, Arenas & pools, BucketConverter && bucket_converter)
+{
+    std::atomic<UInt32> next_bucket_idx_to_merge = 0;
+    auto converter = [&](Arena * pool, ThreadGroupStatusPtr thread_group, const std::atomic_flag * cancelled) {
+        SCOPE_EXIT_SAFE(
```
chenziliang: We can refactor out line 124 ~ 130 to `scheduleOrThrowOnError` lambda param at line 164 like 

```
auto converter = [&](Arena * pool, const std::atomic_flag * cancelled) {
        BlocksList blocks;
        while (true)
        {
            if (cancelled && cancelled->test())
                break;

            UInt32 bucket_idx = next_bucket_idx_to_merge.fetch_add(1);
            if (bucket_idx >= buckets.size())
                break;

            auto bucket = buckets[bucket_idx];
            blocks.splice(blocks.end(), bucket_converter(bucket, pool));
        }
        return blocks;
};
```


```
    for (size_t thread_id = 0; thread_id < num_threads; ++thread_id) 
    {
            thread_pool->scheduleOrThrowOnError([&pools, thread_id, group = CurrentThread::getGroup(), results, &converter, &cancelled] {
                CurrentThread::attachToIfDetached(group);
                SCOPE_EXIT_SAFE({CurrentThread::detachQueryIfNotDetached();});
                (*results)[thread_id] = converter(pools[thread_id].get(), &cancelled);
            });
    }
```

src/Interpreters/Streaming/Aggregator.h
chenziliang: Let's use `initAggregatesPool()` or `resetAndCreateAggregatesPools()`

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -371,6 +371,17 @@ struct AggregatedDataVariants : private boost::noncopyable
...
+    void reset();
+
+    void resetAggregatesPool()
+    {
+        aggregates_pools = Arenas(1, std::make_shared<Arena>());
+        aggregates_pool = aggregates_pools.back().get();
+        aggregates_pool->enableRecycle(true);
+    }
+
+    void resetRetractedPool() { retracted_pool = std::make_unique<Arena>(); }
```
chenziliang: Same. `initRetractPool()` or `resetAndCreateRetractPool()`

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -371,6 +371,17 @@ struct AggregatedDataVariants : private boost::noncopyable
         /// proton: ends;
     }
 
+    void reset();
```
chenziliang, line 373: Let's give a comment here. 
```
\reset clean up all in memory states  in Aggregator and the corresponding arena pools used to hold these states
```

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -79,6 +80,9 @@ inline void initDataVariants(
     result.keys_size = params.keys_size;
     result.key_sizes = key_sizes;
     result.init(method_chosen);
+
+    if (params.tracking_changes)
+        result.resetRetractedPool();
```
chenziliang: `initRetractPool` is far less confusing.

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -143,6 +239,17 @@ void AggregatedDataVariants::convertToTwoLevel()
     }
 }
 
+void AggregatedDataVariants::serialize(WriteBuffer & wb, const Aggregator & aggregator_) const
+{
+    aggregator_.checkpoint(*this, wb);
```
chenziliang, line 229: AggregatedDataVariants has an aggregator point, what's the difference between the aggregator passed in and the one it points to ?

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -143,6 +239,17 @@ void AggregatedDataVariants::convertToTwoLevel()
...
+}
+
+void AggregatedDataVariants::deserialize(ReadBuffer & rb, const Aggregator & aggregator_)
+{
+    aggregator = &aggregator_;
```
chenziliang: I feel like we shall separate this assignment and to make it explicitly and get called else where before `deserialize(...)`. Basically it is not good to assign aggregator here

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -687,7 +708,9 @@ class Aggregator final
             group_by(streaming_group_by_),
             delta_col_pos(delta_col_pos_),
             window_keys_num(window_keys_num_),
-            window_params(window_params_)
+            window_params(window_params_),
+            tracking_changes(tracking_changes_),
+            tracking_updated(tracking_updated_)
```
chenziliang: `tracking_updates` ?

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -650,12 +770,43 @@ AggregatedDataVariants::Type Aggregator::chooseAggregationMethodTimeBucketTwoLev
...
+    assert(aggregate_data);
+    if constexpr (!skip_expanded_data)
+    {
+        if (expanded_data_type == ExpandedDataType::UpdatedWithRetracted)
+            new (aggregate_data) RetractedDataEx();
+        else if (expanded_data_type == ExpandedDataType::Updated)
+            new (aggregate_data) UpdatedDataEx();
+    }
+
+    if constexpr (use_compiled_functions)
```
chenziliang: Let's have a comment or comment it out for compiled function since it doesn't work for streaming processing in general for now ? We can't GC the states etc properly

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -650,12 +770,43 @@ AggregatedDataVariants::Type Aggregator::chooseAggregationMethodTimeBucketTwoLev
 }
 /// proton: ends
 
-template <bool skip_compiled_aggregate_functions>
+template <bool use_compiled_functions, bool skip_expanded_data>
```
chenziliang: We don't need a template parameter here but a function parameter

chenziliang:
```
void Aggregator::createAggregateStates(AggregateDataPtr & aggregate_data, bool prefix_with_updates_tracking_state)
{
    ...
    if (prefix_with_updates_tracking_state)
    {
        /// Add updates tracking state before function aggregate states.
        ...
    }
}
```

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -282,8 +389,21 @@ Aggregator::Aggregator(const Params & params_) : params(params_),  log(&Poco::Lo
...
+    else if (params.tracking_updated)
+    {
+        total_size_of_aggregate_states = sizeof(UpdatedDataEx);
+        align_aggregate_states = alignof(UpdatedDataEx);
+        expanded_data_type = ExpandedDataType::Updated;
+    }
+
     // aggregate_states will be aligned as below:
-    // |<-- state_1 -->|<-- pad_1 -->|<-- state_2 -->|<-- pad_2 -->| .....
+    // |<-- [ExpandedDataEx] -->||<-- state_1 -->|<-- pad_1 -->|<-- state_2 -->|<-- pad_2 -->| .....
```
chenziliang: Does it have to be at the head, can we do it at the tail ? Just wondering the pros and cons

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+#include <Common/serde.h>
+
+namespace DB
+{
+using AggregateDataPtr = char *;
+using ConstAggregateDataPtr = const char *;
+
+namespace Streaming
+{
+SERDE struct UpdatedDataEx
```
chenziliang: Lets rename it as `TrackingUpdates` or `TrackingUpdatesData`

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /// Used for tracking the group is empty or not
+    UInt32 final_count = 0;
+
+    /// Used for tracking the group is updated or not
+    bool updated_since_last_finalization = true;
+};
+
+SERDE struct RetractedDataEx : UpdatedDataEx
```
chenziliang: `RetractedDataEx` => `TrackingUpdatesWithRetract`

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+using ConstAggregateDataPtr = const char *;
+
+namespace Streaming
+{
+SERDE struct UpdatedDataEx
+{
+    static ALWAYS_INLINE UpdatedDataEx & data(AggregateDataPtr __restrict place) { return *reinterpret_cast<UpdatedDataEx *>(place); }
+    static ALWAYS_INLINE const UpdatedDataEx & data(ConstAggregateDataPtr __restrict place) { return *reinterpret_cast<const UpdatedDataEx *>(place); }
+
+    static ALWAYS_INLINE bool isEmpty(ConstAggregateDataPtr __restrict place) { return data(place).final_count == 0; }
```
chenziliang: `isEmpty()` -> `empty()`

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+    static ALWAYS_INLINE bool isUpdated(ConstAggregateDataPtr __restrict place) { return data(place).updated_since_last_finalization; }
```
chenziliang: `isUpdated()` -> `updated()`

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+    static ALWAYS_INLINE void setUpdated(AggregateDataPtr __restrict place) { data(place).updated_since_last_finalization = true; }
+    static ALWAYS_INLINE void resetUpdated(AggregateDataPtr __restrict place) { data(place).updated_since_last_finalization = false; }
+
+    static void addBatch(size_t row_begin, size_t row_end, AggregateDataPtr * places, const IColumn * delta_col)
+    {
+        if (delta_col == nullptr)
```
chenziliang: If we are tracking changes already, is `delta_col` always there ?

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+    {
+        --final_count;
+        updated_since_last_finalization = true;
+    }
+
+    /// Used for tracking the group is empty or not
+    UInt32 final_count = 0;
+
+    /// Used for tracking the group is updated or not
+    bool updated_since_last_finalization = true;
```
chenziliang: Why it is true by default ?

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+    ALWAYS_INLINE void negate()
+    {
+        --final_count;
+        updated_since_last_finalization = true;
+    }
+
+    /// Used for tracking the group is empty or not
+    UInt32 final_count = 0;
+
+    /// Used for tracking the group is updated or not
```
chenziliang:
```
/// Used to track if the target group tracked has updates since last finalization 
```

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+        updated_since_last_finalization = true;
+    }
+
+    ALWAYS_INLINE void negate()
+    {
+        --final_count;
+        updated_since_last_finalization = true;
+    }
+
+    /// Used for tracking the group is empty or not
```
chenziliang:
```
/// Used to track if the target to be tracked has zero sum changes
```

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+    UInt32 final_count = 0;
```
chenziliang: `final_count` => `UInt64 updates = 0`

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+    /// Used for tracking the group is updated or not
+    bool updated_since_last_finalization = true;
+};
+
+SERDE struct RetractedDataEx : UpdatedDataEx
+{
+    static ALWAYS_INLINE AggregateDataPtr & getRetracted(AggregateDataPtr & place) { return reinterpret_cast<RetractedDataEx *>(place)->retracted_data; }
+    static ALWAYS_INLINE bool hasRetracted(ConstAggregateDataPtr __restrict place) { return reinterpret_cast<const RetractedDataEx *>(place)->retracted_data; }
+
+    template <bool use_retracted_data>
```
chenziliang, line 103: This method look very weird since we are already in RetractedDataEx and t appears it shall not be a template again.

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+    static ALWAYS_INLINE AggregateDataPtr & getData(AggregateDataPtr & place)
+    {
+        if constexpr (use_retracted_data)
+            return getRetracted(place);
+        else
+            return place;
+    }
+
+    /// Used for tracking group changes
```
chenziliang:
```
/// Used to track changes for the target to be tracked
```

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+    /// Used for tracking the group is empty or not
+    UInt32 final_count = 0;
+
+    /// Used for tracking the group is updated or not
+    bool updated_since_last_finalization = true;
+};
+
+SERDE struct RetractedDataEx : UpdatedDataEx
+{
+    static ALWAYS_INLINE AggregateDataPtr & getRetracted(AggregateDataPtr & place) { return reinterpret_cast<RetractedDataEx *>(place)->retracted_data; }
```
chenziliang: `getRetract(...)`

src/Interpreters/Streaming/AggregateDataEx.h
chenziliang:
```
/// Compared with \TrackingUpdates, \TrackingUpdatesWithRetract has retract semantic
```

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+            return getRetracted(place);
+        else
+            return place;
+    }
+
+    /// Used for tracking group changes
+    AggregateDataPtr retracted_data = nullptr;
+};
+
+enum class ExpandedDataType : uint8_t
```
chenziliang:
```
enum class TrackingUpdatesType : uint8_t
{
    Updates = 0,
    UpdatesWithRetract = 1,
}
```

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+{
+    None = 0,
+    Updated = 1, /// Allow tracking group is empty or updated
```
chenziliang: we don't need the comment here since we comment in the struct already

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
...
+    bool updated_since_last_finalization = true;
+};
+
+SERDE struct RetractedDataEx : UpdatedDataEx
+{
+    static ALWAYS_INLINE AggregateDataPtr & getRetracted(AggregateDataPtr & place) { return reinterpret_cast<RetractedDataEx *>(place)->retracted_data; }
+    static ALWAYS_INLINE bool hasRetracted(ConstAggregateDataPtr __restrict place) { return reinterpret_cast<const RetractedDataEx *>(place)->retracted_data; }
+
+    template <bool use_retracted_data>
+    static ALWAYS_INLINE AggregateDataPtr & getData(AggregateDataPtr & place)
```
chenziliang: We probably need a func parameter instead of template parameter here for `use_retract_data`

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -719,23 +870,18 @@ template <typename Method>
     AggregateDataPtr overflow_row) const
 {
     typename Method::State state(key_columns, key_sizes, aggregation_state_cache);
+    assert(!no_more_keys);
```
chenziliang: if we assert, can we remove it from the function parameter once for all ?

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -787,25 +809,39 @@ class Aggregator final
...
+    BlocksList convertRetractedToBlocks(AggregatedDataVariants & data_variants) const;
+
+    /// \return: merged retracted data if exists
+    /// NOTE: The merged data is as `normal data`, which should use `convertToBlocks` to convert
+    AggregatedDataVariantsPtr mergeRetractedGroups(ManyAggregatedDataVariants & data_variants) const;
+
+    /// Used for merge legacy retracted data into result
+    void mergeRetractedInto(AggregatedDataVariants & result, AggregatedDataVariants && retracted_result) const;
+
+    bool hasExpandedData() const { return expanded_data_type != ExpandedDataType::None; }
```
chenziliang:
```
bool needTrackUpdates() const noexcept { return tracking_updates_type; }
```

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -1207,30 +1206,31 @@ class Aggregator final
...
-
     /// proton: starts
     void setupAggregatesPoolTimestamps(size_t row_begin, size_t row_end, const ColumnRawPtrs & key_columns, Arena * aggregates_pool) const;
 
-    inline bool shouldClearStates(ConvertAction action, bool final_) const;
+public:
+    /// Existed versions:
+    ///   STATE V1 - Legacy version (REVISION 1)
+    ///   STATE V2 - REVISION 1 (Enable revision)
+    ///   STATE V3 - REVISION 3 (Add expanded data)
```
chenziliang: `//// Add updates tracking state`

src/Interpreters/Streaming/AggregateDataEx.h
```diff
@@ -0,0 +1,124 @@
+#pragma once
```
chenziliang: Lets rename the file to `UpdatesTrackingData.h`

src/Interpreters/Streaming/Aggregator.h
chenziliang: Probably we can just use one `std::optional<TrackingUpdatesType>` to cover these 2 bools

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -904,6 +940,8 @@ class Aggregator final
 
     bool all_aggregates_has_trivial_destructor = false;
 
+    ExpandedDataType expanded_data_type = ExpandedDataType::None;
```
chenziliang: `std::optional<TrackingUpdatesType> tracking_updates_type;`

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -74,15 +76,11 @@ namespace Streaming
   *  best suited for different cases, and this approach is just one of them, chosen for a combination of reasons.
   */
 
-enum class ConvertAction : uint8_t
+enum class AggregateStateType
```
chenziliang: Once we have `std::optional<TrackingUpdatesType>`, do we still need this ?

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -282,8 +389,21 @@ Aggregator::Aggregator(const Params & params_) : params(params_),  log(&Poco::Lo
     total_size_of_aggregate_states = 0;
     all_aggregates_has_trivial_destructor = true;
 
+    if (params.tracking_changes)
```
chenziliang: Yes, we can just use `std::optional<TrackingUpdatesType>` for these 2 bools

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -719,23 +870,18 @@ template <typename Method>
...
-        {
-            return executeImplBatch<false, true>(method, state, aggregates_pool, row_begin, row_end, aggregate_instructions, overflow_row);
-        }
-        else
-#endif
-        {
-            return executeImplBatch<false, false>(method, state, aggregates_pool, row_begin, row_end, aggregate_instructions, overflow_row);
-        }
+    /// TODO: So far not support compiled functions with expanded data
+    if (compiled_aggregate_functions_holder && !hasExpandedData())
```
chenziliang: `!needTrackUpdates()`

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -821,66 +968,24 @@ template <bool no_more_keys, bool use_compiled_functions, typename Method>
...
-                    }
-
-#if defined(MEMORY_SANITIZER)
-
-                    /// We compile only functions that do not allocate some data in Arena. Only store necessary state in AggregateData place.
-                    for (size_t aggregate_function_index = 0; aggregate_function_index < aggregate_functions.size(); ++aggregate_function_index)
-                    {
-                        if (!is_aggregate_function_compiled[aggregate_function_index])
-                            continue;
+        assert(!no_more_keys);
```
chenziliang: We probably can clean up `no_more_keys` optimization in streaming aggregator once for all ?

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -1026,28 +1134,12 @@ template <bool use_compiled_functions>
         }
     }
 
-    return should_finalize;
-}
+    if (hasExpandedData())
+        UpdatedDataEx::addBatchSinglePlace(row_begin, row_end, res, aggregate_instructions ? aggregate_instructions->delta_column : nullptr);
```
chenziliang: if tracking updates is there, does it mean delta column shall aways be there otherwise there is something wrong during  query planning ?

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -1170,14 +1262,14 @@ std::pair<bool, bool> Aggregator::executeOnBlock(
...
     {
-        /// TODO: Enable compilation after investigation
-// #if USE_EMBEDDED_COMPILER
-//         if (compiled_aggregate_functions_holder)
-//         {
-//             executeWithoutKeyImpl<true>(result.without_key, row_begin, row_end, aggregate_functions_instructions.data(), result.aggregates_pool);
-//         }
-//         else
-// #endif
+        /// TODO: So far not support compiled functions with expanded data
```
chenziliang: Let's keep commenting out the JIT code

src/Interpreters/Streaming/AggregationUtils.cpp
```diff
@@ -0,0 +1,113 @@
...
+    {
+        if (!final)
+        {
+            const auto & aggregate_column_name = params.aggregates[i].column_name;
+            aggregate_columns[i] = res_header.getByName(aggregate_column_name).type->createColumn();
+
+            /// The ColumnAggregateFunction column captures the shared ownership of the arena with the aggregate function states.
+            ColumnAggregateFunction & column_aggregate_func = assert_cast<ColumnAggregateFunction &>(*aggregate_columns[i]);
+
+            for (auto & pool : aggregates_pools)
```
chenziliang, line 48: Let's add the original comments back and do we need call `column_aggregate_func.setKeepState(...)` as the original code does ?

src/Interpreters/Streaming/AggregationUtils.cpp
```diff
@@ -0,0 +1,113 @@
...
+            aggregate_columns_data[i]->reserve(rows);
+        }
+        else
+        {
+            final_aggregate_columns[i] = aggregate_functions[i]->getReturnType()->createColumn();
+            final_aggregate_columns[i]->reserve(rows);
+
+            if (aggregate_functions[i]->isState())
+            {
+                auto callback = [&](IColumn & subcolumn)
```
chenziliang: subclolumn -> column

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -1877,137 +1694,126 @@ void NO_INLINE Aggregator::convertToBlockImplFinal(
...
+    return finalizeBlock(params, getHeader(/* final */ true), std::move(out_cols), /* final */ true, places.size());
 }
 
-template <typename Method, typename Table>
-void NO_INLINE Aggregator::convertToBlockImplNotFinal(
-    Method & method,
-    Table & data,
-    std::vector<IColumn *>  key_columns,
-    AggregateColumnsData & aggregate_columns) const
+template <typename Method, bool use_compiled_functions, AggregateStateType type, typename Table>
```
chenziliang: It appears not a good idea to add another dimension to the template, our binary size is increasing so quickly and didn't see big benifits.

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -1566,59 +1445,60 @@ bool Aggregator::checkLimits(size_t result_size, bool & no_more_keys) const
...
+Block Aggregator::convertToBlockImpl(
+    Method & method, Table & data, Arena * arena, Arenas & aggregates_pools, bool final, size_t rows, bool clear_states, AggregateStateType type) const
 {
     if (data.empty())
-        return;
-
-    if (key_columns.size() != params.keys_size)
-        throw Exception{"Aggregate. Unexpected key columns size.", ErrorCodes::LOGICAL_ERROR};
+    {
+        auto && out_cols = prepareOutputBlockColumns(params, aggregate_functions, getHeader(final), aggregates_pools, final, rows);
```
chenziliang, line 1269: Personally, i still like the `previous prepareBlockAndFill*` method since it is very clear : prepare, fill and finalize in one go, and caller pass in the fill callback.  Is there a strong reason we need break it up to 3 functions since we will need pass around big chunk of return value and glue things together?

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -1566,59 +1445,60 @@ bool Aggregator::checkLimits(size_t result_size, bool & no_more_keys) const
...
     {
-        convertToBlockImplNotFinal(method, data, std::move(raw_key_columns), aggregate_columns);
+        assert(type == AggregateStateType::Normal);
+        res = convertToBlockImplNotFinal(method, data, aggregates_pools, rows);
     }
 
     /// In order to release memory early.
     /// proton: starts. For streaming aggr, we hold on to the states
-    if (clear_states)
+    if (clear_states && type == AggregateStateType::Normal)
```
chenziliang: we may need revisit how we evaluate clear_states, instead of patching every where with additional different checking logic. It would be quite hard to maintain.

cmake/autogenerated_versions.txt
```diff
@@ -2,7 +2,7 @@
 
 # NOTE: has nothing common with DBMS_TCP_PROTOCOL_VERSION,
 # only DBMS_TCP_PROTOCOL_VERSION should be incremented on protocol changes.
-SET(VERSION_REVISION 2)
+SET(VERSION_REVISION 3)
```
chenziliang: A minor version update or a major one

src/Common/HashTable/TimeBucketHashTable.h
```diff
@@ -108,7 +108,9 @@ class TimeBucketHashTable : private boost::noncopyable, protected Hash /// empty
     using ConstLookupResult = typename Impl::ConstLookupResult;
 
     /// FIXME, choose a better perf data structure
+    /// Usually we don't have too many time buckets
     std::map<Int64, Impl> impls;
+    std::unordered_map<Int64, bool/*updated*/> bucket_updated_flags;
```
chenziliang: `updated_buckets`

src/Common/HashTable/TimeBucketHashTable.h
```diff
@@ -289,6 +292,7 @@ class TimeBucketHashTable : private boost::noncopyable, protected Hash /// empty
         {
             DB::writeIntBinary(p.first);
             p.second.write(wb);
+            DB::writeBoolText(bucket_updated_flags[p.first], wb);
```
chenziliang: `DB::writeBinary`

src/Common/HashTable/TimeBucketHashTable.h
```diff
@@ -438,4 +454,45 @@ class TimeBucketHashTable : private boost::noncopyable, protected Hash /// empty
 
         return buckets;
     }
+
+    bool isUpdatedBucket(Int64 bucket_) const
```
chenziliang: `isBucketUpdated`

src/Common/HashTable/TimeBucketHashTable.h
```diff
@@ -438,4 +454,45 @@ class TimeBucketHashTable : private boost::noncopyable, protected Hash /// empty
...
+    {
+        auto it = bucket_updated_flags.find(bucket_);
+        if (it != bucket_updated_flags.end())
+            return it->second;
+
+        return false;
+    }
+
+    void resetUpdated(Int64 bucket_)
```
chenziliang: `resetUpdatedBucket(...)`

src/Common/HashTable/TimeBucketHashTable.h
```diff
@@ -438,4 +454,45 @@ class TimeBucketHashTable : private boost::noncopyable, protected Hash /// empty
...
+        if (it != bucket_updated_flags.end())
+            it->second = false;
+    }
+
+    void writeBucketUpdatedFlags(DB::WriteBuffer & wb) const
+    {
+        DB::writeVarUInt(bucket_updated_flags.size(), wb);
+        for (const auto & [bucket, updated] : bucket_updated_flags)
+        {
+            DB::writeIntBinary(bucket, wb);
```
chenziliang, line 479: Let's use `DB::writeVarInt / VarUInt / writeBinary` where possible and avoid full binary / text serde since they are slower and bigger

src/Common/HashTable/TimeBucketHashTable.h
```diff
@@ -438,4 +454,45 @@ class TimeBucketHashTable : private boost::noncopyable, protected Hash /// empty
...
+    }
+
+    void resetUpdated(Int64 bucket_)
+    {
+        auto it = bucket_updated_flags.find(bucket_);
+        if (it != bucket_updated_flags.end())
+            it->second = false;
+    }
+
+    void writeBucketUpdatedFlags(DB::WriteBuffer & wb) const
```
chenziliang: writeUpdatedBuckets(...)

src/Common/HashTable/TimeBucketHashTable.h
```diff
@@ -438,4 +454,45 @@ class TimeBucketHashTable : private boost::noncopyable, protected Hash /// empty
...
+    {
+        DB::writeVarUInt(bucket_updated_flags.size(), wb);
+        for (const auto & [bucket, updated] : bucket_updated_flags)
+        {
+            DB::writeIntBinary(bucket, wb);
+            DB::writeBoolText(updated, wb);
+        }
+    }
+
+    void readBucketUpdatedFlags(DB::ReadBuffer & rb)
```
chenziliang: readUpdatedBuckets(...)

src/Common/HashTable/TimeBucketHashTable.h
```diff
@@ -438,4 +454,45 @@ class TimeBucketHashTable : private boost::noncopyable, protected Hash /// empty
...
+    {
+        size_t size = 0;
+        DB::readVarUInt(size, rb);
+        bucket_updated_flags.clear();
```
chenziliang: we shall assert bucket_updated_flags is empty

src/Common/HashTable/TwoLevelHashTable.h
```diff
@@ -365,5 +385,31 @@ class TwoLevelHashTable :
...
+    void resetUpdated(Int64 bucket_)
+    {
+        bucket_updated_flags[bucket_] = false;
+    }
+
+    void writeBucketUpdatedFlags(DB::WriteBuffer & wb) const
+    {
+        DB::writeVarUInt(NUM_BUCKETS, wb);
+        for (const auto & elem : bucket_updated_flags)
+            DB::writeBoolText(elem, wb);
```
chenziliang: DB::writeBinary

src/Processors/Transforms/Streaming/AggregatingHelper.cpp
```diff
@@ -30,12 +30,19 @@ Chunk mergeBlocksToChunk(BlocksList && blocks)
     return merged_chunk;
 }
 
-Chunk convertToChunkImpl(AggregatedDataVariants & data, const AggregatingTransformParams & params, ConvertAction action)
+Chunk convertToChunkImpl(AggregatedDataVariants & data, const AggregatingTransformParams & params, AggregateStateType type)
```
chenziliang: As mentioned `AggregateStateType` can be normalized with `std::optional<TrackingUpdatesType>`

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -1084,65 +1088,60 @@ class Aggregator final
...
         Arena * aggregates_pool,
-        Method & retracted_method,
         Arena * retracted_pool,
         size_t row_begin,
         size_t row_end,
         ColumnRawPtrs & key_columns,
         AggregateFunctionInstruction * aggregate_instructions) const;
 
+    template <typename Method, bool is_two_level>
+    void mergeUpdatedGroupsImpl(ManyAggregatedDataVariants & non_empty_data, Arena * arena) const;
```
chenziliang: Let's use `Update` / `Retract` instead of `Updated / Retracted` for most places except `updated()` bool func

src/Processors/Transforms/Streaming/AggregatingHelper.h
```diff
@@ -38,16 +38,13 @@ Chunk mergeAndSpliceAndConvertBucketsToChunk(
     ManyAggregatedDataVariants & data, const AggregatingTransformParams & params, const std::vector<Int64> & buckets);
 
 /// Only used for emit changelog
-/// @brief Based on new/updated groups @p retracted_data , only convert the state of changed groups (retracted: last state, aggregated: current state)
-///  \data: current aggregated state of all groups
-///  \retracted_data: only have last state of changed groups (i.e. new/updated/deleted)
+/// @brief only convert the state of changed groups (retracted: last state, aggregated: current state)
+///  \data: current aggregated state of all groups (contains retracted states and updated states)
```
chenziliang:
```
/// Changelog chunk converters are used for changelog emit. They can return a pair of chunks : one 
/// for retraction and one for updates. And those 2 chunks are expected to be passed to downstream
/// consecutively otherwise the down stream aggregation result may not be correct or emit incorrect 
/// intermediate results. To facilitate the downstream processing, we usually mark the `consecutive` 
/// flag bit for these chunks.
/// \return {retract_chunk, update_chunk} pair, retract_chunk if not empty, contains retract data 
///         because of the current updates; update_chunk if not empty, contains the result for the
///         latest update data
```

src/Processors/Transforms/Streaming/AggregatingTransform.cpp
```diff
@@ -498,7 +498,7 @@ void AggregatingTransform::checkpoint(CheckpointContextPtr ckpt_ctx)
         }
 
         /// Serializing no shared data
-        params->aggregator.checkpoint(variants, wb);
+        DB::serialize(variants, wb, params->aggregator);
```
chenziliang: Does variants already points to params->aggregator at this point ?

src/Processors/Transforms/Streaming/AggregatingTransformWithSubstream.cpp
```diff
@@ -270,7 +270,7 @@ void AggregatingTransformWithSubstream::checkpoint(CheckpointContextPtr ckpt_ctx
         for (const auto & [id, substream_ctx] : substream_contexts)
         {
             assert(id == substream_ctx->id);
-            substream_ctx->serialize(wb, getVersion());
+            serialize(*substream_ctx, wb, getVersion());
```
chenziliang: i feel like `substream_ctx->serialize(wb, ...)` is probably better than the current version. Don't know why we change it.

src/Processors/Transforms/Streaming/AggregatingTransformWithSubstream.cpp
```diff
@@ -294,7 +294,7 @@ void SubstreamContext::serialize(WriteBuffer & wb, VersionType version) const
 {
     DB::Streaming::serialize(id, wb);
 
-    aggregating_transform->params->aggregator.checkpoint(variants, wb);
+    DB::serialize(variants, wb, aggregating_transform->params->aggregator);
```
chenziliang: i personally feel like the previous version / style might be better than free function. 

```
variants.serialize(wb, aggregator);
```

src/Processors/Transforms/Streaming/GlobalAggregatingTransform.cpp
```diff
@@ -40,35 +41,58 @@ GlobalAggregatingTransform::GlobalAggregatingTransform(
...
+            {retract_enabled,
              /// Field serializer
-             [this](const std::any & field, WriteBuffer & wb, VersionType) {
-                 const auto & data = std::any_cast<const ManyRetractedDataVariants &>(field);
-                 DB::writeIntBinary(data.size(), wb);
-                 for (const auto & elem : data)
-                     params->aggregator.checkpoint(*elem, wb);
+             [](const std::any & field, WriteBuffer & wb, [[maybe_unused]] VersionType version) {
+                 assert(version >= IMPL_V2_MIN_VERSION);
+                 DB::writeBoolText(std::any_cast<bool>(field), wb);
```
chenziliang: bool binary

src/Common/serde.h
```diff
@@ -27,6 +27,27 @@ void ALWAYS_INLINE deserialize(S & s, RB & rb, VersionType version, Args &&... a
...
+    = requires(S & s, RB & rb, Args &&... args) { s.deserialize(rb, std::forward<Args>(args)...); };
+
+template <typename WB, typename... Args, Serializable<WB, Args...> S>
+void ALWAYS_INLINE serialize(const S & s, WB & wb, Args &&... args)
+{
+    s.serialize(wb, std::forward<Args>(args)...);
+}
+
+template <typename RB, typename... Args, Deserializable<RB, Args...> S>
+void ALWAYS_INLINE deserialize(S & s, RB & rb, Args &&... args)
```
chenziliang: I actually don't understand why we need this since calling s.deserialize / serialize on sites are more readable than yet another forwarding / layer : both at code layer and machine code layer and for debugging purpose

chenziliang: I would highly suggest writing direct / simple code without possible perf regression.

src/Processors/Transforms/Streaming/GlobalAggregatingTransform.cpp
```diff
@@ -103,14 +127,17 @@ std::pair<bool, bool> GlobalAggregatingTransform::executeOrMergeColumns(Chunk &
...
-            chunk.detachColumns(), 0, num_rows, *aggregated_variants, *retracted_variants, key_columns, aggregate_columns, no_more_keys);
+
+        /// Enable retract after first finalization
+        auto retract_enabled = many_data->getField<bool>();
+        if (retract_enabled) [[likely]]
+            return params->aggregator.executeAndRetractOnBlock(
+                chunk.detachColumns(), 0, num_rows, variants, key_columns, aggregate_columns, no_more_keys);
+        else
+            return params->aggregator.executeOnBlock(
+                chunk.detachColumns(), 0, num_rows, variants, key_columns, aggregate_columns, no_more_keys);
```
chenziliang: as mentioned elsewhere, maybe we can remove no_more_keys optimization for streaming aggregation

src/Processors/Transforms/Streaming/GlobalAggregatingTransform.cpp
```diff
@@ -40,35 +41,58 @@ GlobalAggregatingTransform::GlobalAggregatingTransform(
...
-             [this](std::any & field, ReadBuffer & rb, VersionType) {
-                 auto & data = std::any_cast<ManyRetractedDataVariants &>(field);
-                 size_t num;
-                 DB::readIntBinary(num, rb);
-                 data.resize(num);
-                 for (auto & elem : data)
+             [this](std::any & field, ReadBuffer & rb, VersionType version) {
+                 if (version >= IMPL_V2_MIN_VERSION)
+                 {
+                     DB::readBoolText(std::any_cast<bool &>(field), rb);
```
chenziliang: i think we shall use `std::variant` for field instead of std::any. `std::variant` give us a list of possibilities which are essential to serde; however std::any all kinds of possibilities.

src/Processors/Transforms/Streaming/GlobalAggregatingTransform.h
```diff
@@ -28,6 +28,8 @@ class GlobalAggregatingTransform final : public AggregatingTransform
     bool prepareFinalization(Int64 min_watermark) override;
 
     void finalize(const ChunkContextPtr & chunk_ctx) override;
+
+    static constexpr VersionType IMPL_V2_MIN_VERSION = 3;
```
chenziliang: lets just use V2 = 3;

src/Processors/Transforms/Streaming/GlobalAggregatingTransformWithSubstream.h
```diff
@@ -21,6 +21,8 @@ class GlobalAggregatingTransformWithSubstream final : public AggregatingTransfor
 
 private:
     void finalize(const SubstreamContextPtr & substream_ctx, const ChunkContextPtr & chunk_ctx) override;
+
+    static constexpr VersionType IMPL_V2_MIN_VERSION = 3;
```
chenziliang: Same, V2 = 3

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -1274,283 +1366,70 @@ Block Aggregator::convertOneBucketToBlockImpl(
...
-BlocksList
-Aggregator::mergeAndConvertToBlocks(ManyAggregatedDataVariants & data_variants, bool final, ConvertAction action, size_t max_threads) const
-{
-    auto prepared_data_ptr = prepareVariantsToMerge(data_variants);
-    if (prepared_data_ptr->empty())
-        return {};
 
-    if (unlikely(params.overflow_row))
-        throw Exception(ErrorCodes::NOT_IMPLEMENTED, "Overflow row processing is not implemented in streaming aggregation");
+bool Aggregator::checkLimits(size_t result_size, bool & no_more_keys) const
```
chenziliang: For now, probably we can disable limit checking since for streaming processing, it is hard to limit a size since it is long running

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -2588,9 +2136,98 @@ void NO_INLINE Aggregator::mergeSingleLevelDataImpl(ManyAggregatedDataVariants &
...
+}
+
+template <typename Method>
+BlocksList Aggregator::mergeAndConvertTwoLevelToBlocksImpl(
+    ManyAggregatedDataVariants & non_empty_data, bool final, bool clear_states, ThreadPool * thread_pool) const
+{
+    auto & first = *non_empty_data.at(0);
+
+    std::vector<Int64> buckets;
+    if (first.isStaticBucketTwoLevel())
```
chenziliang, line 1936: lets add `{}` for this if

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -103,6 +107,73 @@ Columns materializeKeyColumns(Columns & columns, ColumnRawPtrs & key_columns, co
...
+Arena * getArena(AggregatedDataVariants & variants, AggregateStateType type)
+{
+    if (type == AggregateStateType::OnlyRetracted)
+        return variants.retracted_pool.get();
+    else
+        return variants.aggregates_pool;
+}
+
+template <typename BucketConverter>
+BlocksList concurrentBucketConvert(ThreadPool * thread_pool, const std::vector<Int64> & buckets, Arena * arena, Arenas & pools, BucketConverter && bucket_converter)
```
chenziliang: `convertBucketsInParallel(...)`

chenziliang:
Summary,

1. Always highly recommended move the refactor part to a different PR especially the current one is critical enough, that will make the PR / review process way more efficient
2. I didn't see the importance / necessity of breaking prepareBlockAndFill into 3 pieces so far. If we really do need, a separate PR will be way clearer.
3. We may be able to clean up duplicate semantics like AggregateStateType vs TrackingUpdatesType
4. We may be able to clean up no_more_keys in this PR or in future PR
5. Avoid template as much as possible unless it has real benefits and we will be blowing our binary size / compile time quickly and also harder to maintain / debug. Simple / direct code is usually better.
6. Prefer writeVar for serde since they save disk / network bandwidth and fast as well. Avoid using writeText since they are slow and big.
7. Let's start to comment more as much as possible. Current code base is kinda of slowly out of control regarding what each function does, what they mean. A bit too many branches / conditions to understand. Without documentation / comments, it will be very hard for different people to pick up.

yl-lisen:
> Summary,
> 
> 1. Always highly recommended move the refactor part to a different PR especially the current one is critical enough, that will make the PR / review process way more efficient
> 2. I didn't see the importance / necessity of breaking prepareBlockAndFill into 3 pieces so far. If we really do need, a separate PR will be way clearer.
> 3. We may be able to clean up duplicate semantics like AggregateStateType vs TrackingUpdatesType
> 4. We may be able to clean up no_more_keys in this PR or in future PR
> 5. Avoid template as much as possible unless it has real benefits and we will be blowing our binary size / compile time quickly and also harder to maintain / debug. Simple / direct code is usually better.
> 6. Prefer writeVar for serde since they save disk / network bandwidth and fast as well. Avoid using writeText since they are slow and big.
> 7. Let's start to comment more as much as possible. Current code base is kinda of slowly out of control regarding what each function does, what they mean. A bit too many branches / conditions to understand. Without documentation / comments, it will be very hard for different people to pick up.

Thanks, these are great suggestions.
I'm so sorry, i've separated part of the implementation, but it's still too big, I will separate the refactor parts in next PR

src/Interpreters/Streaming/Aggregator.h
yl-lisen: > For global aggregation, do we need enable recycle ?

it will be disable in `AggregatedDataVariants::init()`

src/Interpreters/Streaming/Aggregator.cpp
yl-lisen: if there is no data, the member variable `aggregator`  in AggregatedDataVariants is nullptr, so we can not use itself aggregator.

yl-lisen: commented here

src/Interpreters/Streaming/Aggregator.cpp
yl-lisen: in fact, we have supported it, for some aggr functions supported compile , for example `min/max/count .etc` , and i think they can GC correctly  ?

yl-lisen: we can not support it first, and decide whether to enable JIT after performing separate perf verification later for streaming query.

src/Interpreters/Streaming/Aggregator.cpp
yl-lisen: Both are good, but if we do it at the tail ,  need extra variable to save the offset and offset the pointer

src/Interpreters/Streaming/AggregateDataEx.h
yl-lisen: nope, only exists for changelog input. but for `no changelog input` and `emit changelog`, we need tracking updates with retract, the `delta_col` is null

src/Interpreters/Streaming/AggregateDataEx.h
yl-lisen: when create state, it should always is updated before finalizing, although this initial value does not affect anything

src/Interpreters/Streaming/AggregateDataEx.h
yl-lisen: For `TrackingUpdates`, there is always delta_col.
For `TrackingUpdatesWithRetract`, delta_col may be nullptr.
But I don’t think it’s worth implementing two codes

src/Interpreters/Streaming/Aggregator.cpp
yl-lisen: For `TrackingUpdatesWithRetract`, the delta_col may be nullptr.

src/Interpreters/Streaming/Aggregator.cpp
yl-lisen: It is a batch process (loop hashmap), so it is added to avoid too many branches during loop processing.
Yes, we can pass it in as a parameter, unless there is a performance problem, and then do another optimization

src/Interpreters/Streaming/Aggregator.cpp
yl-lisen: `prepareBlockAndFill` Is old implementation. for now  community is new implementation and have some optimization on it. 
What I was thinking at the time was to make these changes to the new implementation to avoid doing it again in the future.

src/Interpreters/Streaming/Aggregator.cpp
yl-lisen: This change should be removed, it was missed. 
`clear_states` already handles the branch in the outermost call, I will remove `&& type == AggregateStateType::Normal`

cmake/autogenerated_versions.txt
yl-lisen: it's not necessary, the version will increases automatically after release.
This is only to ensure that CI testing can handle the state based on the new revision.

src/Common/HashTable/TimeBucketHashTable.h
yl-lisen: This method is called after deserialization of TwoLevelHashMap. so it always not empty.

src/Processors/Transforms/Streaming/GlobalAggregatingTransform.h
yl-lisen: This naming is simple but confusing

src/Processors/Transforms/Streaming/GlobalAggregatingTransformWithSubstream.h
yl-lisen: This naming is simple but confusing

src/Processors/Transforms/Streaming/GlobalAggregatingTransform.cpp
yl-lisen: I don't think "any -> variant" makes sense, just goes from unrestricted to semi-restricted. This is no simplification either for implementation or understanding, and different implementations still need to know what they are and what they do. For `std::variant`, if there is new content, additional type declarations need to be added.

src/Common/serde.h
yl-lisen: Yes, it is actually useless. It is only used to unify the code style.

src/Interpreters/Streaming/Aggregator.cpp
yl-lisen: applied it

src/Interpreters/Streaming/Aggregator.cpp
yl-lisen: So far I haven't found any use of arenas during the conversion process, meaning the new arena doesn't require GC and a temporary replacement can be used

yl-lisen: will use temp arenas

src/Processors/Transforms/Streaming/AggregatingHelper.cpp
yl-lisen: TrackingUpdatesType indicates that the aggregate states requires tracking updates. But here we need to specify what content we want to convert (`All` or `OnlyUpdates`, or `OnlyRetract`)

I will refine the `AggregateStateType ` to `ConvertType` :
```c++
enum class ConvertType : uint8_t
{
    Normal = 0,
    OnlyUpdates = 1,
    OnlyRetract = 2,
};
```

src/Interpreters/Streaming/Aggregator.h
yl-lisen: TrackingUpdatesType indicates that the aggregate states requires tracking updates. But here we need to specify what content we want to convert (All or OnlyUpdates, or OnlyRetract)

I will refine the AggregateStateType  to ConvertType :

enum class ConvertType : uint8_t
{
    Normal = 0,
    OnlyUpdates = 1,
    OnlyRetract = 2,
};

src/Processors/Transforms/Streaming/AggregatingTransform.cpp
yl-lisen: if there is no data, variants points to nullptr aggregator.

src/Interpreters/Streaming/Aggregator.cpp
chenziliang: I understand the nullptr case, but when we do serialize, data variants already points to a valid aggregator ?

src/Common/HashTable/TimeBucketHashTable.h
chenziliang: We are doing the deserialization here, if before the deserialization, it already has data, there is something wrong since we expect during deserialization, everything is empty ?

src/Processors/Transforms/Streaming/AggregatingHelper.cpp
chenziliang: Seems make sense, can we rename `OnlyUpdates` -> `Updates`; `OnlyRetract` -> `Retract` ?

src/Processors/Transforms/Streaming/GlobalAggregatingTransform.cpp
chenziliang: Yes, this is the purpose to explicitly list what it is possible contains instead of a `void *`. We like to make things explicit especially for serde. When it can contain a different type like `std::string`, making it explicit has benefits we are not breaking something

src/Processors/Transforms/Streaming/GlobalAggregatingTransform.h
chenziliang: from versioning system point of view , we shall just use v2, v3, v4 etc instead of a min version, max version which already has overlapped meaning. Version number itself is already self-documenting the things, we don't need min / max.

chenziliang: Client usually does do `if (v > v2 && v < v4) ... `

src/Processors/Transforms/Streaming/AggregatingTransform.cpp
```diff
@@ -189,22 +189,11 @@ void AggregatingTransform::consume(Chunk chunk)
...
-        auto success = params->aggregator.mergeOnBlock(block, variants, no_more_keys);
-        return {!success, false};
-    }
-    else
-    {
-        /// Blocking finalization during execution on current variant
-        std::lock_guard lock(variants_mutex);
-        return params->aggregator.executeOnBlock(std::move(columns), 0, num_rows, variants, key_columns, aggregate_columns, no_more_keys);
-    }
+    assert(!params->only_merge && !no_more_keys);
```
chenziliang, line 192: Do we need no_more_keys here ?

src/Processors/Transforms/Streaming/AggregatingTransformWithSubstream.cpp
```diff
@@ -226,10 +226,10 @@ std::pair<bool, bool> AggregatingTransformWithSubstream::executeOrMergeColumns(C
     /// according to partition keys
     auto num_rows = chunk.getNumRows();
 
-    assert(!params->only_merge);
+    assert(!params->only_merge && !no_more_keys);
```
chenziliang, line 229: no_more_keys ? We can get rid of it in separate PR if necessary since this one is big enough

src/Interpreters/Streaming/UpdatesTrackingData.h
```diff
@@ -0,0 +1,105 @@
...
+    /// Used to track if the target to be tracked has zero sum changes
+    UInt64 updates = 0;
+
+    /// Used to track if the target group tracked has updates since last finalization 
+    bool updated_since_last_finalization = true;
+};
+
+enum class TrackingUpdatesType : uint8_t
+{
+    None = 0,
```
chenziliang: Do we need None ? I feel like it shall be a `std::optional<TrackingUpdatesType>` ? Or do we need an enum since we remove the retract ?

src/Common/serde.h
chenziliang: Let's clean them up to avoid future misuse

src/Interpreters/Streaming/Aggregator.h
chenziliang: Let's put a comment there 

```
/// Enable GC for arena by default. For cases like global aggregation, we will disable it further in \init
```

src/Interpreters/Streaming/Aggregator.h
```diff
@@ -74,15 +76,10 @@ namespace Streaming
...
 {
-    Unkonwn = 0,
-    DistributedMerge,
-    WriteToTmpFS,
-    Checkpoint,
-    StreamingEmit,
-    InternalMerge,
-    RetractedEmit
+    Normal = 0,
+    OnlyUpdates = 1,
```
chenziliang: So we don't need `Retract` anymore ?

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -103,6 +104,66 @@ Columns materializeKeyColumns(Columns & columns, ColumnRawPtrs & key_columns, co
 
     return materialized_columns;
 }
+
+template <typename BucketConverter>
+BlocksList convertBucketsInParallel(ThreadPool * thread_pool, const std::vector<Int64> & buckets, BucketConverter && bucket_converter)
+{
+    std::atomic<UInt32> next_bucket_idx_to_merge = 0;
+    auto converter = [&](Arena * pool, const std::atomic_flag * cancelled) {
+        BlocksList blocks;
```
chenziliang, line 113: We can create arena in the lambda, after it is done, it got GCed ?

```
Arena arena;
BlocksList ...
....
```

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -789,8 +856,9 @@ template <bool no_more_keys, bool use_compiled_functions, typename Method>
                     inst->state_offset,
                     [&](AggregateDataPtr & aggregate_data)
                     {
-                        aggregate_data = aggregates_pool->alignedAlloc(total_size_of_aggregate_states, align_aggregate_states);
-                        createAggregateStates(aggregate_data);
+                        auto data = aggregates_pool->alignedAlloc(total_size_of_aggregate_states, align_aggregate_states);
+                        createAggregateStates(data, /*prefix_with_updates_tracking_state*/ false);
```
chenziliang: `/*prefix_with_updates_tracking_state=*/`

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -3321,33 +2322,40 @@ std::vector<Int64> Aggregator::bucketsBefore(const AggregatedDataVariants & resu
...
 {
+    auto version = getVersion();
     /// Serialization layout
     /// [version] + [states layout]
-    VersionType version = getVersion();
     writeIntBinary(version, wb);
 
     if (version <= 1)
-        return doCheckpointLegacy(data_variants, wb);
+        return const_cast<Aggregator *>(this)->doCheckpointLegacy(data_variants, wb);
```
chenziliang, line 2328: if (version <= 1)
else if (version <= 2)
else

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -4263,30 +3214,117 @@ void Aggregator::deserializeAggregateStates(AggregateDataPtr & place, ReadBuffer
...
-void Aggregator::clearDataVariants(AggregatedDataVariants & data_variants) const
+void Aggregator::doCheckpointV3(const AggregatedDataVariants & data_variants, WriteBuffer & wb) const
 {
-    /// Clear states
-    destroyAllAggregateStates(data_variants);
+    /// Serialization layout, there are 2 cases:
+    /// 1) Without key: [uint8][uint16][aggr-func-state-without-key]
+    /// 2) Otherwise: [uint8][uint16][aggr-func-state-for-overflow-row][is_two_level][aggr-func-state-in-hash-map]
+    bool inited = !data_variants.empty();
+    writeBoolText(inited, wb);
```
chenziliang: write binary

src/Interpreters/Streaming/Aggregator.cpp
chenziliang: This make sense. And let's cherry pick more regarding the aggregator improvements.

src/Interpreters/Streaming/UpdatesTrackingData.h
```diff
@@ -0,0 +1,105 @@
+#pragma once
```
chenziliang: Let's rename the header file as `TrackingUpdatesData` to match the struct name.

src/Processors/Transforms/Streaming/GlobalAggregatingTransform.cpp
chenziliang: Let's defer std::any vs std::variant in future PR

src/Interpreters/Streaming/Aggregator.cpp
yl-lisen: No, it's nullptr unless we explicitly assign it before

src/Common/HashTable/TimeBucketHashTable.h
yl-lisen:
```c++
template <typename Map, typename MappedDeserializer>
void deserializeTwoLevelHashMap(Map & map, MappedDeserializer && mapped_deserializer, Arena & pool, ReadBuffer & rb)
{
    deserializeHashMap<Map, MappedDeserializer>(map, std::move(mapped_deserializer), pool, rb);
    map.readUpdatedBuckets(rb); /// recover buckets updated status
}
```

For above impl,before `readUpdatedBuckets`, the `deserializeHashMap ` will reinsert all key, which will cause the `bucket_updated_flags ` all are true.

src/Interpreters/Streaming/UpdatesTrackingData.h
yl-lisen: This makes it easier to seralize/deserialize. Otherwise, we need to write 1）whether tracking is required, 2）and if so, the type of tracking.

src/Interpreters/Streaming/Aggregator.h
yl-lisen: will add it in a separate PR later

src/Interpreters/Streaming/Aggregator.cpp
yl-lisen: can do it in separate PR

qijun-niu-timeplus (approved):

PR #476, opened 2024-01-09
Disable backfill sorting by default

### LINKED ISSUE #475: Disable backfill sorting by default
Usually, we don't care whether the backfilled data is in order, and this operation is very expensive for large amounts of historical data.

So can disable the backfill sorting by default. 
And if need sort the backfilled data, we can use a setting `force_backfill_in_order=true` 

There are three settings about backfill data:
- **enable_backfill_from_historical_store** (default is `true`)
   Enable backfill data from historical data store

- **emit_aggregated_during_backfill** (default is `false`)
    Enable emit aggr result during backfill historical data

- **force_backfill_in_order** (default is `false`)
    Requires backfill data in order

yl-lisen:
This close #475 

There are three settings about backfill data:
- **enable_backfill_from_historical_store** (default is `true`)
   Enable backfill data from historical data store

- **emit_aggregated_during_backfill** (default is `false`)
    Enable emit aggr result during backfill historical data

- **force_backfill_in_order** (default is `false`)
    Requires backfill data in order

### DIFF
--- a/src/Storages/SelectQueryInfo.h
+++ b/src/Storages/SelectQueryInfo.h
@@ -184,7 +184,7 @@ struct SelectQueryInfo
     SeekToInfoPtr seek_to_info_of_right_stream; /// Rewind info for right streaming store in streaming query
 
     /// if is true, means need sort backfill input by event time (ascending)
-    bool require_in_order_backfill = true;
+    bool require_in_order_backfill = false;
 
     Streaming::WindowParamsPtr streaming_window_params;
 
--- a/src/Core/Settings.h
+++ b/src/Core/Settings.h
@@ -790,7 +790,8 @@ static constexpr UInt64 operator""_GiB(unsigned long long value)
     M(UInt64, keep_windows, 0, "How many streaming windows to keep from recycling", 0) \
     M(String, seek_to, "", "Seeking to an offset of the streaming/historical store to seek", 0) \
     M(Bool, enable_backfill_from_historical_store, true, "Enable backfill data from historical data store", 0) \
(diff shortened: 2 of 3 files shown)


sunset3000 (approved):

src/Storages/SelectQueryInfo.h
```diff
@@ -184,7 +184,7 @@ struct SelectQueryInfo
     SeekToInfoPtr seek_to_info_of_right_stream; /// Rewind info for right streaming store in streaming query
 
     /// if is true, means need sort backfill input by event time (ascending)
-    bool require_in_order_backfill = true;
+    bool require_in_order_backfill = false;
```
chenziliang, line 187: Do we need disable watermark during backfill when in order backfill is false ? Do we emit tumble windows during backfill or global aggregation ?

yl-lisen: For 1) , controlled by `emit_aggregated_during_backfill ` , when it is false, we don't generate watermark  for historical data
(But there are still some limitations need to be resolved in future, when exists subquery with join/aggr, we can not get the historical flag from subquery input)

For 2) by default `emit_aggregated_during_backfill  = false` , don't emit the aggr result

yl-lisen: you can see this logic here.
https://github.com/timeplus-io/proton/blob/037555c1bf7f1568ee1e3c3e974a809d5e407ade/src/Interpreters/InterpreterSelectQuery.cpp#L3495-L3509C2

PR #480, opened 2024-01-11
enhance UDF name validation

### LINKED ISSUE #479: Unknown function exception for a UDA
**Describe what's wrong**

1. Run a query with native 'group_uniq_array' aggregate function, it works well
```SQL
with cte as (select array_join(['a', 'b', 'a']) as a)
select group_uniq_array(a) from cte;
```
2. However, if replace 'group_uniq_array' with a UDA 'group_distinct'
UDA definition
```SQL
CREATE OR REPLACE AGGREGATE FUNCTION group_distinct(value string)
RETURNS array(string)
LANGUAGE JAVASCRIPT AS $$
{
    initialize: function () {
        this.set = new Set();
    },

    process: function (values, deltas) {
        let is_changelog = deltas !== undefined;
        for (let i = 0; i < values.length; i++) {
            if (is_changelog) {
                if (deltas[i] === -1 && this.set.has(values[i])) {
                    this.set.delete(values[i]);
                } else if (deltas[i] === 1 && values[i] !== "") {
                    if (!this.set.has(values[i])) {
                        this.set.add(values[i]);
                    }
                }
            } else {
                if (!this.set.has(values[i]) && values[i] !== "") {
                    this.set.add(values[i]);
                }
            }
        }
    },

    finalize: function () {
        return [...this.set];
    },

    serialize: function () {
        return JSON.stringify([...this.set]);
    },

    deserialize: function (state_str) {
        this.set = new Set(JSON.parse(state_str));
    },

    merge: function (state) {
        let other_set = JSON.parse(state);
        other_set.forEach((v) => this.set.add(v));
    }
}
$$;
```
```SQL
with cte as (select array_join(['a', 'b', 'a']) as a)
select group_distinct(a) from cte;
```
it throws the below exception
```TEXT
Code: 46. DB::Exception: Received from 127.0.0.1:8463. DB::Exception: Unknown function group_distinct. There is an aggregate function with the same name, but ordinary function is expected here. Maybe you meant: ['array_distinct','geo_distance']. (UNKNOWN_FUNCTION)
```
**How to reproduce**

**Additional context**

lizhou1111:
When you create a UDA funtion like below:
```sql
CREATE OR REPLACE AGGREGATE FUNCTION group_distinct(value string)
RETURNS array(string)
LANGUAGE JAVASCRIPT AS $$
....
$$;
```
Proton won't recognize is as aggregation function, the reason is in this [code ](https://github.com/timeplus-io/proton/blob/develop/src/AggregateFunctions/AggregateFunctionFactory.cpp#L290C13-L290C13).Because the `_distinct` is key word suffix.
The other key word suffixes are below:
- `_distinct_streaming`
- `_distinct_retract`
- `_sample_state`
- `_or_default`
- `_distinct`
- `_resample`
- `_for_each`
- `_or_null`
- `_merge`
- `_state`
- `_array`
- `_null`
- `_map`
- `_if`
this closes #479

### DIFF
--- a/src/Functions/UserDefined/UDFHelper.cpp
+++ b/src/Functions/UserDefined/UDFHelper.cpp
@@ -54,7 +54,7 @@ void validateUDFName(const String & func_name)
         throw Exception(ErrorCodes::UDF_INVALID_NAME, "UDF name's first char shall be an alphabetic or underscore");
 
     if (AggregateFunctionCombinatorPtr combinator = AggregateFunctionCombinatorFactory::instance().tryFindSuffix(func_name))
-        throw Exception(ErrorCodes::UDF_INVALID_NAME, "UDF name can not end up with " + combinator->getName() + ", because it is key word suffix");
+        throw Exception(fmt::format("UDF name can not end up with {}, because it is key word suffix", combinator->getName()), ErrorCodes::UDF_INVALID_NAME);
 }
 
 namespace
--- a/src/Functions/UserDefined/UDFHelper.cpp
+++ b/src/Functions/UserDefined/UDFHelper.cpp
@@ -1,5 +1,6 @@
 #include <Functions/UserDefined/UDFHelper.h>
 
+#include <AggregateFunctions/AggregateFunctionCombinatorFactory.h>
 #include <DataTypes/DataTypeFactory.h>
 #include <Functions/UserDefined/IUserDefinedSQLObjectsLoader.h>
 #include <Functions/UserDefined/UserDefinedExecutableFunction.h>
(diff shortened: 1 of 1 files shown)


src/Functions/UserDefined/UDFHelper.cpp
```diff
@@ -51,6 +52,9 @@ void validateUDFName(const String & func_name)
 
     if (!std::isalpha(func_name[0]) && func_name[0] != '_')
         throw Exception(ErrorCodes::UDF_INVALID_NAME, "UDF name's first char shall be an alphabetic or underscore");
+
+    if (AggregateFunctionCombinatorPtr combinator = AggregateFunctionCombinatorFactory::instance().tryFindSuffix(func_name))
+        throw Exception(ErrorCodes::UDF_INVALID_NAME, "UDF name can not end up with " + combinator->getName() + ", because it is key word suffix");
```
yl-lisen: use format `Exception(ErrorCodes::UDF_INVALID_NAME, "UDF name can not end up with {}, because it is key word suffix", combinator->getName());`

yl-lisen (approved):

Issue #421, opened 2023-12-18
Improve checkpoint versioning stability


yl-lisen:
**Use case**

In proton, query state management is works via checkpoints. currently we use a strict checkpoint/recover implementation, which is easily broken when there are any modifications about query pipeline,  processors or its header changes.

So we want to improve the stability of checkpoints, and reduce broken scenarios as much as possible

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

jovezhong:
(Jove Github Bot) added it to the current sprint.

PR #495, opened 2024-01-16
Enable proton revision for checkpoint

yl-lisen:
This is phase 1 for #421 

- Enable proton revision for checkpoint
- revise partial serialize/deserialize interface to support versioning

### DIFF
--- a/src/Common/serde.h
+++ b/src/Common/serde.h
@@ -1,0 +1,63 @@
+#pragma once
+
+#include <IO/VarInt.h>
+#include <base/types.h>
+#include <Common/VersionRevision.h>
+
+namespace DB
+{
+/// REQUIRES: The object must support versioned serialization/deserialization
+template <typename S, typename WB>
+concept VersionedSerializable = requires(const S & s, WB && wb, VersionType version) { s.serialize(wb, version); };
+
+template <typename S, typename RB>
+concept VersionedDeserializable = requires(S && s, RB && rb, VersionType version) { s.deserialize(rb, version); };
+
+template <typename T>
+concept HasGetVersionFromRevision = requires(T && a) {
+    {
+        a.getVersionFromRevision(ProtonRevision::getVersionRevision())
+    } -> std::convertible_to<VersionType>;
+};
+
+template <typename WB, VersionedSerializable<WB> S>
+void ALWAYS_INLINE serialize(const S & s, WB && wb, std::optional<VersionType> version = std::nullopt)
+{
+    if (version.has_value())
+    {
+        s.serialize(wb, version.value());
+    }
+    else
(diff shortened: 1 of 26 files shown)


src/Common/serde.h
```diff
@@ -0,0 +1,63 @@
...
+
+template <typename T>
+concept HasGetVersionFromRevision = requires(T && a) {
+    {
+        a.getVersionFromRevision(ProtonRevision::getVersionRevision())
+    } -> std::convertible_to<VersionType>;
+};
+
+template <typename WB, VersionedSerializable<WB> S>
+void ALWAYS_INLINE serialize(const S & s, WB && wb, std::optional<VersionType> version = std::nullopt)
```
chenziliang: This appears too complex and has too many branches. I like to recommend
1) make code simple to understand 
2) make code branch less as much as possible 

Simple code is usually faster to execute

src/Common/serde.h
```diff
@@ -0,0 +1,49 @@
...
+{
+    /// By default use `Revision` for version management, unless it has its own versions
+    if constexpr (HasGetVersionFromRevision<S>)
+        s.serialize(wb, static_cast<VersionType>(s.getVersionFromRevision(ProtonRevision::getVersionRevision())));
+    else
+        s.serialize(wb, static_cast<VersionType>(ProtonRevision::getVersionRevision()));
+}
+
+template <typename RB, VersionedDeserializable<RB> S>
+void ALWAYS_INLINE deserialize(S & s, RB && rb)
```
chenziliang: It seems we have a `&&` and we reading data from  it.

I still think we probably have too much complexity here and hence lots of slowness and confusion

src/Common/serde.h
```diff
@@ -0,0 +1,49 @@
...
+
+template <typename T>
+concept HasGetVersionFromRevision = requires(T && a) {
+    {
+        a.getVersionFromRevision(ProtonRevision::getVersionRevision())
+    } -> std::convertible_to<VersionType>;
+};
+
+template <typename WB, VersionedSerializable<WB> S>
+void ALWAYS_INLINE serialize(const S & s, WB && wb)
```
chenziliang: it seems weird we have a && and we are writing data to it.

src/Common/serde.h
yl-lisen: 1) change `&&` to `&`, it's make sense

2) I think it's no slowness, and no confusion, which just used to unify the serialization/deserialization interface to simplify the interface call of version management objects

yl-lisen: It's a little complicated, I'll readjust it

sunset3000 (approved):

PR #499, opened 2024-01-17
fix incorrect subscribe mode check for internal subquery

### LINKED ISSUE #498: CREATE MATERIALIZED VIEW Failed for streaming query
**Describe what's wrong**

```sql
create stream dim_t(id int, type int);
create stream vk(id int, value float) primary key id settings mode='versioned_kv';
create materialized view mv as with list as (select id from table(dim_t)) select id, value from vk where id in list;
```
**But got exception:**
Received exception from server (version 1.3.31):
Code: 48. DB::Exception: Received from localhost:8463. DB::Exception: SUBSCRIBE can only work with streaming query. (NOT_IMPLEMENTED)

**How to reproduce**

**Additional context**

yl-lisen:
Fix #498

### DIFF
--- a/src/Processors/QueryPlan/QueryPlan.cpp
+++ b/src/Processors/QueryPlan/QueryPlan.cpp
@@ -198,8 +198,6 @@ QueryPipelineBuilderPtr QueryPlan::buildQueryPipeline(
     last_pipeline->addResources(std::move(resources));
 
     /// proton : starts
-    last_pipeline->setExecuteMode(queryExecuteMode(isStreaming(), query_context->getSettingsRef()));
-
     if (isStreaming() != last_pipeline->isStreaming())
         throw Exception(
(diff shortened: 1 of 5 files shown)


src/Processors/QueryPlan/QueryPlan.cpp
```diff
@@ -198,8 +198,6 @@ QueryPipelineBuilderPtr QueryPlan::buildQueryPipeline(
     last_pipeline->addResources(std::move(resources));
 
     /// proton : starts
-    last_pipeline->setExecuteMode(queryExecuteMode(isStreaming(), query_context->getSettingsRef()));
-
```
yl-lisen, line 202: already exists `setExecuteMode` in src/Interpreters/InterpreterSelectWithUnionQuery.cpp:403

sunset3000 (approved):

PR #517, opened 2024-01-23
support more emit modes in window

yl-lisen:
PR checklist:
- Did you run ClangFormat ?

- Support emit intermediate windows (if specified `emit periodic` or `emit on update`)

- Support only emit update groups (if specified `emit on update` or `emit periodic on update`

- Tweak emit clause syntax:
```sql
EMIT [STREAM|CHANGELOG]
    [AFTER WATERMARK WITH DELAY <interval>]
    [PERIODIC <interval>]
    [ON UPDATE]
    [[ AND ]TIMEOUT <interval>]
    [[ AND ]LAST <last-x> [ON PROCTIME]]
```
**(MODIFED)**
`[AFTER WATERMARK WITH DELAY <interval>]` :  Allow time skew in a certain range `[<max_event_ts - delay_interval>, <max_event_ts>]

**(NEW)** `[ON UPDATE]` : emit per update  and only emit update per `group by key`
**(NEW)** `[PERIODIC <interval> ON UPDATE]` : periodic emit and only emit update  per `group by key`


src/Parsers/Streaming/ParserEmitQuery.cpp
```diff
@@ -4,27 +4,32 @@
...
-bool ParserEmitQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & expected, [[ maybe_unused ]] bool hint)
+bool ParserEmitQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & expected, [[maybe_unused]] bool hint)
 {
     /// EMIT [STREAM|CHANGELOG]
-    ///         - [PERIODIC INTERVAL '3' SECONDS]
-    ///         - [AFTER WATERMARK]
-    ///         - [DELAY INTERVAL '3' SECONDS]
-    ///         - [LAST <last-x>]
-    /// For each sub-option can be combined with 'AND', we shall select a matching mod based on the combination of parsed content finally.
+    ///     [AFTER WATERMARK [WITH DELAY <interval>]|[WITHOUT DELAY]]
```
chenziliang: Let's remove `WITHOUT DELAY` since it is just more verbose and more typing.

tests/stream/test_stream_smoke/0034_emit_mode.yaml
```diff
@@ -0,0 +1,318 @@
...
+    expected_results:
+      - query_id: emit-mode-1-1
+        expected_results:
+          - [1, 1, 1]
+          - [1, 2, 2]
+          - [1, 3, 3]
+      - query_id: emit-mode-1-2
+        expected_results:
+          - [1, 1]
+          - [1, 1]
```
chenziliang: key 1 are not changed, we still emit key 1 multiple times ?

chenziliang: Same for ke y2

tests/stream/test_stream_smoke/0034_emit_mode.yaml
```diff
@@ -0,0 +1,318 @@
...
+            query_type: table
+            kill: emit-mode-1-1,emit-mode-1-2,emit-mode-1-3,emit-mode-1-4,emit-mode-1-5
+            kill_wait: 2
+            wait: 1
+            query: insert into test_35_stream(i, _tp_time) values(3, '2024-01-24 10:00:10');
+
+    expected_results:
+      - query_id: emit-mode-1-1
+        expected_results:
+          - [1, 1, 1]
```
chenziliang, line 318: What's the difference between `emit periodic and emit periodic on update` ? From the testing results, they look the same ?

src/Core/Streaming/Watermark.h
```diff
@@ -9,5 +9,48 @@ namespace Streaming
...
+    OutOfOrdernessInBatch,
+
+    /// (Built-in): Watermark is generated based on new window of a batch events:
+    /// - Allow time skew in same window and one batch
+    OutOfOrdernessInWindowAndBatch,
+
+    /// TODO: more strategies ...
+};
+
+enum class WatermarkEmitMode
```
chenziliang: Why we need this since the WatermarkStrategy already says this purpose ?

src/Core/Streaming/Watermark.h
```diff
@@ -9,5 +9,48 @@ namespace Streaming
...
+    /// No watermark
+    None,
+
+    /// Watermark is generated based on each event, this is used for `EMIT AFTER WATERMARK WITHOUT DELAY`
+    /// - Not allow time skew
+    Ascending,
+
+    /// Watermark is generated based on last event with a delay interval, this is used for `EMIT AFTER WATERMARK WITH DELAY <delay_interval>`:
+    /// - Allow time skew in a certain range `[<max_event_ts - delay_interval>, <max_event_ts>]`)
+    BoundedOutOfOrderness,
```
chenziliang, line 27: We may need revise the names since we don't understand what they means.

src/Core/Streaming/Watermark.h
chenziliang: This shall be just `EmitMode` instead of WatermarkEmitMode or better `EmitStrategy` ?

chenziliang: `WatermarkStrategy` tells us how / when to stamp watermark.
`EmitStrategy` tells us how to do finalization (what to emit) when watermark has been observed.

chenziliang: So for EmitStrategy, we only have 4 combinations :
1. `emit all states` in a tumble / hop window or global window when a watermark is observed.
2. `emit only changed states per group key` in a timed window or global window when a watermark is observed. This emit strategy works for session window and substream by default from day one 
3. `emit changelog`, combined with 1) and 2), we have 2 here. But emit changelog for all states` probably doesn't make sense at all, `emit changelog for updates` make lots of sense though

chenziliang: So the valid emit strategy for now could be

```
1. emit all states
2. emit only keyed and changed states
3. emit changelog for keyed and changed states
```

src/Interpreters/Streaming/WindowCommon.h
```diff
@@ -273,6 +273,9 @@ struct SessionWindowParams : WindowParams
     bool start_with_inclusion;
     bool end_with_inclusion;
 
+    /// TODO: So far, always assign session window in aggr
+    bool assign_window_pushdown = true;
```
chenziliang:
```
/// So far, only for session window, we evaluate the watermark and window for the events in Aggregate Transform
/// For other windows, we assigned the watermark in Watermark stamper step.
```

chenziliang: Previously we are kind mixing `window assignment and watermark assignment` for tumble / hopping / global windows since this is probably more performant.

Right now, we are separating these 2 steps, which are good for maintainability moving forward.

chenziliang: For session window, fundamentally, it doesn't have a fixed watermark because it highly depends on inputs and the evaluation of start / end expression.

So, it seems window assignment for now all happens at Aggregate transform step ? The Watermark evaluation happens in WatermarkStamper step and for session window, we don't have watermark

chenziliang: `pushdown_window_assignment`

src/Processors/QueryPlan/Streaming/AggregatingStep.h
```diff
@@ -22,7 +24,8 @@ class AggregatingStep : public ITransformingStep
         size_t merge_threads_,
         size_t temporary_data_merge_threads_,
         bool emit_version_,
-        bool emit_changelog_);
+        bool emit_changelog_,
```
chenziliang, line 27: `bool emit_changelog_` is redundant moving forward since emit strategy shall cover that

src/Processors/Transforms/Streaming/WindowAggregatingTransformWithSubstream.h
```diff
@@ -21,12 +21,16 @@ class WindowAggregatingTransformWithSubstream : public AggregatingTransformWithS
...
-    virtual WindowsWithBuckets getFinalizedWindowsWithBuckets(Int64 watermark, const SubstreamContextPtr & substream_ctx) const = 0;
+    virtual WindowsWithBuckets getWindowsWithBuckets(const SubstreamContextPtr & substream_ctx) const = 0;
+    virtual Window getLastFinalizedWindow(const SubstreamContextPtr & substream_ctx) const = 0;
     virtual void removeBucketsImpl(Int64 watermark, const SubstreamContextPtr & substream_ctx) = 0;
     virtual bool needReassignWindow() const = 0;
 
     std::optional<size_t> window_start_col_pos;
     std::optional<size_t> window_end_col_pos;
+
+    bool only_emit_finalized_windows = true;
```
chenziliang, line 32: these bools are basically a derive from the emit strategy enums ? It appears making the whole code path complex

src/Processors/Transforms/Streaming/AggregatingHelper.h
```diff
@@ -31,12 +31,26 @@ Chunk mergeAndConvertToChunk(ManyAggregatedDataVariants & data, const Aggregatin
...
 /// splice aggregatd state of multiple buckets and convert them to chunk
-Chunk spliceAndConvertBucketsToChunk(
+Chunk spliceAndConvertToChunk(AggregatedDataVariants & data, const AggregatingTransformParams & params, const std::vector<Int64> & buckets);
+/// Merge many aggregated state of multiple threads, then splice aggregatd state of multiple buckets and convert them to chunk
+Chunk mergeAndSpliceAndConvertToChunk(
+    ManyAggregatedDataVariants & data, const AggregatingTransformParams & params, const std::vector<Int64> & buckets);
+
+/* For emit on update */
+/// Convert aggregated state of update groups tracked to chunk
+Chunk convertUpdateToChunk(AggregatedDataVariants & data, const AggregatingTransformParams & params);
```
chenziliang: `convertUpdatesToChunk`

src/Processors/Transforms/Streaming/AggregatingHelper.h
```diff
@@ -31,12 +31,26 @@ Chunk mergeAndConvertToChunk(ManyAggregatedDataVariants & data, const Aggregatin
 
 /// Only used for two level
 /// splice aggregatd state of multiple buckets and convert them to chunk
-Chunk spliceAndConvertBucketsToChunk(
+Chunk spliceAndConvertToChunk(AggregatedDataVariants & data, const AggregatingTransformParams & params, const std::vector<Int64> & buckets);
+/// Merge many aggregated state of multiple threads, then splice aggregatd state of multiple buckets and convert them to chunk
+Chunk mergeAndSpliceAndConvertToChunk(
+    ManyAggregatedDataVariants & data, const AggregatingTransformParams & params, const std::vector<Int64> & buckets);
+
+/* For emit on update */
```
chenziliang: `/// for EMIT ON UPDATE`

src/Processors/Transforms/Streaming/AggregatingHelper.h
```diff
@@ -31,12 +31,26 @@ Chunk mergeAndConvertToChunk(ManyAggregatedDataVariants & data, const Aggregatin
...
+/// Only used for two level
+/// splice aggregatd state of update groups tracked of multiple buckets and convert them to chunk
+Chunk spliceAndConvertUpdatesToChunk(
     AggregatedDataVariants & data, const AggregatingTransformParams & params, const std::vector<Int64> & buckets);
 /// Merge many aggregated state of multiple threads, then splice aggregatd state of multiple buckets and convert them to chunk
-Chunk mergeAndSpliceAndConvertBucketsToChunk(
+Chunk mergeAndSpliceAndConvertUpdatesToChunk(
     ManyAggregatedDataVariants & data, const AggregatingTransformParams & params, const std::vector<Int64> & buckets);
 
+/* For emit changelog */
```
chenziliang: `/// For EMIT CHANGELOG`

src/Processors/Transforms/Streaming/AggregatingHelper.cpp
```diff
@@ -69,21 +69,58 @@ Chunk convertToChunk(AggregatedDataVariants & data, const AggregatingTransformPa
...
+
+    return {};
+}
+
+Chunk spliceAndConvertUpdatesToChunk(AggregatedDataVariants & data, const AggregatingTransformParams & params, const std::vector<Int64> & buckets)
+{
+    return DB::convertToChunk(params.aggregator.spliceAndConvertUpdatesToBlock(data, buckets));
+}
+
+Chunk mergeAndSpliceAndConvertUpdatesToChunk(
```
chenziliang, line 117: We probably don't need `Splice` in the name

src/Processors/Transforms/Streaming/TumbleHelper.cpp
```diff
@@ -0,0 +1,111 @@
+#include <Processors/Transforms/Streaming/TumbleHelper.h>
```
chenziliang: Let's rename all Helper in the namespace to `Window` like `TumbleWindow`, `HopWindow` and `SessionWindow`

src/Processors/Transforms/Streaming/WatermarkStamper.cpp
```diff
@@ -32,45 +36,73 @@ namespace
...
-        params.mode = WatermarkStamperParams::EmitMode::NONE;
+}
 
-    if (emit->timeout_interval)
-        params.timeout_interval = extractInterval(emit->timeout_interval->as<ASTFunction>());
+void initWatermarkStrategyAndEmitModeForWindowAggr(WatermarkStrategy & strategy, WatermarkEmitMode & mode, WindowParams & params)
+{
+    switch (params.type)
+    {
+        case WindowType::TUMBLE:
```
chenziliang: Let's rename `WindowType::TUMBLE` to `WindowType::Tumble`. Same for others

src/Processors/Transforms/Streaming/WatermarkStamper.cpp
```diff
@@ -32,45 +36,73 @@ namespace
...
+            return TumbleHelper::validateWatermarkStrategyAndEmitMode(strategy, mode, assert_cast<TumbleWindowParams &>(params));
```
chenziliang: What's the benefit to use free functions compared with inheritance as we did switch and all type casts etc here

src/Processors/Transforms/Streaming/WatermarkStamper.cpp
```diff
@@ -361,43 +369,129 @@ void WatermarkStamper::processWatermark(Chunk & chunk)
     }
 
     chunk.setColumns(columns, columns[0]->size());
+}
+
+template <WatermarkStrategy strategy>
+bool WatermarkStamper::applyWatermarkImpl(Int64 event_ts)
```
chenziliang: The same code smells happens here as before as this function is very different for different Watermark strategy and we compile the logic in one function by using templates.

src/Processors/Transforms/Streaming/WatermarkStamper.h
```diff
@@ -66,63 +60,88 @@ SERDE class WatermarkStamper
...
-private:
-    template <typename TimeColumnType, bool apply_watermark_per_row>
+    /// \brief Process events and apply watermark
     void processWatermark(Chunk & chunk);
 
-    void processPeriodic(Chunk & chunk);
-
-    void processTimeout(Chunk & chunk);
+    /// \brief Emit a watermark according to the EmitMode
+    /// \p old_watermark_ts - saved watermark before current processing
```
chenziliang: Let's consistently use either `@brief, \brief, \param etc`, `\param` has builtin support by CLION

src/Processors/Transforms/Streaming/WatermarkStamper.h
```diff
@@ -66,63 +60,88 @@ SERDE class WatermarkStamper
...
+    private:
+        Int64 next_timeout = 0;
+        Int64 interval = 0;
+    };
 
-    /// For timeout
-    Int64 next_timeout_emit_ts = 0;
-    Int64 timeout_interval = 0;
+    /// For periodic or timeout timer
+    InternalTimer periodic_timer;
```
chenziliang: If we leverage `Stopwatch` we probably can do similar thing ?

src/Processors/Transforms/Streaming/WatermarkTransform.cpp
```diff
@@ -16,40 +13,21 @@ extern const int INVALID_EMIT_MODE;
...
-
 WatermarkTransform::WatermarkTransform(
     const Block & header, WatermarkStamperParamsPtr params_, bool skip_stamping_for_backfill_data_, Poco::Logger * log)
     : ISimpleTransform(header, header, false, ProcessorID::WatermarkTransformID)
     , params(std::move(params_))
     , skip_stamping_for_backfill_data(skip_stamping_for_backfill_data_)
 {
-    watermark = initWatermark(*params, log);
-    assert(watermark);
+    watermark = std::make_unique<WatermarkStamper>(*params, log);
```
chenziliang: `log` -> `logger`

src/Processors/Transforms/Streaming/SessionHelper.cpp
```diff
@@ -232,6 +236,62 @@ void assignWindow(
     }
 }
 
+SessionInfoPtr getLastFinalizedSession(const SessionInfoQueue & sessions)
+{
+    for (const auto & session : sessions | std::views::reverse)
```
chenziliang: Let's use iterator and plain loop instead of ranges

chenziliang:
General comments

1. We have so many different enums class / types regarding emit strategy, watermark strategy, emit on updates, emit finalized windows, bools etc and I think they can be simplified. They are so complex right now, i worried quite a lot it is way more complex than any other proton team member can maintain.
2. Comments, we will need write lots of comments to explain what important functions do etc, how they work etc
3. Free function vs Inheritance class / struct. It appears free function is easier to reuse, but too many free functions and switch case / if else may  also cause maintenance issue.
4. I generally agree we shall refactor watermark, and window assignment into different class etc, but this refactor is expected to happen in a separate PR.
5. Emit behavior, it appears we still didn't emit per group by key for `EMIT ON UPDATE` but for that window, this is probably NOT what want in general. 

For example, if we do `SELECT k, count() FROM test GROUP BY k EMIT ON UPDATE`, it only emits for keys with updates, for keys without any updates, don't emit anything.

For another example, if we do `SELECT window_start, device, count() FROM tumble(device_metrics, 5s) GROUP BY window_start, device EMIT ON UPDATE`, it is expected to emit only on specific device with updates in the current tumble window. For devices without updates in the current window, don't emit.

tests/stream/test_stream_smoke/0034_emit_mode.yaml
yl-lisen: for emit-mode-1-2, it's `group by window_start, window_end`, no other keys. so do emit per window changes

src/Core/Streaming/Watermark.h
```diff
@@ -9,5 +9,48 @@ namespace Streaming
 /// The 1970-01-01 00:00:00 is Int64 value 0, earlier times are negative, so we need to use signed Int64 as watermark
 constexpr Int64 INVALID_WATERMARK = std::numeric_limits<Int64>::min();
 constexpr Int64 TIMEOUT_WATERMARK = std::numeric_limits<Int64>::max();
+
+
+enum class WatermarkStrategy
+{
+    Unknown,
```
chenziliang: We probably don't need both `None / Unknown` since for streaming processing, there has to be a WatermarkStrategy

src/Core/Streaming/Watermark.h
```diff
@@ -9,5 +9,48 @@ namespace Streaming
...
+
+    /// No watermark
+    None,
+
+    /// Watermark is generated based on each event, this is used for `EMIT AFTER WATERMARK WITHOUT DELAY`
```
chenziliang: For every event ? What's the use case ? Probably proc time processing ?

src/Core/Streaming/Watermark.h
chenziliang: Just `Delay` which matches the emit clause is more readable / understandable, OutOfOrder is the result, not the strategy itself

src/Core/Streaming/Watermark.h
```diff
@@ -9,5 +9,48 @@ namespace Streaming
...
+    /// - Not allow time skew
+    Ascending,
+
+    /// Watermark is generated based on last event with a delay interval, this is used for `EMIT AFTER WATERMARK WITH DELAY <delay_interval>`:
+    /// - Allow time skew in a certain range `[<max_event_ts - delay_interval>, <max_event_ts>]`)
+    BoundedOutOfOrderness,
+
+    /// (Built-in): Watermark is generated based on a batch events:
+    /// - Allow time skew in one batch events
+    OutOfOrdernessInBatch,
```
chenziliang: For every batch, we generate a watermark ?

src/Core/Streaming/Watermark.h
```diff
@@ -9,5 +9,48 @@ namespace Streaming
 /// The 1970-01-01 00:00:00 is Int64 value 0, earlier times are negative, so we need to use signed Int64 as watermark
 constexpr Int64 INVALID_WATERMARK = std::numeric_limits<Int64>::min();
 constexpr Int64 TIMEOUT_WATERMARK = std::numeric_limits<Int64>::max();
+
+
+enum class WatermarkStrategy
```
chenziliang:
```
/// WatermarkStrategy defines how proton observes events' timestamps and stamps / generates watermark
/// for them when they flow from the source through the whole processing pipeline.
/// Not every event will be associated with a watermark in general since sometimes events are batched or 
/// sometimes we just don't want to do that because it is not necessary or inefficient. 
```

src/Core/Streaming/Watermark.h
chenziliang: Observing the timestamp, stamp watermark, assign window, emit states are all hand in hand things, we may need think about them carefully.

Previously we basically did everything (except emit states) in the stamper with session window as an exception.

I like the idea to separate the  `Observing the timestamp, stamp watermark` and `assigning window` and `emit states`.

chenziliang: If we separate watermark stamping and window assignment, watermark itself could be pretty simple and having these use cases
1. Ascending - basically proc time processing (EMIT ON PROCTIME) or row number processing  (row number as a timestamp, per row emit in future). For every event in a chunk, stamp a timestamp for each of them. This seems a stateless watermark strategy (probably we still have state incase wall clock rewind / adjusted on that machine ?)
2. BoundedOutOfOrderness - (EMIT ... WITH DELAY), observing the timestamp column for each event, maintain the max timestamp observed so far, if the out of order for an event is within delay, still assign a watermark to it, update the max observed timestamp as we go.
3. OutOfOrdernessInWindowAndBatch - do we need that ? Since for tumble, hopping window, we know the delay automatically ? it seems a special case of 2 ?

chenziliang: For session window, do we need a watermark ?

src/Core/Streaming/Watermark.h
yl-lisen: unused, remove it

src/Core/Streaming/Watermark.h
yl-lisen: For session window, we need events timestamp is in ascending, which is required by session start predicates/end predicates

src/Processors/Transforms/Streaming/WindowAggregatingTransformWithSubstream.h
yl-lisen: yes, there are different finalizing path in AggregatingTransform for different emit strategy

tests/stream/test_stream_smoke/0034_emit_mode.yaml
yl-lisen: will have new commit to tweak the behavior

tests/stream/test_stream_smoke/0034_emit_mode.yaml
yl-lisen: will have new commit to tweak the behavior

src/Core/Streaming/Watermark.h
yl-lisen: Yes, there are more and more actions. Separating these can better maintain them.
In fact, `EmitStrategy` should only be used within AggregatingTransform.

yl-lisen: I will separate them in another PR.

src/Processors/Transforms/Streaming/AggregatingHelper.cpp
yl-lisen: `mergeAndConvertUpdatesToChunk` is existed, it's used in global aggr.
`mergeAndSpliceAndConvertUpdatesToChunk` is used in window aggr.

src/Processors/Transforms/Streaming/WatermarkStamper.h
yl-lisen: do it in a separate PR

src/Processors/QueryPlan/Streaming/AggregatingStep.h
yl-lisen: do it in a separate PR

src/Processors/Transforms/Streaming/AggregatingHelper.h
```diff
@@ -47,6 +63,16 @@ Chunk mergeAndSpliceAndConvertBucketsToChunk(
...
 ChunkPair mergeAndConvertToChangelogChunk(ManyAggregatedDataVariants & data, const AggregatingTransformParams & params);
+
+[[maybe_unused]] ALWAYS_INLINE static bool onlyEmitFinalizedWindows(EmitMode mode) noexcept
+{
+    return mode == EmitMode::Watermark;
+}
+
+[[maybe_unused]] ALWAYS_INLINE static bool onlyEmitUpdates(EmitMode mode) noexcept
+{
+    return static_cast<uint8_t>(mode) & EMIT_UPDATES_MASK;
```
chenziliang: Let's revisit this, it appears something wrong here since we are stuffing bit into an enum ?

PR #534, opened 2024-01-30
feature: [External Stream] Format settings declared on creation time should be taken into account

### LINKED ISSUE #533: [External Stream] Format settings declared on creation time should be taken into account
**Use case**

Currently, format settings can be declared on creation time when create a external stream, for example:

```sql
create or replace external stream users (
           user_id uint32,
           message string,
           timestamp datetime,
           metric float32
           ) settings type='kafka',
           topic='src',
           brokers='localhost:9092',
           data_format='JSONEachRow',
           input_format_skip_unknown_fields=true;
```

The above code uses ` input_format_skip_unknown_fields=true` and expects `SELECT` queries will ignore unknow fileds. However this does not work. One has to specify the setting in the `SELECT` queries like:

```sql
SELECT * FROM users SETTINGS  input_format_skip_unknown_fields=true
```

This is troublesome, because such settings likely should be applied to every `SELECT` queries against that stream.

The format settings declared on creation time should be taken into account.

**Additional context**

zliang-min:
PR checklist:
- Did you run ClangFormat ?

Closes #533 . Please refer to the issue to see the problem this PR resolves.

### DIFF
--- a/src/Storages/ExternalStream/ExternalStreamSettings.h
+++ b/src/Storages/ExternalStream/ExternalStreamSettings.h
@@ -2,6 +2,8 @@
 
 #include <Core/BaseSettings.h>
 #include <Core/Settings.h>
+#include <Formats/FormatFactory.h>
+#include <Interpreters/Context.h>
 
 namespace DB
 {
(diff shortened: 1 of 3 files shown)


yokofly:
need run all test, instead re-run failed tests @zliang-min

src/Storages/ExternalStream/ExternalStreamSettings.h
```diff
@@ -43,6 +45,25 @@ DECLARE_SETTINGS_TRAITS(ExternalStreamSettingsTraits, LIST_OF_EXTERNAL_STREAM_SE
...
+    FormatSettings getFormatSettings(const ContextPtr & context)
+    {
+        FormatFactorySettings settings {};
+        const auto & settings_from_context = context->getSettingsRef();
+
+        /// settings from context have higher priority
+#define SET_CHANGED_SETTINGS(TYPE, NAME, DEFAULT, DESCRIPTION, FLAGS) \
+        if ((NAME).changed) \
+            settings.NAME = (NAME); \
+        if (settings_from_context.NAME.changed) \
```
yl-lisen:
```c++
if (settings_from_context.NAME.changed)
 ...
else if ((NAME).changed)
...
```

yl-lisen: or 
```c++
auto settings = *this; /// clone created settings

...
if (settings_from_context.NAME.changed) /// apply changed setting from context
   ....
```

yl-lisen: This can be improved, as above

zliang-min: Since `this` has more fields than `FormatFactorySettings` (and `this` could grow even bigger if more external streams implemented), thus I used the first one as you suggested.

yl-lisen (approved):

jovezhong:

jovezhong:

PR #546, opened 2024-02-01
feature: External Table ( for ClickHouse )

zliang-min:
PR checklist:
- Did you run ClangFormat ?

New feature for creating External Tabels to connect to non-streaming data sources. In this first implementation, a ClickHouse External Table has been implemented.

Examples:

To create an external table:
```sql
CREATE EXTERNAL TABLE my_first_table SETTINGS type='clickhouse',address='my_clickhouse_server:9000',database='my_database',table='my_table',user='my_user',passowrd='my_password'
```

External tables do not accept the column list, it fetches its table structure from the remote data source. After the table is created, one can use `DESCRIBE my_table` to see its structure.

Then, it can be queried just like any tables/streams:
```sql
-- read data
SELECT * FROM my_external_table;

-- write data
INSERT INTO my_external_table VALUES (...);

-- use it as a materialized view target
CREATE MATERIALIZED VIEW some_view INTO my_external_table AS SELECT ...;
```

To drop an external table:
```sql
DROP EXTERNAL TABLE my_table
```


src/Client/ClickHouseClient.h
```diff
@@ -0,0 +1,56 @@
+#pragma once
```
chenziliang: Does it worth doing

1. Create a subfolder called `ClickHouse` under `Client` to separate the source code at file system folder layer. I feel like we may have more things to stuff here
2. Create a namespace called `ch` or `CH`, so DB::ch or DB::CH::Client instead of ClickHouseClient etc

src/Client/ClickHouseClient.h
```diff
@@ -0,0 +1,56 @@
...
+
+    ConnectionParameters params;
+    std::unique_ptr<Connection> connection;
+    size_t poll_interval;
+
+    bool has_running_query {false};
+    bool cancelled {false};
+    size_t processed_rows {0};
+    Block polled_data;
+    std::unique_ptr<Exception> server_exception {nullptr};
```
chenziliang: we don't need necessary to init unique_ptr to nullptr

src/Client/ClickHouseClient.cpp
```diff
@@ -0,0 +1,286 @@
...
+extern const int UNKNOWN_PACKET_FROM_SERVER;
+extern const int UNEXPECTED_PACKET_FROM_SERVER;
+}
+
+namespace
+{
+
+std::unique_ptr<Connection> createConnection(const ConnectionParameters & parameters)
+{
+    auto ret = std::make_unique<Connection>(
```
chenziliang: Let's leave a comment or file an issue for connection pooling. We can revisit `Connection` concept / pooling systematically for external stream / tables etc

src/Client/ClickHouseClient.cpp
```diff
@@ -0,0 +1,286 @@
...
+}
+
+void ClickHouseClient::reset()
+{
+    cancelled = false;
+    processed_rows = 0;
+    server_exception = nullptr;
+}
+
+void ClickHouseClient::executeQuery(const String & query, const String & query_id, bool fail_quick)
```
chenziliang: Using this high level connection to send select query works but probably slow for ingestion. I was thinking  we could use low level api of connection like `Connection::sendData(...)` for ingestion.

A very good source code we can refer is `StorageDistributed.cpp`, check out its read / write function and you can follow the rabbit hole to get DistributedSink, RemoteSink, RemoteInserter, Interpreter / Cluster, Connection Pooling and how everything compile together to make it efficient by using low level APIs.

chenziliang: StorageDistributed essentially proxy query to remote ClickHouse but at low level with greater efficiency than a plain text send which is not native.

src/Client/ClickHouseClient.cpp
```diff
@@ -0,0 +1,286 @@
...
+    }
+}
+
+void ClickHouseClient::executeInsertQuery(const String & query, const String & query_id)
+{
+    executeQuery(query, query_id);
+    receiveEndOfQuery();
+}
+
+std::optional<Block> ClickHouseClient::pollData()
```
chenziliang: we may want to expose a timeout to `pollData(...)` to let the caller to decide how long we like to poll / wait for network packets etc. We may need have a sentinel to tell user this is the end of stream (done with the query etc) ? Having said this StorageDistributed and the related code has all of these kind of handling.

src/Client/Connection.cpp
```diff
@@ -985,6 +985,10 @@ void Connection::initBlockInput()
         }
 
         block_in = std::make_unique<NativeReader>(*maybe_compressed_in, server_revision);
+        /// proton: starts
```
chenziliang, line 988: What we can do is copying the Connection.h / Connection.cpp and Cluster.h/.cpp etc to `ClickHouse` folder, moving forward, we vendor ClickHouse instead of creating Proton folder and vendor our own protocol code.

chenziliang: Everything in Clickhouse folder, we can update / modify at different paces as we go to make it work for ClickHouse external tables.

src/DataTypes/DataTypeAggregateFunction.cpp
```diff
@@ -159,7 +159,7 @@ SerializationPtr DataTypeAggregateFunction::doGetDefaultSerialization() const
 }
 
 
-static DataTypePtr create(const ASTPtr & arguments)
+static DataTypePtr create(const ASTPtr & arguments/* proton: starts */, bool compatible_with_clickhouse = false/* proton: ends */)
```
chenziliang: For this one liner, we can just mark it something like this

```
static DataTypePtr create(const ASTPtr & arguments, bool compatible_with_clickhouse = false) /// proton : updated
```

src/Databases/DatabaseOnDisk.cpp
```diff
@@ -696,6 +697,18 @@ ASTPtr DatabaseOnDisk::parseQueryFromMetadata(
     auto ast = tryParseQuery(parser, pos, pos + query.size(), error_message, /* hilite = */ false,
                              "in file " + metadata_file_path, /* allow_multi_statements = */ false, 0, settings.max_parser_depth);
 
+    /// proton: starts
```
chenziliang: We can enhance `tryParseQuery` to handle ClickHouse external table case since it is very special (without columns definition); the pros is there may be quite few places calling `tryParseQuery` and we will handle CH external parsing in a centralize place ?

chenziliang: A very simple trick is wrapping tryParseQuery with 2 functions

```
ASTPtr tryParseQuery(...)
{
    auto ast = doTryParseQuery(...); /// existing logic
    if (ast) return ast;
    else return external table parsing..
}
```

src/Interpreters/InterpreterDropQuery.cpp
```diff
@@ -137,6 +142,11 @@ BlockIO InterpreterDropQuery::executeToTableImpl(ContextPtr context_, ASTDropQue
 
         if (query.kind == ASTDropQuery::Kind::Detach)
         {
+            /// proton: starts
+            if (table->isExternalTable())
+                throw Exception("Cannot DETACH external table", ErrorCodes::SYNTAX_ERROR);
```
chenziliang: Let's use a different error code, like NOT_IMPLEMENTED, SUPPORTED instead of SYNTAX_ERROR

src/Parsers/ParserDropExternalTableQuery.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <Parsers/ExpressionElementParsers.h>
+#include <Parsers/ParserDropExternalTableQuery.h>
+
+namespace DB
+{
+
+bool DB::ParserDropExternalTableQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & expected, [[ maybe_unused ]] bool hint)
+{
+    ParserKeyword s_drop("DROP");
+    ParserKeyword s_external_table("EXTERNAL TABLE");
```
chenziliang, line 13: Do we need a different `DROP EXTERNAL TABLE` or just DROP ?

src/Storages/ExternalTable/StorageExternalTable.cpp
```diff
@@ -0,0 +1,99 @@
...
+    /// First, setStorageMetadata should be allowed to fail (the only failable part is getTableStructure function call), otherwise it will block Proton from starting up.
+    /// Second, when it fails, the exception should be caught, otherwise, Proton will fail to start.
+    /// TODO we could use cache to save the table structure, so that when Proton restarts it could read from the cache directly.
+    try
+    {
+        setStorageMetadata(args);
+    }
+    catch (const Exception & e)
+    {
+        LOG_ERROR(&Poco::Logger::get("ExternalTable-ClickHouse" + args.table_id.getFullTableName()),
```
chenziliang: Probably we can just use `ExternalTable` as the logger name and log the details in the message

src/Storages/ExternalTable/StorageExternalTable.cpp
```diff
@@ -0,0 +1,99 @@
...
+                  "Failed to fetch table structure, error: {}. Will keep retrying in background", e.what());
+        background_jobs.scheduleOrThrowOnError([this](){
```
chenziliang, line 42: Since this is async, do we need wait for it done it construtor of StorageExternalTable?

chenziliang: If we wait, seems we don't need make `background_jobs` as a data member and just sync call to `getTableStructure()`

src/Storages/ExternalTable/ExternalTableFactory.h
```diff
@@ -0,0 +1,27 @@
...
+
+    using Creator = std::function<IExternalTablePtr(const String & /*name*/, ExternalTableSettingsPtr /*settings*/)>;
+
+    IExternalTablePtr getExternalTable(const String & name, ExternalTableSettingsPtr settings) const;
+    void registerExternalTable(const String & type, Creator creator);
+
+private:
+    ExternalTableFactory();
+
+    std::unordered_map<std::string, Creator> creators;
```
chenziliang, line 24: Do we need this factory or we can just use existing table engine factory ?

src/Storages/ExternalTable/ClickHouse/ClickHouseSource.cpp
```diff
@@ -0,0 +1,74 @@
...
+        return {};
+    }
+
+    if (!started)
+    {
+        started = true;
+        client->executeQuery(query);
+    }
+
+    LOG_INFO(logger, "polling data");
```
chenziliang: Let's comment this log line since it will be too busy

src/Storages/ExternalTable/ClickHouse/ClickHouseSource.cpp
```diff
@@ -0,0 +1,74 @@
...
+    auto block = client->pollData();
+    client->throwServerExceptionIfAny();
+    if (!block)
+    {
+        LOG_INFO(logger, "no more data");
+        return {};
+    }
+
+    LOG_INFO(logger, "received {} rows", block->rows());
```
chenziliang: Same here

src/Storages/ExternalTable/ClickHouse/ClickHouse.cpp
```diff
@@ -0,0 +1,125 @@
+#include <Client/ClickHouseClient.h>
+#include <DataTypes/DataTypeFactory.h>
+#include <Storages/ExternalTable/ClickHouse/ClickHouse.h>
+#include <Storages/ExternalTable/ClickHouse/ClickHouseSink.h>
+#include <Storages/ExternalTable/ClickHouse/ClickHouseSource.h>
+#include <Storages/ColumnsDescription.h>
+#include "Storages/ExternalTable/ExternalTableFactory.h"
```
chenziliang: Lets use absolute include `<>`

src/Storages/ExternalTable/ClickHouse/ClickHouse.cpp
```diff
@@ -0,0 +1,125 @@
...
+{
+    return std::make_shared<ClickHouseSink>(database, table, metadata_snapshot->getSampleBlock(), connection_params, context, logger);
+}
+
+ColumnsDescription ClickHouse::getTableStructure()
+{
+    ColumnsDescription ret {};
+
+    ClickHouseClient client {connection_params, logger};
+    auto query = "DESCRIBE TABLE " + (database.empty() ? "" : backQuoteIfNeed(database) + ".") + backQuoteIfNeed(table);
```
chenziliang: we can try `auto query = fmt::format("DESCRIBE TABLE {}...")` which may be a bit more readable.

src/Storages/ExternalTable/ClickHouse/ClickHouse.cpp
```diff
@@ -0,0 +1,125 @@
...
+        for (size_t i = 0; i < rows; ++i)
+        {
+            ColumnDescription col_desc {};
+            {
+                const auto & col = block->getByName("name");
+                col_desc.name = col.column->getDataAt(i).toString();
+            }
+            {
+                const auto & col = block->getByName("type");
+                col_desc.type = factory.get(col.column->getDataAt(i).toString(), true);
```
chenziliang: We can explicitly comment for the `true` parameter like `/*is_clickhouse_compatible=*/ true`

src/Storages/ExternalTable/StorageExternalTable.cpp
chenziliang: Sync to get table structure during create seems makes more sense. We have a bool flag `attach` to differentiate if this external table is an existing one load from file system or the initial creation DDL request, for which we can check external stream

jovezhong:
Hi @zliang-min , please test the behavior of connecting to a view or materialized view in ClickHouse.

Probably Proton can still read it but cannot write to it. If the user creates an external table connecting to a ClickHouse view, then try to use it as target for Proton Materialized view. The materialized view DDL should fail, instead of creation is done but no result while MV is running

src/Formats/NativeReader.cpp
```diff
@@ -140,7 +140,7 @@ Block NativeReader::read()
         /// Type
         String type_name;
         readStringBinary(type_name, istr);
-        column.type = data_type_factory.get(type_name);
+        column.type = data_type_factory.get(type_name/* proton: starts */, compatible_with_clickhouse/* proton: ends */);
```
chenziliang, line 143: As time goes, we may choose to vendor NativeReader / Writer by copying to a ClickHouse folder since it is highly protocol, checkpoint / state etc related.

src/Client/ClickHouseClient.cpp
zliang-min: `ClickHouseClient` takes a `ConnectionParams`, which contains `ConnectionTimeouts` which have all kinds of timeout settings. That's why `pollData` itself does not need a timeout parameter.

src/Client/Connection.cpp
zliang-min: Eventually, we probably will need to do this at the point when we start to have significant differences in these files ( no matter it's because we need to customize the protocol for proton, or more changes need to be ported back from ClickHouse ), the downside is that it makes things harder when we need to port changes from the ClickHouse repo. But I think we can leave this to a later time, for now, there is no rush to do this ( I have already created the `ClickHouse` folder, we can move things to that folder any time when we think it's good time to do so ).

src/DataTypes/DataTypeAggregateFunction.cpp
zliang-min: Good to know!

src/Databases/DatabaseOnDisk.cpp
zliang-min: Actually, we don't need to change `tryParseQuery`, which should not be bound to any parser. What need to need change is to update `ParserCreateQuery` and make `PraserCreateExternalTableQuery` part of it.

src/Parsers/ParserDropExternalTableQuery.cpp
zliang-min: I think in general it's a good thing to have a counter part of `CREATE`. The benefits IMO are that: 1. it is more precise, 2. prevent mistakes. For example, `DROP EXTERNAL TABLE foo` will make sure that `foo` is actually an external table, so people won't accidentally drop a stream or view (because of typo, or bad copy/paste).

So, unless there is a strong concern, I will keep this.

src/Storages/ExternalTable/ClickHouse/ClickHouseSource.cpp
zliang-min: Removed.

src/Storages/ExternalTable/ExternalTableFactory.h
zliang-min: I could not find any existing factory that can be used directly for external table implementations, and probably it's not a good idea to mix things up. So having its own factory for external tables should make sense.

src/Formats/NativeReader.cpp
zliang-min: Yeah, if we keep changing things here, we will need to do that.

src/Client/ClickHouseClient.cpp
chenziliang: it could be different, those params settings is for connection, this one is for how long we wait for data which could be different ?

src/Client/Connection.cpp
chenziliang: This is a good concern, vendor which worth more thinking. If we vendor ourself to a different folder when we change our own protocol and leaving the main structure keeping aligned with the CH community makes the porting things easier. Let's keep as it is for now.

src/Storages/ExternalTable/ExternalTableFactory.h
chenziliang: Because of the column descriptions missing ?

src/Client/ClickHouseClient.cpp
chenziliang: Let's defer all of these for second PR since when we reuse most of the existing StorageDistributed related code, lots of things are already considered probably.

src/Client/ClickHouseClient.cpp
zliang-min: Created a ticket for such improvement: https://github.com/timeplus-io/proton/issues/551

src/Client/ClickHouseClient.h
zliang-min: I name the namespace `ClickHouse`, since `ch` or `CH` is ambiguous.

src/Client/ClickHouseClient.cpp
zliang-min: Switched to ConnectionPool.

src/Storages/ExternalTable/StorageExternalTable.cpp
zliang-min: I actually forgot to check `attach`. So the logic should be this ( I commented in the code ):

Two situations:
* Create a new table. In this case, we want it fails the create query if it fails to fetch the columns description. So that users know that there is something with the connection and they can try again once the issue is resolved.
* Attach a table (Proton restarts). In this case, even it fails to fetch the columns description, we want to make sure that:
  - it does not terminate Proton, otherwise Proton will never start again
  - it does not block Proton from starting, otherwise Proton will get stuck
  So, we let it keep retrying in the background, and hoping it will eventually succeeded (until the user drops the table).

zliang-min:
> Hi @zliang-min , please test the behavior of connecting to a view or materialized view in ClickHouse.
> 
> Probably Proton can still read it but cannot write to it. If the user creates an external table connecting to a ClickHouse view, then try to use it as target for Proton Materialized view. The materialized view DDL should fail, instead of creation is done but no result while MV is running

Checked, it works for views and materialized views (read-only of course).

chenziliang (approved):

src/Storages/ExternalTable/ExternalTableFactory.h
zliang-min: Let's make it another discussion and make another PR if needed.

jovezhong:

jovezhong:

PR #577, opened 2024-02-25
Enhancement/issue 576 import large csv files

### LINKED ISSUE #576: FIX CSV import
I tried to import a dataset comprising of over 700 CSV files totallng a billion rows. 
Despite best effort and despite all the help from the engineering team in the slack channel, it's basically impossible.

**How to reproduce**

1) Download dataset
https://support.kraken.com/hc/en-us/articles/360047543791-Downloadable-historical-market-data-time-and-sales-

2) Select the top ten largest CSV files:

du -ah | sort -rh | head -n 10
 31G	.
2.6G	./XBTEUR.csv
2.1G	./XBTUSD.csv
1.6G	./USDTUSD.csv
1.6G	./USDTEUR.csv
1.4G	./ETHUSD.csv
1.2G	./ETHEUR.csv
1.1G	./ETHXBT.csv
663M	./XRPEUR.csv
628M	./EURUSD.csv

Pick any of them, say XBTEUR, create a table:

```
CREATE STREAM default.kraken_xbteur
(
  `timestamp` datetime64(3),
  `price` float64,
  `volume` float64,
  `_tp_time` datetime64(3, 'UTC') DEFAULT timestamp CODEC(DoubleDelta, LZ4),
  INDEX _tp_time_index _tp_time TYPE minmax GRANULARITY 2
)
ENGINE = Stream(1, 1, rand())
PARTITION BY to_YYYYMM(timestamp)
ORDER BY to_start_of_hour(_tp_time)
SETTINGS event_time_column = 'timestamp', index_granularity = 8192
```

Try to import into the newly created table:

```
INSERT INTO kraken_xbteur (timestamp,price,volume) SELECT timestamp,price,volume FROM file('XBTEUR.csv', 'CSV', 'timestamp datetime64(3), price float64, volume float64')  
```

There are multiple problems, as discussed in Slack. 
1) File size limit is set to 100 MB. This is ridiculous. Just fix this in the next release and set it to a more realistic 15 GB.
2) Starting the proton server with a custom configuration fails with multiple errors due to missing env variable and incorrect file permission.

If you use the Rust client to insert batch-wise, you encounter another error related to the size limit regardless of batch-size.

**Additional context**

When this is solved, please publish benchmarks that demonstrates that Proton can handle larger imports.

Specifically, for any serious application proton must demonstrate that it can import the following datasets:

* [1.1 Billion Taxi Rides Benchmarks](https://tech.marksblogg.com/benchmarks.html)
* [11.5 billion rows/ticks Forex ](https://clickhouse.com/blog/getting-data-into-clickhouse-part-3-s3)

In finance, you want to insert all historical trade data into a table, connect to a live stream, and then keep filling the table to keep the trade data up to date. It's self evident that this cannot be done with Proton until large CSV file import gets fixed. And even then I want to see real import benchmarks because I noticed some performance problems.

chenziliang:
PR checklist:
- Did you run ClangFormat ?

Fix #576


src/Interpreters/executeQuery.cpp
```diff
@@ -647,7 +647,7 @@ static std::tuple<ASTPtr, BlockIO> executeQueryImpl(
 
         /// Put query to process list. But don't put SHOW PROCESSLIST query itself.
         ProcessList::EntryPtr process_list_entry;
-        if (!internal && !ast->as<ASTShowProcesslistQuery>())
+        if (!internal && !ast->as<ASTShowProcesslistQuery>() && !ast->as<ASTInsertQuery>())
```
yokofly, line 651: any special reason?

src/Core/Defines.h
```diff
@@ -27,7 +27,7 @@
   *  and if the parts were small, then it would be costly then to combine them.
   */
 #define DEFAULT_INSERT_BLOCK_SIZE \
-    1048449 /// 1048576 - PADDING_FOR_SIMD - (PADDING_FOR_SIMD - 1) bytes padding that we usually have in arrays
+    65409 /// 65536 - PADDING_FOR_SIMD - (PADDING_FOR_SIMD - 1) bytes padding that we usually have in arrays
```
yokofly, line 30: Decrease DEFAULT_INSERT_BLOCK_SIZE if there is not enough RAM.
this is for saving memory, but we can still change this by query level settings.

chenziliang: Ingest block size is usually not tunable at query time and it is not for RAM saving

src/Interpreters/executeQuery.cpp
chenziliang: Concurrent ingest consumes the quota which doesn't make much sense for our use cases

yokofly (approved):

PR #579, opened 2024-02-26
feature: support kafka schema registry for Protobuf and Avro.

zliang-min:
PR checklist:
- Did you run ClangFormat ?

New settings added for using kafka schema registry: `kafka_schema_registry_url` and `kafka_schema_registry_credentials`. Example:

```sql
CREATE EXTERNAL STREAM my_stream (
  -- columns goes here ...
) SETTINGS
type = 'kafka',
brokers = '...',
topic = '...',
data_format = 'ProtobufSingle',
kafka_schema_registry_url = 'http://url.to/my/schema/registry',
kafka_schema_registry_credentials = '$api_key}:${secret}';
```

`kafka_schema_registry_credentials` is optional, it contains the the API key and secret and a `:` in between. Also, kafka schema registry can be used with `data_format = 'Avro'`.

Note: in this PR, schema reference is not supported yet.


src/Formats/FormatSchemaFactory.cpp
```diff
@@ -4,6 +4,10 @@
 #include <IO/ReadBufferFromString.h>
 #include <Processors/Formats/ISchemaWriter.h>
 
+/// proton: starts
```
chenziliang: FormatSchema is added by us ? If yes, we can skip the start/end comment

src/Formats/KafkaSchemaRegistry.h
```diff
@@ -0,0 +1,26 @@
...
+namespace DB
+{
+
+/// A helper class helps working with Kafka schema registry.
+class KafkaSchemaRegistry final
+{
+public:
+    static UInt32 readSchemaId(ReadBuffer & in);
+
+    /// `credentials_` is expected to be formatted in "<username>:<password>".
```
chenziliang: `/// \param credentials_ it is expected to be ...`

src/Formats/KafkaSchemaRegistry.cpp
```diff
@@ -0,0 +1,119 @@
...
+extern const int INCORRECT_DATA;
+}
+
+KafkaSchemaRegistry::KafkaSchemaRegistry(const String & base_url_, const String & credentials_): base_url(base_url_)
+{
+    if (credentials_.empty())
+        return;
+
+    auto pos = credentials_.find(':');
+    if (pos == credentials_.npos)
```
chenziliang: we can do 

```
if (auto pos = credentials_.find(':'); pos == credentials_.npos)
{
}
else
{
}
```

Or 

```
if (auto pos = credentials_.find(':'); pos != credentials_.npos)
{
    credentials.setUsername(credentials_.substr(0, pos));
    credentials.setPassword(credentials_.substr(pos + 1));
}
else
{
    credentials.setUsername(credentials_);
}
```

src/Formats/KafkaSchemaRegistry.cpp
```diff
@@ -0,0 +1,119 @@
...
+{
+
+namespace ErrorCodes
+{
+extern const int INCORRECT_DATA;
+}
+
+KafkaSchemaRegistry::KafkaSchemaRegistry(const String & base_url_, const String & credentials_): base_url(base_url_)
+{
+    if (credentials_.empty())
```
chenziliang: Probably we don't need this test since setting an empty username shall be no-op

src/Formats/KafkaSchemaRegistry.cpp
```diff
@@ -0,0 +1,119 @@
...
+    else
+    {
+        credentials.setUsername(credentials_.substr(0, pos));
+        credentials.setPassword(credentials_.substr(pos + 1));
+    }
+}
+
+String KafkaSchemaRegistry::fetchSchema(UInt32 id)
+{
+    assert(!base_url.empty());
```
chenziliang: we can assert in ctor since this is the only place we got it inited

src/Formats/KafkaSchemaRegistry.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    try
+    {
+        try
+        {
+            Poco::URI url(base_url, "/schemas/ids/" + std::to_string(id));
```
chenziliang: `fmt::format("/schemas/ids/{}", id)` shall be do the work and do it pretty good ?

src/Formats/KafkaSchemaRegistry.cpp
```diff
@@ -0,0 +1,119 @@
+#include <Formats/KafkaSchemaRegistry.h>
+#include <IO/HTTPCommon.h>
+#include <IO/ReadHelpers.h>
+#include <Poco/JSON/Parser.h>
```
chenziliang, line 6: Let's separate 3rd party libs includes to a new include segment

src/Formats/KafkaSchemaRegistry.cpp
```diff
@@ -0,0 +1,119 @@
...
+    {
+        readBinaryBigEndian(magic, in);
+        readBinaryBigEndian(schema_id, in);
+    }
+    catch (const Exception & e)
+    {
+        if (e.code() == ErrorCodes::CANNOT_READ_ALL_DATA)
+        {
+            /// empty or incomplete message without magic byte or schema id
+            throw Exception("Missing magic byte or schema identifier.", ErrorCodes::INCORRECT_DATA);
```
chenziliang: Let's use this format of exception construction

```
throw Exception(error_code, fmt_string, fmt_args);
```

src/Formats/KafkaSchemaRegistry.cpp
```diff
@@ -0,0 +1,119 @@
...
+        }
+        else
+            throw;
+    }
+
+    if (magic != 0x00)
+    {
+        throw Exception("Invalid magic byte before schema identifier."
```
chenziliang: Same here

```
throw Exception(ErrorCodes::INCORRECT_DATA, "Invalid magic byte before schema identifier. Must be zero byte, found 0x{:x} instead", magic);
```

src/Formats/ProtobufSchemas.h
```diff
@@ -21,28 +25,28 @@ namespace protobuf
...
-/// proton: starts
-struct SchemaValidationError;
-using SchemaValidationErrors = std::vector<SchemaValidationError>;
-/// proton: ends
 
 /** Keeps parsed google protobuf schemas parsed from files.
   * This class is used to handle the "Protobuf" input/output formats.
   */
-class ProtobufSchemas : private boost::noncopyable
+class ProtobufSchemas : private boost::noncopyable, public google::protobuf::io::ErrorCollector /* proton: updated */
```
chenziliang: We can remove `private boost::noncopyable` probably since the data member std::mutex is not copyable which effectively make ProtobufSchemas non-copyable.

src/Formats/ProtobufSchemas.cpp
```diff
@@ -106,17 +83,24 @@ const google::protobuf::Descriptor * ProtobufSchemas::getMessageTypeForFormatSch
 }
 
 /// proton: starts
-SchemaValidationErrors ProtobufSchemas::validateSchema(std::string_view schema)
+/// Overrides google::protobuf::io::ErrorCollector:
+void ProtobufSchemas::AddError(int line, google::protobuf::io::ColumnNumber column, const std::string & message)
+{
+    throw Exception(
```
chenziliang: Same exception construction comment as above

src/Processors/Formats/ISchemaWriter.h
```diff
@@ -30,8 +16,8 @@ class IExternalSchemaWriter
 
     virtual ~IExternalSchemaWriter() = default;
 
-    /// Validates the schema input.
-    virtual SchemaValidationErrors validate() = 0;
+    /// Validates the schema input. Should throws exceptions on validation failures.
```
chenziliang: `Should throw`

src/Processors/Formats/Impl/AvroRowInputFormat.cpp
```diff
@@ -663,64 +752,23 @@ class AvroConfluentRowInputFormat::SchemaRegistry
...
-            }
+            return avro::compileJsonSchemaFromString(schema);
         }
-        catch (Exception & e)
+        catch (const avro::Exception & e)
         {
-            e.addMessage("while fetching schema id = " + std::to_string(id));
-            throw;
+          auto ex = Exception(e.what(), ErrorCodes::INCORRECT_DATA);
+          ex.addMessage("while fetching schema id = " + std::to_string(id));
```
chenziliang: we can consider `fmt::format(...)` here if it works

src/Processors/Formats/Impl/AvroRowInputFormat.cpp
```diff
@@ -663,64 +752,23 @@ class AvroConfluentRowInputFormat::SchemaRegistry
...
+          throw std::move(ex);
```
chenziliang, line 685: `throw ex` is good

src/Formats/KafkaSchemaRegistry.cpp
```diff
@@ -0,0 +1,119 @@
...
+            throw;
+        }
+        catch (const Poco::Exception & e)
+        {
+            throw Exception(Exception::CreateFromPocoTag{}, e);
+        }
+    }
+    catch (Exception & e)
+    {
+        e.addMessage("while fetching schema id = " + std::to_string(id));
```
chenziliang: we can consider using `fmt::format(...)` here

src/Formats/KafkaSchemaRegistry.cpp
```diff
@@ -0,0 +1,119 @@
...
+String KafkaSchemaRegistry::fetchSchema(UInt32 id)
+{
+    assert(!base_url.empty());
+
+    try
+    {
+        try
+        {
+            Poco::URI url(base_url, "/schemas/ids/" + std::to_string(id));
+            LOG_TRACE((&Poco::Logger::get("KafkaSchemaRegistry")), "Fetching schema id = {}", id);
```
chenziliang: We can consider cache the `logger` in the registry class since it is used twice already

src/Processors/Formats/Impl/AvroRowInputFormat.cpp
```diff
@@ -729,57 +777,62 @@ static LRUCache<std::string, ConfluentSchemaRegistry>  schema_registry_cache(SCH
 
 static std::shared_ptr<ConfluentSchemaRegistry> getConfluentSchemaRegistry(const FormatSettings & format_settings)
 {
-    const auto & base_url = format_settings.avro.schema_registry_url;
+    /// proton: starts
+    const auto & base_url = format_settings.schema.kafka_schema_registry_url.empty() ? format_settings.avro.schema_registry_url : format_settings.schema.kafka_schema_registry_url;
```
chenziliang, line 702: We are using different urls for avro and kafka but share the same credential : `kafka_schema_registry_credentials` ? Or for avro, `kafka_schema_registry_credentials` will be empty always ?

chenziliang: Do we need `kafka_schema_registry_url` ? or we can just consolidate them like existing code does ?

src/Processors/Formats/Impl/AvroRowInputFormat.cpp
```diff
@@ -729,57 +777,62 @@ static LRUCache<std::string, ConfluentSchemaRegistry>  schema_registry_cache(SCH
...
+    const auto & credentials = format_settings.schema.kafka_schema_registry_credentials;
     auto [schema_registry, loaded] = schema_registry_cache.getOrSet(
-        base_url,
-        [base_url]()
+        base_url + credentials,
+        [base_url, credentials]()
```
chenziliang: if this is a sync call, we can ref capture here for base_url and credentials ? like 

```
[&] {
}
```

src/Processors/Formats/Impl/AvroRowInputFormat.h
```diff
@@ -155,7 +155,7 @@ class AvroConfluentRowInputFormat final : public IRowInputFormat
     std::unordered_map<SchemaId, AvroDeserializer> deserializer_cache;
     const AvroDeserializer & getOrCreateDeserializer(SchemaId schema_id);
 
-    avro::InputStreamPtr input_stream;
+    /// avro::InputStreamPtr input_stream; /* proton: updated */
```
chenziliang: I am thinking eliminating them all for deletion since we know what we are doing to do and we are doing better than CH community.

src/Processors/Formats/Impl/AvroRowInputFormat.cpp
```diff
@@ -910,9 +969,18 @@ void registerInputFormatAvro(FormatFactory & factory)
         ReadBuffer & buf,
         const Block & sample,
         const RowInputFormatParams & params,
-        const FormatSettings & settings)
+        const FormatSettings & settings) -> InputFormatPtr
     {
-        return std::make_shared<AvroRowInputFormat>(sample, buf, params, settings);
+        /// proton: starts
+        /// Use only one format name "Avro" to support both shema registry and non-schema registry use cases, rather than using another name "AvroConfluent"
```
chenziliang: `both schema registry ...`

src/Processors/Formats/Impl/AvroRowInputFormat.cpp
```diff
@@ -910,9 +969,18 @@ void registerInputFormatAvro(FormatFactory & factory)
...
+
+        if (settings.avro.schema_registry_url.empty() && settings.schema.kafka_schema_registry_url.empty())
+            return std::make_shared<AvroRowInputFormat>(sample, buf, params, settings);
```
chenziliang, line 867: Let's comment `/// Non-schema registry case`

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -66,21 +83,43 @@ void ProtobufRowInputFormat::syncAfterError()
...
+        [](ReadBuffer & buf, const Block & sample, IRowInputFormat::Params params, const FormatSettings & settings) -> std::shared_ptr<IInputFormat>
         {
-            return std::make_shared<ProtobufRowInputFormat>(buf, sample, std::move(params),
-                FormatSchemaInfo(settings, "Protobuf", true),
-                with_length_delimiter);
+            if (settings.schema.kafka_schema_registry_url.empty())
+                return std::make_shared<ProtobufRowInputFormat>(
+                    buf, sample, std::move(params), FormatSchemaInfo(settings, "Protobuf", true), /*with_length_delimiter=*/false);
+
+            if (!settings.schema.format_schema.empty())
```
chenziliang, line 111: Since we have done the validation here, do we still need do validation in ProtobufRowInputFormat or  ProtobufConfluentRowInputFormat ? Can we just assert there instead ?

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -66,21 +83,43 @@ void ProtobufRowInputFormat::syncAfterError()
...
+    factory.registerInputFormat(
+        "ProtobufSingle",
+        [](ReadBuffer & buf, const Block & sample, IRowInputFormat::Params params, const FormatSettings & settings) -> std::shared_ptr<IInputFormat>
         {
-            return std::make_shared<ProtobufRowInputFormat>(buf, sample, std::move(params),
-                FormatSchemaInfo(settings, "Protobuf", true),
-                with_length_delimiter);
+            if (settings.schema.kafka_schema_registry_url.empty())
+                return std::make_shared<ProtobufRowInputFormat>(
+                    buf, sample, std::move(params), FormatSchemaInfo(settings, "Protobuf", true), /*with_length_delimiter=*/false);
```
chenziliang, line 109: So we are using `Protobuf` instead of `ProtobufSingle` for length_delimiter = false case ?

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
 }
 
 /// proton: starts
+namespace
+{
+using ConfluentSchemaRegistry = ProtobufConfluentRowInputFormat::SchemaRegistryWithCache;
+#define SCHEMA_REGISTRY_CACHE_MAX_SIZE 1000
+/// Cache of Schema Registry URL + credentials -> SchemaRegistry
+LRUCache<std::string, ConfluentSchemaRegistry>  schema_registry_cache(SCHEMA_REGISTRY_CACHE_MAX_SIZE);
```
chenziliang: Can we wrap this in a function like 

```
auto & schemaRegitryCache()
{
    using ConfluentSchemaRegistry = ProtobufConfluentRowInputFormat::SchemaRegistryWithCache;

    /// Cache of Schema Registry URL + credentials -> SchemaRegistry
    static LRUCache<std::string, ConfluentSchemaRegistry>  schema_registry_cache(/*max_size_=*/1000);
    return schema_registry_cache;
}
```

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+
+std::shared_ptr<ConfluentSchemaRegistry> getConfluentSchemaRegistry(const String & base_url, const String & credentials)
+{
+    auto [schema_registry, loaded] = schema_registry_cache.getOrSet(
+        base_url + credentials,
+        [base_url, credentials]()
```
chenziliang: ref capture

src/Processors/Formats/Impl/ProtobufRowInputFormat.h
```diff
@@ -7,6 +7,9 @@
 #    include <Processors/Formats/IRowInputFormat.h>
 #    include <Processors/Formats/ISchemaReader.h>
 #    include <Processors/Formats/ISchemaWriter.h>
+/// proton: starts
```
chenziliang, line 11: Let's insert an empty line before this segment

src/Processors/Formats/Impl/ProtobufRowInputFormat.h
```diff
@@ -61,12 +69,36 @@ class ProtobufSchemaReader : public IExternalSchemaReader
 };
 
 /// proton: starts
+/// Confluent framing + Protobuf binary datum encoding. Mainly used for Kafka.
+/// Uses 3 caches:
+/// 1. global: schema registry cache (base_url -> SchemaRegistry)
```
chenziliang: `(base_url + credentials -> SchemaRegistry)` ?

src/Processors/Formats/Impl/ProtobufRowInputFormat.h
```diff
@@ -61,12 +69,36 @@ class ProtobufSchemaReader : public IExternalSchemaReader
...
+/// 2. SchemaRegistry: schema cache (schema_id -> schema)
+/// 3. ProtobufConfluentRowInputFormat: deserializer cache (schema_id -> AvroDeserializer)
+/// This is needed because KafkaStorage creates a new instance of InputFormat per a batch of messages
```
chenziliang, line 78: Wondering how do we do cache invalidation ? do we support it now and how it works in prod for schema evolution / changes ?

src/Processors/Formats/Impl/ProtobufRowInputFormat.h
```diff
@@ -61,12 +69,36 @@ class ProtobufSchemaReader : public IExternalSchemaReader
...
+class ProtobufConfluentRowInputFormat final : public IRowInputFormat
+{
+public:
+    ProtobufConfluentRowInputFormat(ReadBuffer & in_, const Block & header_, Params params_, const FormatSettings & format_settings_);
+    String getName() const override { return "ProtobufConfluentRowInputFormat"; }
+
+    // void setReadBuffer(ReadBuffer & buf) override;
+
+    class SchemaRegistryWithCache;
```
chenziliang, line 87: We can forward declare before the class definition ?

chenziliang: Ok, it is expected to be an inner class

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+
+        auto const * ret = registry_pool()->BuildFile(descriptor);
+        auto mt_count = ret->message_type_count();
+        if (mt_count < 1)
+            throw Exception(ErrorCodes::INVALID_DATA, "No message type in schema");
+        return ret;
+    }
+
+    KafkaSchemaRegistry registry;
+    google::protobuf::DescriptorPool registry_pool_;
```
chenziliang: `registry_pool`

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+
+    google::protobuf::DescriptorPool* registry_pool() { return &registry_pool_; }
```
chenziliang: `auto * regstryPool() noexcept` and do we need it ?

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
chenziliang: I feel like it shall be `descriptor_pool` ?

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+    const google::protobuf::FileDescriptor * getSchema(uint32_t id)
+    {
+        const auto * loaded_descriptor = registry_pool()->FindFileByName(std::to_string(id));
+        if (loaded_descriptor)
+            return loaded_descriptor;
+
+        return fetchSchema(id);
+    }
+
+    const google::protobuf::FileDescriptor* fetchSchema(uint32_t id)
```
chenziliang: We will need some clang-format for the code we added

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+            descriptor = descriptor->nested_type(i);
+        }
+
+        return descriptor;
+    }
+
+private:
+    const google::protobuf::FileDescriptor * getSchema(uint32_t id)
+    {
+        const auto * loaded_descriptor = registry_pool()->FindFileByName(std::to_string(id));
```
chenziliang: Do we need call `registry_pool()` function ?

```
if (const auto * loaded_descriptor = descriptor_pool.FindFileByName(std::to_string(id)); loaded_descriptor != nullptr)
    return loaded_descriptor;
```

chenziliang: Does `id` uniquely identify the descriptor ?

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+        if (loaded_descriptor)
+            return loaded_descriptor;
+
+        return fetchSchema(id);
+    }
+
+    const google::protobuf::FileDescriptor* fetchSchema(uint32_t id)
+    {
+        auto schema = registry.fetchSchema(id);
+        std::string schema_content {schema.data(), schema.size()};
```
chenziliang: since `registry.fetchSchema(id)` already returns a full copy of a schema string, do we need do another copy (`schema_content`) ?

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+        google::protobuf::io::ArrayInputStream input{schema.data(), static_cast<int>(schema.size())};
+        google::protobuf::io::Tokenizer tokenizer(&input, this);
+        google::protobuf::FileDescriptorProto descriptor;
+        descriptor.set_name(std::to_string(id));
+        google::protobuf::compiler::Parser parser;
+        parser.RecordErrorsTo(this);
+        parser.Parse(&tokenizer, &descriptor);
+
+        auto const * ret = registry_pool()->BuildFile(descriptor);
```
chenziliang: We will need check nullness of ret

```
if (const auto * descriptor = descriptor_pool.BuildFile(descriptor_proto); descriptor && descriptor->message_type_count() >= 1)
   return descriptor;

throw ...
```

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+    auto schema_id = KafkaSchemaRegistry::readSchemaId(*in);
+
+
+    Int64 indexes_count = 0;
+    readVarInt(indexes_count, *in);
+
+    std::vector<Int64> indexes;
+    indexes.reserve(indexes_count < 1 ? 1 : indexes_count);
+
+    if (indexes_count < 1)
```
chenziliang, line 265: nit : Let's put a curly braces here for the `if` clause

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+        indexes.push_back(0);
+    else
+    {
+        for (size_t i = 0; i < static_cast<size_t>(indexes_count); i++)
```
chenziliang: `for (Int64_t i = 0; ...`

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+    ProtobufReader reader {*in};
+    serializer = ProtobufSerializer::create(
+        header.getNames(),
+        header.getDataTypes(),
+        missing_column_indices,
+        *registry->getMessageType(schema_id, indexes),
+        /*with_length_delimiter=*/false,
+        reader);
+
+    size_t row_num = columns.empty() ? 0 : columns[0]->size();
```
chenziliang, line 290: is it possible we have empty columns to read ?

chenziliang: i c, `columns` are grown automatically when we read, and initially no rows in the columns

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+    void AddError(int line, google::protobuf::io::ColumnNumber column, const std::string & message) override
+    {
+        throw Exception(ErrorCodes::INVALID_DATA, "Failed to parse schema, line={}, column={}, message={}", line, column, message);
+    }
+
+    const google::protobuf::Descriptor * getMessageType(uint32_t schema_id, const std::vector<Int64> & indexes)
+    {
+        assert(!indexes.empty());
+
+        const auto *fd = getSchema(schema_id);
```
chenziliang: clang-format

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+        auto mt_count = fd->message_type_count();
+        if (mt_count < indexes[0] + 1)
+            throw Exception(ErrorCodes::INVALID_DATA, "Invalid message index={} max_index={}", indexes[0], mt_count);
+
+        const auto *descriptor = fd->message_type(indexes[0]);
```
chenziliang: clang-format

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+
+        for (auto i : std::vector<Int64>(indexes.begin() + 1, indexes.end()))
```
chenziliang: Let's use std::span in `std::ranges` which avoids a copy

chenziliang: We may need put some comments here to explain the loop. It appears to me, we start from the initial index and follow the chain to get the last descriptor

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
```diff
@@ -100,20 +139,162 @@ NamesAndTypesList ProtobufSchemaReader::readSchema()
...
+        reader);
+
+    size_t row_num = columns.empty() ? 0 : columns[0]->size();
+    if (!row_num)
+        serializer->setColumns(columns.data(), columns.size());
+
+    serializer->readRow(row_num);
+
+    row_read_extension.read_columns.clear();
+    row_read_extension.read_columns.resize(columns.size(), true);
```
chenziliang: if missing_column_indices is zero, i was thinking we can optimize this as  following ? and i recalled the framework (parent driving template function) does the clear before the call

```
assert(row_read_extension.read_columns.empty());
if (!missing_column_indices.empty())
{
    row_read_extension.read_columns.resize(columns.size(), true);
     for (size_t column_idx : missing_column_indices)
        row_read_extension.read_columns[column_idx] = false;
}

```

src/Formats/FormatSchemaFactory.cpp
zliang-min: Oops, that's right.

chenziliang:
In general, it looks good and let's clean up the old code which is commented out

src/Processors/Formats/Impl/AvroRowInputFormat.cpp
zliang-min: `throw ex` reports `Throw expression should throw anonymous temporary values instead`.

src/Processors/Formats/Impl/AvroRowInputFormat.cpp
zliang-min: Actually we probably should remove `avro.schema_registry_url` eventually. I don't know why it was designed in that way, because schema registry is not Avro specific. `kafka_schema_registry` is more generic, and it can be used in Protobuf, Avro, and JSON Schema (if we want to support in the future).

`kafka_schema_registry_credentials` is needed for authentication. The previous Avro implementation does not support authentication, which means it could not work with Conflent cloud.

src/Processors/Formats/Impl/AvroRowInputFormat.cpp
zliang-min: Yes, I think it should be fine, because eventually, `KafkaSchemaREgistry` will copy the URL and credentials.

src/Processors/Formats/Impl/AvroRowInputFormat.h
zliang-min: Yeah, I think that should be fine, because these code are quite specific and reasonably easy to understand if even we want to pick up some changes from CH in the future.

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
zliang-min: No, I didn't change that in this PR. It's still `ProtobufSingle`. It's that for `FormatSchemaInfo` they all belong to `"Protobuf"`.

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
zliang-min: Neither ProtobufRowInputFormat nor ProtobufConfluentRowInputFormat duplicates this validation, here is the only place does it.

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
zliang-min: I put a comment there and put a link to the explanation.

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
zliang-min: Yes, schema IDs are unique, each one points to a specific version of a schema.

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
zliang-min: That's a useless code being forgotten to be deleted.

src/Processors/Formats/Impl/ProtobufRowInputFormat.h
zliang-min: I think we don't need to worry about the invalidation much for now. Schemas (on a schema registry) are immutable, i.e. someone updates a schema (also called evolves the schema), a new one will be created with a new schema ID, so in fact, the existing cache is still valid ( one can still use the old schema to write data, it's toally valid ).

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
zliang-min: lI applied your version, and it works fine in my tests, so I think this is fine.

src/Processors/Formats/Impl/ProtobufRowInputFormat.cpp
zliang-min: No, it's actually not needed (tested), I believe it's a left-over after some refactor.

chenziliang (approved):

Issue #607, opened 2024-03-15
aggregation function disctinct has some problems.


Sake169:
The `distinct` function does not support multiple arguments, but the clickhouse has already supported.
Cickhouse:
```sql
CREATE TABLE users (uid Int16, name String, age Int16) ENGINE=Memory;

INSERT INTO users VALUES (1231, 'John', 33);
INSERT INTO users VALUES (1231, 'John', 48);
INSERT INTO users VALUES (8888, 'Alice', 50);

SELECT countDistinct(uid, name) FROM users;
-- output
2
```
Proton:
```sql
drop stream if exists test;
create stream if not exists test(id int, value1 int, value2 int);
select count_distinct(value1, value2) from test;
-- output
Received exception from server (version 1.5.3):
Code: 42. DB::Exception: Received from localhost:8463. DB::Exception: Aggregate function count requires either zero or one argument. (NUMBER_OF_ARGUMENTS_DOESNT_MATCH)
```

Sake169:
![image](https://github.com/timeplus-io/proton/assets/66816439/3799e67b-16d4-4aae-b122-b678315e95fb)
Because proton replaced the `count_distinct` with their own streaming function (`count_distinct_streaming`), so in this function it won't replac the `count_distinct_streaming` to `uniq_exact` function, since it doesn't match.But in clickhouse the `countDistinct` will be replaced with `uniqExact` in this function.And it can handle multiple distinct keys.

PS:Clickhouse only replaces the `countDistinct` if it has multiple distinct keys.

PR #608, opened 2024-03-15
distinct_retract supports numeric processing

### LINKED ISSUE #519: improve distinct_retract aggregation logic.
distinct_retract is a connector, `distinct` means filter the duplicate data to the next aggregation operator, and `retract` is the internal impl about [CDC](https://en.wikipedia.org/wiki/Change_data_capture) with changelog. (we use virtual column tp_delta, +1 means data in, and -1 means retract.)

## short: this issue wanna tackle 2 things:
1. replace the map->hashmap, this file shall be found in  https://github.com/timeplus-io/proton/blob/develop/src/AggregateFunctions/Streaming/CountedValueHashMap.h
2. add extra logic for numeric data.

## detailed 
Since we have introduced counted_value_hash_map in https://github.com/timeplus-io/proton/pull/484, the distinct still uses an ordered map, actually, the distinct connector does not care about the order of the element.
the related code can be found at
https://github.com/timeplus-io/proton/blob/00d0411f03caeaf0faee06431dab81b0874910ea/src/AggregateFunctions/Streaming/AggregateFunctionDistinctRetract.h#L19-L25

another thing we wanna improve/refactor is currently we use the `stringRef` style for all the input, recalling another distinct implementation, we notice the function has to do dispatch with input data type. that's say, current distinct_retract needs extra serialize/deser for numeric data, but, we are input the numeric data, we do not need such string serialization overhead.

https://github.com/timeplus-io/proton/blob/00d0411f03caeaf0faee06431dab81b0874910ea/src/AggregateFunctions/Streaming/AggregateFunctionDistinct.cpp#L37-L63

upstream https://github.com/timeplus-io/proton/pull/426

Sake169:
PR checklist:
- Did you run ClangFormat ?

Current the `distinct_retract` function serialized no matter what the datatype is into string, it is very expensive, now it will just serialize the multiple key into string just like the `distinct_streaming`.
this should close #519


jovezhong:
Thank you for contributing this, @Sake169 
Let me start the build process, in the meanwhile, we will look into the code

jovezhong:
The related ticket is https://github.com/timeplus-io/proton/issues/607

Issue #544, opened 2024-02-01
Reuse same pooled Kafka(Simple)Consumer 


chenziliang:
**Describe what enhancement you'd like to have**

librdkafka shall allow reuse an existing Kafka consumer for different topic partition.

For example, 
at time t1, user queries topic1 by using external stream, we create a Kafka consumer and pool it.
at time t2 while t1 query is still running, user queries topic2 by using external stream, for now we will create another Kafka consumer and pool it.

We shall be able to reuse Kafka consumer at t1 for t2 query instead of recreating another Kafka consumer which is actually very expensive and most importantly it is not scale.

Probably same for producer if we create separate producer on the fly for every external stream and if there are hundreds of them.

zliang-min:
A consumer is bound to a specific topic parition ( using the legacy simpler consumer API ) or a specific list of paritions (using the new consumer API ). So, I am not quite sure what it exactly means to resue a consumer, or maybe you are talking about the `KafkaWALSimpleConsumer` ? But I want to understand more about why "recreating another kafka consumer is very expensive". At the end of the day, in order to consume from another topic/partition, another consumer has to be created anyways ( from the pont of view of `librdkafka` ). Could you elaborate, please?

From `librdkafka` side, what cachable are `rd_kafkt_t` and `rd_kafka_topic_t` ( if we use the new consumer API, we don't even need to cache `rd_kafka_topic_t` ).

BTW, I think we should refactor the pooling/caching for kafka external stream. Currently it uses `KafkaWALPool`, which I think (from the code) was originally deisnged for logstore storage. It looks to me that it's not the best idea to mix logstore and external stream together. It also has leaking issues like [this](https://github.com/timeplus-io/proton/issues/236). There are some useful concepts/tools existing in the `klog` namespace that can be reused in external stream, but it does not mean we have to resue everything there to make it more complex than it has to be.

zhongyuankai:
When hundreds of kafka streams are created, tens of thousands of threads will be created, and the number of brokers in the kafka cluster is 160. Is there any good solution?

zliang-min:
[...]

PR #618, opened 2024-03-20
Feature/kafka properties for source

### LINKED ISSUE #236: Kafka external stream keeps reporting warning after fails to be created
**Describe what's wrong**

When I tried to create a kafka external stream which connects to a kafka cluster which does not exist, on the client side, it reports the following error as expected:
```
Code: 2003. DB::Exception: Received from 127.0.0.1:8463. DB::Exception: foo topic does't exist. (RESOURCE_NOT_FOUND)
```

However, on the server side, it keeps printing warning logs:
```
%3|1698688956.289|FAIL|rdkafka#consumer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/bootstrap: Connect to ipv4#127.0.0.1:9092 failed: Connection refused (after 0ms in state CONNECT)
2023.10.30 11:02:36.289582 [ 89967498 ] {} <Warning> KafkaWALSimpleConsumer: Error occurred, error=Local: Broker transport failure, reason=localhost:9092/bootstrap: Connect to ipv4#127.0.0.1:9092 failed: Connection refused (after 0ms in state CONNECT)

%3|1698688957.294|FAIL|rdkafka#consumer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/bootstrap: Connect to ipv6#[::1]:9092 failed: Connection refused (after 0ms in state CONNECT)
2023.10.30 11:02:37.294835 [ 89967498 ] {} <Warning> KafkaWALSimpleConsumer: Error occurred, error=Local: Broker transport failure, reason=localhost:9092/bootstrap: Connect to ipv6#[::1]:9092 failed: Connection refused (after 0ms in state CONNECT)
```

That means we have resource leaked.

**How to reproduce**

**Additional context**

### LINKED ISSUE #291: when query external stream with new shard added in kafka, it can't work
Test step: 

1. create a multi shard topic (such as a=3)
2. create external stream
3. Simple test such as (write kafka and read external stream, write external stream and read kafka)
4. Add partition in kafka(rpk topic add-partitions [TOPICS...] --num [#] [flags]) and check if partition is add successfully
5. Specify the new shard you want read  in external stream(settings shards=’shard_num’)
6. Specify the new shard you want write  in kafka(rpk topic produce topic_name -p shard_num)

```sql

-- redpanda
-- rpk topic create test -p 3 

-- proton
drop stream if exists test_external1;
create external stream if not exists test_external1(raw string) settings type = 'kafka', brokers = 'redpanda-1:9092', topic = 'test';
select * from test_external1;

-- redpanda
-- rpk topic produce test
1111
Produced to partition 0 at offset 0 with timestamp 1700211318417.
2222
Produced to partition 1 at offset 0 with timestamp 1700211319772.
3333
Produced to partition 2 at offset 0 with timestamp 1700211321669.

-- proton
┌─raw──┐
│ 1111 │
└──────┘
┌─raw──┐
│ 2222 │
└──────┘
┌─raw──┐
│ 3333 │
└──────┘

-- redpanda
-- rpk topic add-partitions test --num 2
redpanda@6576278380c2:/$ rpk topic list test
NAME  PARTITIONS  REPLICAS
test  5           1
redpanda@6576278380c2:/$ 

-- proton
select * from test_external1 settings shards='4'
-- error:Code: 471. DB::Exception: Received from localhost:8463. DB::Exception: Invalid topic partition 4 for topic test, biggest partition is 3. (INVALID_SETTING_VALUE)
```

### LINKED ISSUE #543: External kafka stream's `properties` are not honored
**Use case**

```
CREATE EXTERNAL STERAM kafka(raw string) SETTINGS properties='enable.auto.commit=true';

SELECT * FROM kafka;
```

**Additional context**

### LINKED ISSUE #617: External stream still gets created if sasl_mechanism is wrong
**Describe what's wrong**

@ye11ow reported in Slack:

> I’m trying to create an ext stream with wrong sasl type
> ```
> CREATE EXTERNAL STREAM ext_calvin_test(raw string)  
> SETTINGS type='kafka',  
> brokers='pkc-ld537.ca-central-1.aws.confluent.cloud:9092',  
> topic='calvin_test',  
> security_protocol='SASL_SSL',  
> sasl_mechanism='SCRAM-SHA-512',  <- This is wrong so the ext stream should not work
> username='YYY',  
> password='XXX'
> ```
> What I expect is that the creation will fail. However, I was able to create the ext stream

**How to reproduce**

**Additional context**

zliang-min:
PR checklist:
- Did you run ClangFormat ?

This PR has completely rewritten how producers and consumers are managed in Kafka external streams.

First, all librdkafka resource management code (for external stream) have been moved to the `RdKafka` namespace. In Kafka external stream implementaion, it no longer uses the code from the `klog` namespace (directly). The goal is that, eventually, the Kafka external stream implementation and `klog` will all use `RdKafka` for communicating with Kafka via the librdkafka library. And `klog` will only be used the Proton internal Kafka log stream.

With this PR, resources from librdkafka will be shared by multiple sources/sinks of the same external stream as much as possible. This will reduce resource consumption compared to the current implementation. Esp. when there are a lot of topics and topics have a lot partitions. refers #544 

This also fixed the resource leakage issue with the current pooling implementation. closes #236 

This also fixed external stream won't pick up new paritions when used as a source. closes #291 ( Note that, a running MV or query that selects from an external stream won't automatically updated to pick up the new paritions, it needs to be re-created or re-run ).

This also fixed the issue that the `properties` setting does not work for external stream source. closes #543 

This also fixed the issue that an external stream still gets created even some settings are not correct. closes #617 

This also fixed an issue when using `ProtobufSingle` to write multiple rows of data to an external stream. (no ticket)


src/Storages/ExternalStream/StorageExternalStreamImpl.h
```diff
@@ -46,7 +46,11 @@ class StorageExternalStreamImpl : public std::enable_shared_from_this<StorageExt
 
     FormatSettings getFormatSettings(const ContextPtr & context) const
     {
-        return settings->getFormatSettings(context);
+        auto ret = settings->getFormatSettings(context);
+        /// This is needed otherwise using an external stream with ProtobufSingle format as the target stream
+        /// of a MV (or in `INSERT ... SELECT ...`), i.e. more than one rows sent to the stream, exception will be thrown.
+        ret.protobuf.allow_multiple_rows_without_delimiter = true;
```
yl-lisen, line 52: it seems always allow write multiple rows into external stream - Kafka .
What scenarios require reporting an error?

src/Storages/ExternalStream/Kafka/Topic.cpp
```diff
@@ -0,0 +1,52 @@
...
+extern const int RESOURCE_NOT_FOUND;
+}
+
+namespace RdKafka
+{
+
+Topic::Topic(rd_kafka_t & rk, const std::string & name) : rdk(rk)
+{
+    /// rd_kafka_topic_new takes ownership of topic_conf
+    rkt.reset(rd_kafka_topic_new(&rk, name.c_str(), /*conf=*/nullptr));
```
yl-lisen, line 18: `rd_kafka_topic_new` is a legacy API ,  and don't support set the partition num ?

`rd_kafka_NewTopic_new` would be better to use this interface

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
```diff
@@ -69,15 +74,23 @@ KafkaSource::KafkaSource(
 
 KafkaSource::~KafkaSource()
 {
-    LOG_INFO(log, "Stop streaming reading from topic={} shard={}", consume_ctx.topic, consume_ctx.partition);
-    consumer->stopConsume(consume_ctx);
+    LOG_INFO(logger, "Stop consuming from topic={} shard={}", topic->name(), shard);
```
yl-lisen: I think should move this log into `if (consume_started) { ...} `

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -250,26 +175,28 @@ KafkaSink::KafkaSink(
     /// Polling message deliveries.
     background_jobs.scheduleOrThrowOnError([this, refresh_interval_ms = static_cast<UInt64>(topic_refresh_interval_ms)]() {
         auto metadata_refresh_stopwatch = Stopwatch();
+        /// Use a small sleep interval to avoid blocking operation for a long just (in case refresh_interval_ms is big).
+        UInt64 sleep_ms = 500;
+        if (sleep_ms > refresh_interval_ms)
+            sleep_ms = refresh_interval_ms;
```
yl-lisen: sleep_ms = std::min(500, refresh_interval_ms)

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -5,21 +5,45 @@
...
+    static void onThrottle(struct rd_kafka_s * rk, const char * broker_name, int32_t broker_id, int throttle_time_ms, void * opaque);
+
     Kafka(IStorage * storage, std::unique_ptr<ExternalStreamSettings> settings_, const ASTs & engine_args_, bool attach, ExternalStreamCounterPtr external_stream_counter_, ContextPtr context);
     ~Kafka() override = default;
 
-    void startup() override { }
-    void shutdown() override { }
+    void startup() override { LOG_INFO(logger, "Starting Kafka External Stream"); }
+    void shutdown() override {
+        LOG_INFO(logger, "Shutting down Kafka External Stream");
```
yl-lisen, line 38: we can add external stream name info

yokofly:
failed smoke
```
2024-03-20T10:01:56.2008603Z =========================== short test summary info ============================
2024-03-20T10:01:56.2009692Z FAILED test_stream_smoke/test_query_smoke.py::test_run[external_stream-3-test external stream multi_shard]
2024-03-20T10:01:56.2011119Z FAILED test_stream_smoke/test_query_smoke.py::test_run[external_stream-4-test external stream multi_shard and seek_to]
2024-03-20T10:01:56.2014460Z FAILED test_stream_smoke/test_query_smoke.py::test_run[external_stream-5-test external stream multi_shard and seek_to, now the test framework doesn't support validate the data read from kafkaseek_to result.So we insert data into external stream and read from external stream to validate the result.The data with the same shard_expr result should be divided into the same shard.]
2024-03-20T10:01:56.2017349Z ================== 3 failed, 844 passed in 859.17s (0:14:19) ===================
```

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -5,21 +5,45 @@
...
+        /// Must release all resources here rather than relying on the deconstructor.
+        /// Because the `Kafka` instance will not be destroyed immediately when the external stream gets dropped.
+        consumer_pool.reset();
+        if (producer_topic)
+            producer_topic.reset();
+        if (producer)
+            producer.reset();
```
yl-lisen, line 45: Is the release order required ? 
1) release producer 
2) release consumer
3) release topic

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -36,37 +60,53 @@ class Kafka final : public StorageExternalStreamImpl
...
     const String & dataFormat() const override { return data_format; }
     const String & topic() const { return settings->topic.value; }
-    const klog::KConfParams & properties() const { return kafka_properties; }
-    const klog::KafkaWALAuth & auth() const noexcept { return *auth_info; }
     const ASTPtr & shardingExprAst() const { assert(!engine_args.empty()); return engine_args[0]; }
     bool hasCustomShardingExpr() const;
-    klog::KafkaWALSimpleConsumerPtr getConsumer(int32_t fetch_wait_max_ms = 200) const;
+
+    RdKafka::Producer & getProducer();
+    RdKafka::Topic & getProducerTopic();
```
yl-lisen, line 71: getTopic()

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -36,37 +60,53 @@ class Kafka final : public StorageExternalStreamImpl
...
-
     ASTPtr message_key_ast;
+    Int32 topic_refresh_interval_ms = 0;
+    std::vector<Int32> shards_from_settings;
+
+    ConfPtr conf;
+    /// The Producer instance and Topic instance can be used by multiple sinks at the same time, thus we only need one of each.
+    std::mutex producer_mutex;
+    std::unique_ptr<RdKafka::Producer> producer;
+    std::unique_ptr<RdKafka::Topic> producer_topic;
```
yl-lisen, line 103: `topic`

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -64,10 +217,25 @@ Kafka::Kafka(IStorage * storage, std::unique_ptr<ExternalStreamSettings> setting
         settings->set("one_message_per_row", true);
     }
 
+    if (!context->getSettingsRef().shards.value.empty())
+        shards_from_settings = parseShards(context->getSettingsRef().shards.value);
```
yl-lisen: should parse it in `read()`

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -276,45 +363,185 @@ void Kafka::validateMessageKey(const String & message_key_, IStorage * storage,
...
+    std::scoped_lock lock(producer_mutex);
+    /// Check again in case of losing the race
+    if (producer_topic)
+        return *producer_topic;
+
+    auto topic_ptr = std::make_unique<RdKafka::Topic>(*getProducer().getHandle(), topic());
+    producer_topic.swap(topic_ptr);
+
+    return *producer_topic;
+}
```
yl-lisen: it seems the topic will be created at the first time it's used.
we should create it during initialize the external stream  - Kafka

yl-lisen: For producer or consumer, we can initialize it lazily

src/Storages/ExternalStream/Kafka/Topic.h
```diff
@@ -0,0 +1,33 @@
...
+
+class Topic : boost::noncopyable
+{
+public:
+    Topic(rd_kafka_t & rk, const std::string & name);
+    ~Topic() = default;
+
+    rd_kafka_topic_t * getHandle() const { return rkt.get(); }
+    std::string name() const { return rd_kafka_topic_name(rkt.get()); }
+    int describe() const;
```
yl-lisen: need a comment to explain the return value ?

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -276,45 +363,185 @@ void Kafka::validateMessageKey(const String & message_key_, IStorage * storage,
...
-        std::scoped_lock lock(shards_mutex);
-        /// Recheck again in case in-parallel query chimes in and init the shards already
-        if (shards == 0)
-        {
-            /// We haven't describe the topic yet
-            auto result = getConsumer()->describe(settings->topic.value);
-            if (result.err != ErrorCodes::OK)
-                throw Exception(ErrorCodes::RESOURCE_NOT_FOUND, "{} topic doesn't exist", settings->topic.value);
+    auto consumer = getConsumer();
+    RdKafka::Topic rkt {*consumer->getHandle(), topic()};
```
yl-lisen: just validate topic here, do we need consumer ?

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -276,45 +363,185 @@ void Kafka::validateMessageKey(const String & message_key_, IStorage * storage,
...
+    const StorageSnapshotPtr & storage_snapshot,
+    SelectQueryInfo & query_info,
+    ContextPtr context,
+    QueryProcessingStage::Enum /*processed_stage*/,
+    size_t max_block_size,
+    size_t /*num_streams*/)
+{
+    auto consumer = getConsumer();
+    /// The topic_ptr can be shared between all the sources in the same pipe, because each source reads from a different partition.
+    auto topic_ptr = std::make_shared<RdKafka::Topic>(*consumer->getHandle(), topic());
```
yl-lisen: create topic for each select query ?

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -276,45 +363,185 @@ void Kafka::validateMessageKey(const String & message_key_, IStorage * storage,
...
+
+    /// User can explicitly consume specific kafka partitions by specifying `shards=` setting
+    /// `SELECT * FROM kafka_stream SETTINGS shards=0,3`
+    std::vector<int32_t> shards_to_query;
+    if (!shards_from_settings.empty())
+        shards_to_query = shards_from_settings;
+    else
+    {
+        /// The number of paritions could have been changed, always query the number before
+        auto parition_count = topic_ptr->describe();
```
yl-lisen: partition_count should be a constant value, can we cache it ?  and cache the all_shards_to_query

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -276,45 +363,185 @@ void Kafka::validateMessageKey(const String & message_key_, IStorage * storage,
...
+    size_t /*num_streams*/)
+{
+    auto consumer = getConsumer();
+    /// The topic_ptr can be shared between all the sources in the same pipe, because each source reads from a different partition.
+    auto topic_ptr = std::make_shared<RdKafka::Topic>(*consumer->getHandle(), topic());
+
+    /// User can explicitly consume specific kafka partitions by specifying `shards=` setting
+    /// `SELECT * FROM kafka_stream SETTINGS shards=0,3`
+    std::vector<int32_t> shards_to_query;
+    if (!shards_from_settings.empty())
```
yl-lisen: the `shards_from_settings` was parsed in ctor, but the settings is for every query , so we need parse it from query_context .
here you can get the shards from \param `context->getSettingsRef()`.

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -276,45 +363,185 @@ void Kafka::validateMessageKey(const String & message_key_, IStorage * storage,
...
+    }
+    else*/
+    {
+        /// For queries like `SELECT count(*) FROM tumble(table, now(), 5s) GROUP BY window_end` don't have required column from table.
+        /// We will need add one
+        Block header;
+        if (!column_names.empty())
+            header = storage_snapshot->getSampleBlockForColumns(column_names);
+        else
+            header = storage_snapshot->getSampleBlockForColumns({ProtonConsts::RESERVED_EVENT_TIME});
```
yl-lisen, line 441: The external stream has not `_tp_time` , we can use first column

src/Storages/ExternalStream/Kafka/Topic.cpp
yl-lisen: we should support create a topic with specified shards num (i.e. partition num)

src/Storages/ExternalStream/StorageExternalStreamImpl.h
zliang-min: I can't think of any scenarios requrie reporting an error.

src/Storages/ExternalStream/Kafka/Topic.cpp
zliang-min: There is no `rd_kafka_NewTopic_new` function.

`rd_kafka_topic_new` is not a legacy API. Also, when creating a topic, it does not require setting the partition number. There is another resource type called `rd_kafka_topic_parition_t` which is a combination of topic and partition. But we don't need it here.

src/Storages/ExternalStream/Kafka/Kafka.h
zliang-min: The `logger` already contains the stream name.

src/Storages/ExternalStream/Kafka/Kafka.h
zliang-min: Yes, it should release the `producer_topic` before releasing the `producer`, because the `producer_topic` contains a reference to the `producer`.

src/Storages/ExternalStream/Kafka/Kafka.h
zliang-min: This topic is a producer topic, it should be used to produce messages, because it's created with the `producer`. Calling it `getTopic()` will be incorrect.

src/Storages/ExternalStream/Kafka/Kafka.h
zliang-min: Same as the previous one.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: No, the `rd_kafka_topic_t` object is supposed to be shared with all paritions. Each KafkaSource instance use the same topic to call `startConsume` with their own parition number.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: Why the `producer_topic` should be created during initialization? Again, this topic is a producer topic. In librdkafka, a topic is either a consumer topic or a producer topic, there is no "generic topic". So, if a external stream is never used as a sink ( i.e. it's only used to ingest data from a Kafka topic ), it doesn't make sense to initialize the producer topic at all.

And, in order to create a topic, it's required to firstly create a producer or a consumer ( depends what the topic will be used for ). Thus, it's impossible to create a topic without creating a producer/consumer.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: I explained this in previous comments. Yes, this is required, it's the way how librdkafka works.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: Yes, this is how `rd_kafka_consume_start` works. The same `topic` can't be used at the same time in multple select queries. Not just the topic, a `consumer` can't be used in multple select queries at the same time either, that's why we need a consumer pool here.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: Please the check the comment above. `paritition_count` does not change often for sure, but it's not a constant. Check #291

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: Yes, that's a mistake.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: External stream has `_tp_time`, it's the messages' timestamps.

src/Storages/ExternalStream/Kafka/Topic.h
zliang-min: I updated the function name to make it more clear what it does.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: Fixed.

src/Storages/ExternalStream/Kafka/Kafka.cpp
yl-lisen: If we only create an external stream, but don't read or write it.  the topic should be created, and there is not need consumer and producer

src/Storages/ExternalStream/Kafka/Kafka.cpp
yl-lisen: in the ctor of `RdKafka::Topic`, we always call `rd_kafka_topic_new`, it's behavior is unreasonable

yl-lisen: I c, it's okay
```c++
 * Topic handles are refcounted internally and calling rd_kafka_topic_new()
 * again with the same topic name will return the previous topic handle
 * without updating the original handle's configuration.
 * Applications must eventually call rd_kafka_topic_destroy() for each
 * succesfull call to rd_kafka_topic_new() to clear up resources.
```

src/Storages/ExternalStream/Kafka/Kafka.cpp
yl-lisen: that's ok, the topic just is a handler with reference count.

src/Storages/ExternalStream/Kafka/Kafka.h
yl-lisen: In my understanding, each producer and consumer has its own `Topic`, we can move the ownership of `Topic` to `Producer` and `Consumer`

src/Storages/ExternalStream/Kafka/Kafka.h
yl-lisen: In my understanding, each producer and consumer has its own Topic, we can move the ownership of Topic to Producer and Consumer

yl-lisen: In my understanding, each producer and consumer has its own Topic, we can move the ownership of Topic to Producer and Consumer

src/Storages/ExternalStream/Kafka/Producer.cpp
```diff
@@ -0,0 +1,44 @@
...
+/// Producer will take the ownership of `rk_conf`.
+Producer::Producer(rd_kafka_conf_t * rk_conf, UInt64 poll_timeout_ms, Poco::Logger * logger_) : logger(logger_)
+{
+    char errstr[512];
+    rk.reset(rd_kafka_new(RD_KAFKA_PRODUCER, rk_conf, errstr, sizeof(errstr)));
+    if (!rk)
+    {
+        /// librdkafka only take the ownership of `rk_conf` if `rd_kafka_new` succeeds,
+        /// we need to free it otherwise.
+        rd_kafka_conf_destroy(rk_conf);
```
yl-lisen: we should create and destroy the `rk_conf`  in same scope, likes that
```c++
auto * new_rk_conf = rd_kafka_conf_dup(rk_conf);
rk.reset(rd_kafka_new(RD_KAFKA_PRODUCER, new_rk_conf, errstr, sizeof(errstr)));
if (!rk)
{
    rd_kafka_conf_destroy(new_rk_conf);
    ...
}
```

src/Storages/ExternalStream/Kafka/Consumer.cpp
```diff
@@ -0,0 +1,107 @@
...
+Consumer::Consumer(rd_kafka_conf_t * rk_conf, UInt64 poll_timeout_ms, Poco::Logger * logger_) : logger(logger_)
+{
+    char errstr[512];
+    rk.reset(rd_kafka_new(RD_KAFKA_CONSUMER, rk_conf, errstr, sizeof(errstr)));
+    if (!rk)
+    {
+        /// librdkafka only take the ownership of `rk_conf` if `rd_kafka_new` succeeds,
+        /// we need to free it otherwise.
+        rd_kafka_conf_destroy(rk_conf);
+        throw Exception(klog::mapErrorCode(rd_kafka_last_error()), "Failed to create kafka handle: {}", errstr);
```
yl-lisen: we should create and destroy the rk_conf in same scope
same suggests with Producer.

src/Storages/ExternalStream/Kafka/Consumer.cpp
```diff
@@ -0,0 +1,107 @@
...
+    {
+        auto err = rd_kafka_last_error();
+        throw Exception(klog::mapErrorCode(err), "Failed to stop consuming topic={} parition={} error={}", topic.name(), parition, rd_kafka_err2str(err));
+    }
+}
+
+void Consumer::consumeBatch(Topic & topic, Int32 partition, uint32_t count, int32_t timeout_ms, Consumer::Callback callback, ErrorCallback error_callback) const
+{
+    std::unique_ptr<rd_kafka_message_t *, decltype(free) *> rkmessages{
+        static_cast<rd_kafka_message_t **>(malloc(sizeof(*rkmessages) * count)), free}; /// NOLINT(bugprone-sizeof-expression)
```
yl-lisen: same and clear code `malloc(sizeof(rd_kafka_message_t *) * count)`

src/Storages/ExternalStream/Kafka/Consumer.cpp
```diff
@@ -0,0 +1,107 @@
...
+    for (ssize_t idx = 0; idx < res; ++idx)
+    {
+        auto * rkmessage = rkmessages.get()[idx];
+        try
+        {
+            if (rkmessage->err != RD_KAFKA_RESP_ERR_NO_ERROR)
+                error_callback(rkmessage->err);
+            else
+                callback(rkmessage, res, nullptr);
+            }
```
yl-lisen: indent error

src/Storages/ExternalStream/Kafka/Kafka.h
zliang-min: Technically speaking, we can make a Producer/Consumer owns the Topic, but in this case, each `Kafka` has its own Consumers, Producer and Topic, it owns everything. So, for now, I just want to keep things simple, and each class just represents the same concept in the librdkafka library. Later on, when we work on the `Connection` concept, and start reusing connections between external streams, we will revisit these concepts and figure out how to do it properly (mainly the consumer part is a bit tricky).

src/Storages/ExternalStream/Kafka/Kafka.h
zliang-min: Resolving this comment since it's talking the same thing as the previous ones.

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -36,37 +60,52 @@ class Kafka final : public StorageExternalStreamImpl
...
     bool hasCustomShardingExpr() const;
-    klog::KafkaWALSimpleConsumerPtr getConsumer(int32_t fetch_wait_max_ms = 200) const;
+
+    RdKafka::Producer & getProducer();
+    RdKafka::Topic & getProducerTopic();
+
+    RdKafka::ConsumerPool::Entry getConsumer() const
+    {
+        assert(consumer_pool);
+        return consumer_pool->get(/*max_wait_ms=*/1000);
```
yl-lisen, line 76: add more log for consumer_pool (such as current pool size)

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: Also, because this is in the same external stream, the configuration will always be the same. Hence, `return the previous topic handle without updating the original handle's configuration` is totally fine.

src/Storages/ExternalStream/Kafka/Producer.cpp
zliang-min: I was following the librdkafka convention, but yes, I agree your idea is better.

yl-lisen: Another exception-safe code is
```c++
std::unique_ptr<...> new_rk_conf{rd_kafka_conf_dup(rk_conf), rd_kafka_conf_destroy};
rk.reset(rd_kafka_new(RD_KAFKA_PRODUCER, new_rk_conf.get(), errstr, sizeof(errstr)));
if (rk)
    new_rk_conf.release(); /// the rk will take the ownership of `new_rk_conf`
```

src/Storages/ExternalStream/Kafka/Kafka.h
zliang-min: Every time when a new consumer is created, it will be logged and it will log the number of consumer ( consumers are named in the format of `consumer-<number-of-consumer>` ). For now, I don't think we need any more logs.

src/Storages/ExternalStream/Kafka/Producer.cpp
zliang-min: I don't use unique_ptr here because it does not really bring much benefit in this case, if `rd_kafka_conf_dup` fails, it still fails.

yl-lisen (approved):
LGTM

jovezhong:

jovezhong:

PR #629, opened 2024-03-25
Enhancement/issue 613 control max block size for streaming processing

### LINKED ISSUE #613: Control max block size for streaming processing
When we create a materialized view, the inner streaming query may project too large results and materialized into inner/target stream, which may cause the error TOO_LARGE_RECORD

yl-lisen:
This close #613 

- control max block size for single global aggregation

### DIFF
--- a/src/Interpreters/Streaming/Aggregator.cpp
+++ b/src/Interpreters/Streaming/Aggregator.cpp
@@ -1224,7 +1224,8 @@ Block Aggregator::convertOneBucketToBlockImpl(
     if (type == ConvertType::Updates && !method.data.isBucketUpdated(bucket))
         return {};
 
-    Block block = convertToBlockImpl(method, method.data.impls[bucket], arena, data_variants.aggregates_pools, final, method.data.impls[bucket].size(), clear_states, type);
+    constexpr bool return_single_block = true;
+    Block block = convertToBlockImpl<return_single_block>(method, method.data.impls[bucket], arena, data_variants.aggregates_pools, final, method.data.impls[bucket].size(), clear_states, type);
     block.info.bucket_num = static_cast<int>(bucket);
     method.data.resetUpdatedBucket(bucket); /// finalized
     return block;
@@ -1291,8 +1292,9 @@ bool Aggregator::checkLimits(size_t result_size) const
 }
 
 
-template <typename Method, typename Table>
-Block Aggregator::convertToBlockImpl(
+template <bool return_single_block, typename Method, typename Table>
+Aggregator::ConvertToBlockRes<return_single_block>
+Aggregator::convertToBlockImpl(
     Method & method, Table & data, Arena * arena, Arenas & aggregates_pools, bool final, size_t rows, bool clear_states, ConvertType type) const
 {
     if (data.empty())
@@ -1301,16 +1303,16 @@ Block Aggregator::convertToBlockImpl(
         return {finalizeBlock(params, getHeader(final), std::move(out_cols), final, rows)};
     }
 
-    Block res;
+    ConvertToBlockRes<return_single_block> res;
 
     if (final)
     {
-        res = convertToBlockImplFinal<Method>(method, data, arena, aggregates_pools, rows, clear_states, type);
+        res = convertToBlockImplFinal<return_single_block>(method, data, arena, aggregates_pools, rows, clear_states, type);
     }
     else
     {
         assert(type == ConvertType::Normal);
-        res = convertToBlockImplNotFinal(method, data, aggregates_pools, rows);
+        res = convertToBlockImplNotFinal<return_single_block>(method, data, aggregates_pools, rows);
     }
 
     return res;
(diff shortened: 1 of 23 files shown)


src/Processors/Chunk.h
```diff
@@ -343,6 +343,7 @@ class Chunk
 };
 
 using Chunks = std::vector<Chunk>;
+using ChunkList = std::list<Chunk>;
```
chenziliang, line 346: does `std::forward_list` work  ?

src/Interpreters/Streaming/Aggregator.cpp
```diff
@@ -1480,34 +1482,53 @@ Block Aggregator::insertResultsIntoColumns(PaddedPODArray<AggregateDataPtr> & pl
...
+    Method & method, Table & data, Arena * arena, Arenas & aggregates_pools, size_t, bool clear_states, ConvertType type) const
 {
+    /// +1 for nullKeyData, if `data` doesn't have it - not a problem, just some memory for one excessive row will be preallocated
+    const size_t max_block_size = (return_single_block ? data.size() : std::min(params.max_block_size, data.size())) + 1;
     constexpr bool final = true;
-    auto out_cols = prepareOutputBlockColumns(params, aggregate_functions, getHeader(final), aggregates_pools, final, rows);
+    ConvertToBlockRes<return_single_block> res;
 
-    if constexpr (Method::low_cardinality_optimization)
+    std::optional<OutputBlockColumns> out_cols;
```
chenziliang: Use a plain `OutputBlockColumns` is good which we can move away its underlying data with out a explicit reset ?

src/Processors/Transforms/Streaming/AggregatingTransform.h
```diff
@@ -197,9 +197,12 @@ class AggregatingTransform : public IProcessor
 
 protected:
     void emitVersion(Chunk & chunk);
+    void emitVersion(ChunkList & chunks);
     /// return {should_abort, need_finalization} pair
     virtual std::pair<bool, bool> executeOrMergeColumns(Chunk & chunk, size_t num_rows);
-    void setCurrentChunk(Chunk chunk, Chunk retracted_chunk = {});
+    void setAggregatedResult(Chunk chunk, Chunk retracted_chunk = {});
+    void setAggregatedResult(ChunkList chunks);
+    bool hasAggregatedResult() const { return !aggregated_chunks.empty(); }
```
chenziliang: we can make it `noexcept`

src/Processors/Transforms/Streaming/AggregatingTransform.h
```diff
@@ -257,9 +260,7 @@ class AggregatingTransform : public IProcessor
     bool read_current_chunk = false;
 
     /// Aggregated result which is pushed to downstream output
-    Chunk current_chunk_retracted;
-    Chunk current_chunk_aggregated;
-    bool has_input = false;
+    ChunkList aggregated_chunks;
```
chenziliang, line 263: Since now aggregated_chunks is a list, please make sure it won't blow our memory by add more to the list, we shall always drain the aggregated chunks and then pull more from upstream

src/Processors/Transforms/Streaming/GlobalAggregatingTransform.cpp
```diff
@@ -122,21 +122,25 @@ void GlobalAggregatingTransform::finalize(const ChunkContextPtr & chunk_ctx)
...
+            emitVersion(chunks);
 
-        chunk.setChunkContext(chunk_ctx);
-        setCurrentChunk(std::move(chunk));
+        if (chunks.empty()) [[unlikely]]
+            chunks.emplace_back(getOutputs().front().getHeader().getColumns(), 0);
+
+        chunks.back().setChunkContext(chunk_ctx); /// Set chunk context for the last chunk
+
+        setAggregatedResult(std::move(chunks));
```
chenziliang: I may suggest make setAggregatedResult to accept a `&` and swap otherwise there may be 2 caveats
1. There is a move construct happen. It is probably fast but it is still a ctor and move construction and destruction etc
2. If we forget a std::move, it is bad (perf)

src/Processors/Transforms/Streaming/GlobalAggregatingTransformWithSubstream.cpp
```diff
@@ -101,21 +101,25 @@ void GlobalAggregatingTransformWithSubstream::finalize(const SubstreamContextPtr
...
+            emitVersion(chunks, substream_ctx);
 
-        chunk.setChunkContext(chunk_ctx);
-        setCurrentChunk(std::move(chunk));
+        if (chunks.empty()) [[unlikely]]
+            chunks.emplace_back(getOutputs().front().getHeader().getColumns(), 0);
+
+        chunks.back().setChunkContext(chunk_ctx); /// Set chunk context for the last chunk
+
+        setAggregatedResult(std::move(chunks));
```
chenziliang: A ref is clearly a win and far easier to use / hard to misuse

src/Processors/Transforms/Streaming/AggregatingTransform.h
yl-lisen: yes, it only cache current finalized result, and will push it directly to the output port.

src/Processors/Chunk.h
yl-lisen: in somewhere, we need access the back element

PR #635, opened 2024-03-26
feat: new SSL related kafka external stream settings

zliang-min:
PR checklist:
- Did you run ClangFormat ?

Added two new kafka external stream settings for self-signed server keys or server keys that are not signed by well-known CA's.
* `skip_ssl_cert_check` - allows to skip verifying the server (i.e. broker)'s SSL key.
* `ssl_ca_pem` - allows to specify the CA certificate content, so that users do not need to upload the CA cert file.

### DIFF
--- a/src/Storages/ExternalStream/StorageExternalStreamImpl.h
+++ b/src/Storages/ExternalStream/StorageExternalStreamImpl.h
@@ -1,6 +1,8 @@
 #pragma once
 
+#include <Common/logger_useful.h>
 #include <Formats/FormatFactory.h>
+#include <Poco/Util/AbstractConfiguration.h>
 #include <QueryPipeline/Pipe.h>
 #include <Storages/ExternalStream/ExternalStreamSettings.h>
 #include <Storages/ExternalStream/ExternalStreamCounter.h>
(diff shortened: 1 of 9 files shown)


src/Storages/ExternalStream/StorageExternalStreamImpl.h
```diff
@@ -12,7 +14,11 @@ namespace DB
 class StorageExternalStreamImpl : public std::enable_shared_from_this<StorageExternalStreamImpl>
 {
 public:
-    explicit StorageExternalStreamImpl(std::unique_ptr<ExternalStreamSettings> settings_): settings(std::move(settings_)) {
+    explicit StorageExternalStreamImpl(IStorage * storage, std::unique_ptr<ExternalStreamSettings> settings_, const ContextPtr & context)
+        : storage_id(storage->getStorageID())
+        , settings(std::move(settings_))
+        , tmpdir(fs::path(context->getConfigRef().getString("tmp_path", fs::temp_directory_path())) / "external_streams" / toString(storage_id.uuid))
```
chenziliang, line 20: If we fallback to fs::temp_..., is it safe? Since after reboot, everything is gone ?

zliang-min: Yes, this is safe, because the cert pem file will be created again after reboot.

chenziliang (approved):

PR #636, opened 2024-03-27
Enhancement/issue 632 modify default retention policy for streaming store

### LINKED ISSUE #632: Default retention policy for streams: 1d for streaming storage and forever for historical storage
When the user creates a stream in Timeplus UI, we set 1day for the streaming storage (not configurable) and by default 30 days for historical storage. But if it's CSV upload, by default 'Do not remove' for historical storage, in case very old data in the imported data.

When creating stream via SQL in Proton, or in Timeplus SQL Console, we should apply the same default value. Otherwise the user can easily accumulate a lot of data in the stream storage, and don't know how to set the streaming storage retention (I didn't document this on purpose)

yl-lisen:
PR checklist:
- Did you run ClangFormat ?

This close #632

### DIFF
--- a/programs/server/config.yaml
+++ b/programs/server/config.yaml
@@ -990,6 +990,8 @@ settings:
         logstore_replication_factor: 1
         storage_type: "hybrid"
         logstore_codec: none
+        logstore_retention_bytes: -1 # when this threshold reaches, streaming storage delete old data. -1 means no limit
+        logstore_retention_ms: 86400000 # when this threshold reaches, streaming storage delete old data. -1 means no limit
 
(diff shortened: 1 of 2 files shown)


jovezhong (approved):
Looks good to me. If the user run `show create stream abc` for a stream created with just `create stream abc (..)`, will they get those default values?

yl-lisen:
> Looks good to me. If the user run `show create stream abc` for a stream created with just `create stream abc (..)`, will they get those default values?

No, they will still apply old default values , and don't be shown by `show create stream abc`. @jovezhong

PR #643, opened 2024-04-06
Kafka consumer pool waits for all consumers being freed before being destroyed

zliang-min:
PR checklist:
- Did you run ClangFormat ?

It happened in some tests ( hard to reproduce though ) that when shutting down proton, the following error happened

```
2024.04.01 06:48:41.006027 [ 1852 ] {} <Fatal> BaseDaemon: Signal description: Segmentation fault
2024.04.01 06:48:41.006037 [ 1852 ] {} <Fatal> BaseDaemon: Address: 0x38 Access: read. Address not mapped to object.
2024.04.01 06:48:41.006060 [ 1852 ] {} <Fatal> BaseDaemon: Stack trace: 0x00007f27864bffc4 0x0000000010161531 0x000000001d6d8186 0x000000001a79159b 0x000000001a7a10af 0x000000001590dea0 0x0000000015911b7b 0x00000000158fc0f4 0x00000000170dc0d8 0x00000000170dda33 0x00000000170e8d92 0x0000000013fd98b9 0x0000000013fd9857 0x0000000013fd9857 0x0000000013fd9857 0x0000000013fd9857 0x0000000015a9a838 0x0000000015d97019 0x0000000015d61c59 0x0000000015d57c7e 0x00000000101ba4ac 0x00000000101b3b96 0x000000001acee0c6 0x00000000101a5619 0x000000001ad013b3 0x00000000101a3a5a 0x000000000aff9add 0x00007f27862d7083 0x000000000aff902e
2024.04.01 06:48:41.006072 [ 1852 ] {} <Fatal> BaseDaemon: ########################################
2024.04.01 06:48:41.006082 [ 1852 ] {} <Fatal> BaseDaemon: (version 2.0.22, build id: EF025C63D5BE962C16CC2E814DEA21B6819ACC40, git hash: e26764755a5df6b5f03edf4fd0edd688ce45e49b) (from thread 1) (no query) Received signal Segmentation fault (11)
2024.04.01 06:48:41.006099 [ 1852 ] {} <Fatal> BaseDaemon: Address: 0x38 Access: read. Address not mapped to object.
2024.04.01 06:48:41.006107 [ 1852 ] {} <Fatal> BaseDaemon: Stack trace: 0x00007f27864bffc4 0x0000000010161531 0x000000001d6d8186 0x000000001a79159b 0x000000001a7a10af 0x000000001590dea0 0x0000000015911b7b 0x00000000158fc0f4 0x00000000170dc0d8 0x00000000170dda33 0x00000000170e8d92 0x0000000013fd98b9 0x0000000013fd9857 0x0000000013fd9857 0x0000000013fd9857 0x0000000013fd9857 0x0000000015a9a838 0x0000000015d97019 0x0000000015d61c59 0x0000000015d57c7e 0x00000000101ba4ac 0x00000000101b3b96 0x000000001acee0c6 0x00000000101a5619 0x000000001ad013b3 0x00000000101a3a5a 0x000000000aff9add 0x00007f27862d7083 0x000000000aff902e
2024.04.01 06:48:41.006571 [ 1852 ] {} <Fatal> BaseDaemon: 3. pthread_mutex_lock @ 0x00007f27864bffc4 in ?
2024.04.01 06:48:41.009574 [ 1852 ] {} <Fatal> BaseDaemon: 4. pthread_mutex_lock @ 0x0000000010161531 in /usr/bin/proton
2024.04.01 06:48:41.011540 [ 1852 ] {} <Fatal> BaseDaemon: 5. std::__1::mutex::lock() @ 0x000000001d6d8186 in /usr/bin/proton
2024.04.01 06:48:41.011572 [ 1852 ] {} <Fatal> BaseDaemon: 6. std::__1::__shared_ptr_emplace<PoolBase<DB::RdKafka::Consumer>::PoolEntryHelper, std::__1::allocator<PoolBase<DB::RdKafka::Consumer>::PoolEntryHelper>>::__on_zero_shared() @ 0x000000001a79159b in /usr/bin/proton
2024.04.01 06:48:41.011607 [ 1852 ] {} <Fatal> BaseDaemon: 7. DB::KafkaSource::~KafkaSource() @ 0x000000001a7a10af in /usr/bin/proton
2024.04.01 06:48:41.013693 [ 1852 ] {} <Fatal> BaseDaemon: 8. DB::QueryPipeline::~QueryPipeline() @ 0x000000001590dea0 in /usr/bin/proton
2024.04.01 06:48:41.015909 [ 1852 ] {} <Fatal> BaseDaemon: 9. DB::QueryPipeline::reset() @ 0x0000000015911b7b in /usr/bin/proton
2024.04.01 06:48:41.015935 [ 1852 ] {} <Fatal> BaseDaemon: 10. DB::BlockIO::~BlockIO() @ 0x00000000158fc0f4 in /usr/bin/proton
2024.04.01 06:48:41.017730 [ 1852 ] {} <Fatal> BaseDaemon: 11. DB::StorageMaterializedView::PipelineState::~PipelineState() @ 0x00000000170dc0d8 in /usr/bin/proton
2024.04.01 06:48:41.017759 [ 1852 ] {} <Fatal> BaseDaemon: 12. DB::StorageMaterializedView::~StorageMaterializedView() @ 0x00000000170dda33 in /usr/bin/proton
```

And from the logs, it appears that a MV which depens on an kafaka external stream got destroyed after the external stream got destroyed. It's not clear yet how this happened, but this PR tries to make the kafka external stream shutdown a more reliable and graceful process by waiting for all consumers to be freed up before destroying the consumer pool.


src/Common/PoolBase.h
```diff
@@ -167,6 +170,44 @@ class PoolBase : private boost::noncopyable
...
+            {
+                LOG_ERROR(log, "The waiting for pool items to be freed up timed out");
+                return;
+            }
+
+            LOG_INFO(log, "Waiting for in use pool items to be freed up");
+            try
+            {
+                available.wait_for(lock, std::chrono::milliseconds(timeout_ms));
+            } catch (...)
```
chenziliang: we don't need try catch here

src/Common/PoolBase.h
```diff
@@ -167,6 +170,44 @@ class PoolBase : private boost::noncopyable
...
+                    all_freed_up = false;
+                    break;
+                }
+            }
+            if (all_freed_up)
+                return;
+
+            if (timeout_stopwatch.elapsedMilliseconds() > timeout_ms)
+            {
+                LOG_ERROR(log, "The waiting for pool items to be freed up timed out");
```
chenziliang: we shall wait forever, otherwise we still have race ?

src/Storages/ExternalStream/Kafka/ConsumerPool.h
```diff
@@ -52,6 +52,12 @@ class ConsumerPool : public IConsumerPool, private PoolBase<Consumer>
     {
     }
 
+    ~ConsumerPool() override
+    {
+        LOG_INFO(log, "Waiting for all consumers to be freed");
+        waitForNoMoreInUse(/*timeout_ms=*/5000);
```
chenziliang: don't understand why this fix the problem ? It is the pooled entry trying put back to the pool but the pool is gone ? If this is the case, i think we will need forever instead of timed.

The other way to fix this issues is dtor ConsumerPool last which i think is what the existing pooled entry does

chenziliang: And also please use clang-format to format the new code

zliang-min: > The other way to fix this issues is dtor ConsumerPool last which i think is what the existing pooled entry does

Yup, however, in the cases where this failed, the `Kafka` instacne was destroy before the `KafkaSource` instances.

> If this is the case, i think we will need forever instead of timed.
That was my initial thought, but concerned about if it would blocking forever. But I agreed that it's better to keep waiting in this PR first, and if we find out that this could lead to other issues, we revisit it again.

chenziliang: Can we arrange Kafka instance to get destroyed before Kafka source?

src/Storages/ExternalStream/Kafka/ConsumerPool.h
```diff
@@ -52,16 +54,25 @@ class ConsumerPool : public IConsumerPool, private PoolBase<Consumer>
...
+    {
+        if (stopped.test_and_set())
+            return;
+
+        LOG_INFO(log, "Shutting down consumer pool, waiting for all consumers to be freed");
+        waitForNoMoreInUse();
+        LOG_INFO(log, "All consumers are freed");
+    }
+
     Entry get(Int64 max_wait_ms) override
```
chenziliang, line 67: Look like the interface is not well designed. It shall return `std::optional<Entry>`

src/Common/PoolBase.h
```diff
@@ -167,6 +170,29 @@ class PoolBase : private boost::noncopyable
         return items.size();
     }
 
+    /// proton: starts
+    void waitForNoMoreInUse()
```
chenziliang, line 174: One implementation is just move this function to derived class `ConsumerPool.h` since we have overriden shutdown there, from where `waitForNoMoreInUse()` can be invoked.

src/Storages/ExternalStream/Kafka/ConsumerPool.h
zliang-min: Because `Entry` already acts like a `std::optional` (  `Entry` has a `bool is_null()` function ), that's why I didn't use optional here. Otherwise, the code will need to do something like `if (result && !result->is_null())`. Hopefully this makes sense.

src/Common/PoolBase.h
zliang-min: I thought about that too, the reason why I didn't like it is it requires moving `mutex` and `available` from `private` to `protected` in `PoolBase`. I wanted to keep them as internal details of `PoolBase`. But I am open to make the change, let me know what you think.

src/Storages/ExternalStream/Kafka/ConsumerPool.h
chenziliang: it is a smell design in the base, let's skip fixing it for now.

src/Common/PoolBase.h
chenziliang: i c. Let's leave it as it is.

chenziliang (approved):

yl-lisen (commented):
@zliang-min  During proton shutdown, we do not guarantee the shutdown order of streams/external streams/mvs, so if external-stream was stopped or dropped, we should interrupt the consumers that are in use  instead of waiting for the consumer to end.

chenziliang:
If this cause a deadlock, we may need rearrange Kafka / Kafka source etc dependencies tear down order / notify dependents that the dependencies are going to down etc

yl-lisen:
> If this cause a deadlock, we may need rearrange Kafka / Kafka source etc dependencies tear down order / notify dependents that the dependencies are going to down etc

Yes, for teardown, it's a better behavior. And if drop an external stream, we still need interrupt the consumer instead of waiting

zliang-min:
> @zliang-min During proton shutdown, we do not guarantee the shutdown order of streams/external streams/mvs, so if external-stream was stopped or dropped, we should interrupt the consumers that are in use instead of waiting for the consumer to end.

@yl-lisen The order is not guaranteed should be fine ( even it's weird that a dependency shutdown before its dependant ). But if an external stream is shutdown, that means either proton is shutting down, or that external stream gets dropped. And in either case, Mvs depend on that external stream must shutdown too, right? So the change in this PR should be fine.

> If this cause a deadlock,

No, there is no deadlock.

yl-lisen:
@zliang-min  For mv, yes , we have dependencies check to guarantee must drop mv first.  But for a normal select query from external stream, if we drop this external stream, expect current select query is aborted.

yl-lisen:
> If this cause a deadlock,

it's possible to wait forever, when proton teardown, if firstly stop this external stream, then it will wait until KafkaConsumer done, in other words, need wait the dependant mv was stopped.

zliang-min:
> But for a normal select query from external stream, if we drop this external stream, expect current select query is aborted.

That's true, I was not aware of this case. I have a follow-up PR: https://github.com/timeplus-io/proton/pull/645. @yl-lisen @chenziliang please help review.

jovezhong:

jovezhong:

PR #645, opened 2024-04-07
Shutdown Kafka consumers and producer when the external stream is stopped

zliang-min:
PR checklist:
- Did you run ClangFormat ?

A follow-up change of https://github.com/timeplus-io/proton/pull/643. In this change, it makes sure that if an external stream is stopped, all consumers and producer of that stream will be shutdown, and all ad-hoc queries will fail.


src/Storages/ExternalStream/Kafka/Consumer.h
```diff
@@ -33,14 +30,16 @@ class Consumer : boost::noncopyable
...
+
     std::string name() const { return rd_kafka_name(rk.get()); }
 
 private:
     void backgroundPoll(UInt64 poll_timeout_ms) const;
 
     klog::KafkaPtr rk {nullptr, rd_kafka_destroy};
     ThreadPool poller;
-    std::atomic_flag stopped;
+    bool stopped {false};
```
yl-lisen: multiple thread unsafe ?  one thread call `shutdown`, another thread call `consumeBatch`

src/Storages/ExternalStream/Kafka/Producer.h
```diff
@@ -15,21 +15,22 @@ class Producer : boost::noncopyable
...
+
+    bool isStopped() const { return stopped; }
+
 private:
     void backgroundPoll(UInt64 poll_timeout_ms) const;
 
     klog::KafkaPtr rk {nullptr, rd_kafka_destroy};
     ThreadPool poller;
-    std::atomic_flag stopped;
+    bool stopped;
```
yl-lisen: same with above

src/Storages/ExternalStream/Kafka/Consumer.cpp
```diff
@@ -70,6 +71,9 @@ void Consumer::stopConsume(Topic & topic, Int32 parition)
 
 void Consumer::consumeBatch(Topic & topic, Int32 partition, uint32_t count, int32_t timeout_ms, Consumer::Callback callback, ErrorCallback error_callback) const
 {
+    if (stopped)
+        throw Exception(ErrorCodes::NO_AVAILABLE_KAFKA_CONSUMER, "Cannot consume from stopped consummer");
```
chenziliang: The error code is not reflecting a stopped consumer

src/Storages/ExternalStream/Kafka/Consumer.h
chenziliang: It appears there is still race

src/Storages/ExternalStream/Kafka/Consumer.cpp
chenziliang: Look like there is still race since a single bool can't eliminate the race but lower the possibility.

Probably you can check existing NativeLog source reader etc which we carefully maintain the dtor sequence and dependencies life cycle.

src/Storages/ExternalStream/Kafka/Consumer.h
zliang-min: No, an atomic is not needed here, there won't be a race.

So, the `stopped` flag is to for telling the `Consumer` (and the `Producer` in the below case) that it's already stopped. And, when we set `stopped` to `true`, we don't do anything else ( like, we don't detroy anything ), so there is nothing to guard with it, i.e. there is no such logic in our code:
```c++
if (!stopped)
{
    stopped = true;
    /// do something
}
else
{
    /// do something else
}
```
So there is no race.

The purpose is to to let operations on the `Consumer` fail, but it does not have to fail immediately. Take your case as an example, if one thread call `shutdown`, and another call `consumeBatch`, it's fine, it does not care which one wins, even `consumeBatch` wins, nothing bad will happen, and next time when `consumeBatch` is called, it will fail as expected, and then, `KafkaSource` will stopped reading more data. And also, we don't even need to guard `shutdown`, it can be called multiple times, no side effects.

Please let me know if this does not make sense to you.

src/Storages/ExternalStream/Kafka/Producer.h
zliang-min: Explained above.

src/Storages/ExternalStream/Kafka/Consumer.cpp
zliang-min: Fixed the error codes. For race, please check my above comment and let me know if it makes sense.

zliang-min:
@chenziliang @yl-lisen I reverted the change and use `std::atomic_flag` in `Consumer` and `Producer` again, please take another look. Thanks.

src/Storages/ExternalStream/Kafka/Consumer.h
zliang-min: Regardless it does not have any impact in current implementation, I reverted it back, and now use `std::atomic_flag`.

chenziliang: The race i meant it not the stopped data member but at high level..
1. Client call `isStopped()`, returns false, context switch to 2)
2. Consumer then stopped, switch back to 1)
3. Client already checked consumer is not stopped, but it actually is, continue operating on stopped consumer ?

chenziliang: Same for producer and we run into this problem before

jovezhong:

jovezhong:

PR #647, opened 2024-04-08
Feature/issue 634 support left join for versioned kv streams

### LINKED ISSUE #634: Left join for versioned-kv streams
**Use case**

As a user, i would like to do left join for multiple versioned kv streams in an incremental way, and materialize the joined results to a target system like ClickHouse for further processing. The target system is supposed to have upsert (kv semantic). Only in this way, the intermediate joined result can be overridden by a later join and eventually converge. 

Since we don't need necessary do stream aggregation in the same joined query, Proton shall be able to have a more effective way to do the join and emit update instead of changelog

**Additional context**

yl-lisen:
This close #634 

### DIFF
--- a/src/Interpreters/Streaming/HashJoin.cpp
+++ b/src/Interpreters/Streaming/HashJoin.cpp
@@ -438,6 +438,7 @@ joinColumns(std::vector<KeyGetter> && key_getter_vector, const std::vector<const
                 else if constexpr ((jf.is_latest_join) && jf.right)
                 {
                     /// FIXME
+                    throw Exception(ErrorCodes::NOT_IMPLEMENTED, "Not implemented path in joinColumns().");
                 }
                 else if constexpr (jf.is_latest_join && (jf.inner || jf.left))
                 {
@@ -448,6 +449,7 @@ joinColumns(std::vector<KeyGetter> && key_getter_vector, const std::vector<const
                 else /// ANY LEFT
                 {
                     /// FIXME
+                    throw Exception(ErrorCodes::NOT_IMPLEMENTED, "Not implemented path in joinColumns().");
                 }
             }
         }
@@ -466,7 +468,8 @@ joinColumns(std::vector<KeyGetter> && key_getter_vector, const std::vector<const
             }
         }
 
-        if (!right_row_found)
+        /// Not add missing for 'right join left'
+        if (!right_row_found && added_columns.is_left_block)
             addNotFoundRow<jf.add_missing, jf.need_replication>(added_columns, current_offset);
 
         if constexpr (jf.need_replication)
@@ -591,13 +594,18 @@ struct Inserter
         JoinDataBlockList * blocks,
         size_t original_row,
         size_t row,
-        Arena & pool)
+        Arena & pool,
+        IColumn::Filter * new_keys_filter)
     {
         auto emplace_result = key_getter.emplaceKey(map, original_row, pool);
         auto * mapped = &emplace_result.getMapped();
 
         if (emplace_result.isInserted())
+        {
             mapped = new (mapped) typename Map::mapped_type(std::make_unique<HashJoin::RefListMultiple>());
+            if (new_keys_filter)
+                (*new_keys_filter)[original_row] = 1;
+        }
 
         [[maybe_unused]] auto iter = (*mapped)->insert(blocks, row);
     }
@@ -653,7 +661,8 @@ size_t NO_INLINE insertFromBlockImplTypeCase(
     JoinDataBlockList * blocks,
     size_t start_row,
     ConstNullMapPtr null_map,
-    Arena & pool)
+    Arena & pool,
+    IColumn::Filter * new_keys_filter)
 {
     [[maybe_unused]] constexpr bool mapped_one = std::is_same_v<typename Map::mapped_type, typename HashJoin::MapsOne::MappedType>;
     [[maybe_unused]] constexpr bool mapped_multiple
@@ -681,7 +690,8 @@ size_t NO_INLINE insertFromBlockImplTypeCase(
         else if constexpr (mapped_one)
             Inserter<Map, KeyGetter>::insertOne(join, map, key_getter, blocks, i - start_row, i, pool);
         else if constexpr (mapped_multiple)
-            Inserter<Map, KeyGetter>::insertMultiple(join, map, key_getter, blocks, i - start_row, i, pool);
+            /// So far \new_keys_filter is just used in multiple map for bidirectional join
+            Inserter<Map, KeyGetter>::insertMultiple(join, map, key_getter, blocks, i - start_row, i, pool, new_keys_filter);
         else
             Inserter<Map, KeyGetter>::insertAll(join, map, key_getter, &blocks->lastDataBlock(), i - start_row, i, pool);
     }
@@ -698,14 +708,15 @@ size_t insertFromBlockImplType(
     JoinDataBlockList * blocks,
     size_t start_row,
     ConstNullMapPtr null_map,
-    Arena & pool)
+    Arena & pool,
+    IColumn::Filter * new_keys_filter)
 {
     if (null_map)
         return insertFromBlockImplTypeCase<STRICTNESS, KeyGetter, Map, true>(
-            join, map, rows, key_columns, key_sizes, blocks, start_row, null_map, pool);
+            join, map, rows, key_columns, key_sizes, blocks, start_row, null_map, pool, new_keys_filter);
     else
         return insertFromBlockImplTypeCase<STRICTNESS, KeyGetter, Map, false>(
-            join, map, rows, key_columns, key_sizes, blocks, start_row, null_map, pool);
+            join, map, rows, key_columns, key_sizes, blocks, start_row, null_map, pool, new_keys_filter);
 }
 
 template <Strictness STRICTNESS, typename Maps>
@@ -719,7 +730,8 @@ size_t insertFromBlockImpl(
     JoinDataBlockList * blocks,
     size_t start_row,
     ConstNullMapPtr null_map,
-    Arena & pool)
+    Arena & pool,
+    IColumn::Filter * new_keys_filter)
 {
     switch (type)
     {
@@ -728,7 +740,7 @@ size_t insertFromBlockImpl(
         return insertFromBlockImplType< \
             STRICTNESS, \
             typename KeyGetterForType<HashType::TYPE, std::remove_reference_t<decltype(*maps.TYPE)>>::Type>( \
-            join, *maps.TYPE, rows, key_columns, key_sizes, blocks, start_row, null_map, pool); \
+            join, *maps.TYPE, rows, key_columns, key_sizes, blocks, start_row, null_map, pool, new_keys_filter); \
         break;
         APPLY_FOR_HASH_KEY_VARIANTS(M)
 #undef M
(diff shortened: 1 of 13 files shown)


src/Processors/Transforms/Streaming/JoinTransform.cpp
```diff
@@ -278,21 +269,18 @@ inline void JoinTransform::joinBidirectionally(Chunks chunks)
...
-            {
-                /// Don't watermark this block. We can concat retracted / result blocks or use avoid watermarking
-                auto chunk_ctx = ChunkContext::create();
-                chunk_ctx->setConsecutiveDataFlag();
-                output_chunks.emplace_back(retracted_block.getColumns(), retracted_block_rows, nullptr, std::move(chunk_ctx));
-            }
-
-            if (block.rows())
-                output_chunks.emplace_back(block.getColumns(), block.rows());
+            /// Don't watermark this block. We can concat retracted / result blocks or use avoid watermarking
```
chenziliang:
```
/// Don't watermark retracted chunk since we like the retracted chunk and the following result chunk
/// to process in a consecutive way. For example, avoid emitting result right after processing retracted chunk
/// but without processing the following result chunk. This will prevent transitive emit result we don't like 
/// to have usually. 
/// To have retracted chunk / result chunk processed consecutively, we can either concat them into one bigger
/// chunk or use `consecutive` flag which we use here.
```

src/Processors/Transforms/Streaming/JoinTransform.cpp
```diff
@@ -309,10 +297,11 @@ inline void JoinTransform::rangeJoinBidirectionally(Chunks chunks)
         auto block = input_ports_with_data[i].input_port->getHeader().cloneWithColumns(chunks[i].detachColumns());
         auto joined_blocks = std::invoke(join_funcs[i], join.get(), block);
 
-        std::scoped_lock lock(mutex);
-
-        for (size_t j = 0; j < joined_blocks.size(); ++j)
-            output_chunks.emplace_back(joined_blocks[j].getColumns(), joined_blocks[j].rows());
+        for (auto & joined_block : joined_blocks)
+        {
+            if (joined_block.rows())
```
chenziliang:
```
auto rows = joined_block.rows();
if (rows)
    output_chunks.emplace_back(joined_block.getColumns(), rows);
```

chenziliang: Can we add a `detachColumns()` to Block like Chunk as well ?

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -438,6 +438,7 @@ joinColumns(std::vector<KeyGetter> && key_getter_vector, const std::vector<const
                 else if constexpr ((jf.is_latest_join) && jf.right)
```
chenziliang: `else if constexpr (jf.is_latest_join && jf.right)`

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -438,6 +438,7 @@ joinColumns(std::vector<KeyGetter> && key_getter_vector, const std::vector<const
...
                 {
                     /// FIXME
+                    throw Exception(ErrorCodes::NOT_IMPLEMENTED, "Not implemented path in joinColumns().");
```
chenziliang: `right latest join is not supported`

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -448,6 +449,7 @@ joinColumns(std::vector<KeyGetter> && key_getter_vector, const std::vector<const
                 else /// ANY LEFT
                 {
                     /// FIXME
+                    throw Exception(ErrorCodes::NOT_IMPLEMENTED, "Not implemented path in joinColumns().");
```
chenziliang: `Any left join is not supported`

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -466,7 +468,8 @@ joinColumns(std::vector<KeyGetter> && key_getter_vector, const std::vector<const
             }
         }
 
-        if (!right_row_found)
+        /// Not add missing for 'right join left'
```
chenziliang:
```
/// Don't add missing for `right_side join left_side` in bidirectional join case
```

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -466,7 +468,8 @@ joinColumns(std::vector<KeyGetter> && key_getter_vector, const std::vector<const
...
+        if (!right_row_found && added_columns.is_left_block)
```
chenziliang: When input is real left block, then we will need add not found row **only** for `LEFT JOIN` or `FULL JOIN` cases ?

When input is real right block, then we will need add not found row **only** for `RIGHT JOIN` or `FULL JOIN` cases ?

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -904,8 +931,9 @@ void HashJoin::init()
     /// In case when emitChangeLog()
     if (streaming_strictness == Strictness::Range
         && (left_data.join_stream_desc->data_stream_semantic != DataStreamSemantic::Append
-            || right_data.join_stream_desc->data_stream_semantic != DataStreamSemantic::Append))
-        throw Exception(ErrorCodes::NOT_IMPLEMENTED, "Stream range join only support append-only streams");
+            || right_data.join_stream_desc->data_stream_semantic != DataStreamSemantic::Append
+            || streaming_kind != Kind::Inner))
+        throw Exception(ErrorCodes::NOT_IMPLEMENTED, "Only support inner range join append-only streams");
```
chenziliang: `Only inner range join is supported and the left and right stream must be append-only streams in range join`

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1027,77 +1055,71 @@ void HashJoin::postInit(const Block & left_header, const Block & output_header_,
...
-    }
     else if (bidirectional_hash_join)
-    {
         joinBlockWithHashTable<true>(header, right_data.buffered_data->getCurrentHashBlocksPtr());
+    else
+        joinLeftBlock(header);
 
-        /// Fix the emit changelog header, remove left/right delta, then append joined delta
-        if (emitChangeLog())
+    /// Remove internal left/right delta column
```
chenziliang, line 1108: Do we need guard this with `bidirectional_hash_join` since not all cases have delta column ? we will waste computation if we do this for every join case

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1776,9 +1800,22 @@ Block HashJoin::insertLeftBlockAndJoin(Block & left_block)
     /// For bidirectional join, process one stream at a time
     std::scoped_lock lock(left_data.buffered_data->mutex, right_data.buffered_data->mutex);
 
-    /// If this is a retract block which contains _tp_delta with -1 as it value
-    if (auto joined_retracted_block = eraseExistingKeysAndRetractJoin<true>(left_block); joined_retracted_block)
-        return std::move(*joined_retracted_block);
+    if (emitChangeLog())
+    {
+        /// If this is a retract block which contains _tp_delta with -1 as it value
```
chenziliang: `it value` => `its value`

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1776,9 +1800,22 @@ Block HashJoin::insertLeftBlockAndJoin(Block & left_block)
...
+        if (auto joined_retracted_block = eraseExistingKeysAndRetractJoin<true>(left_block); joined_retracted_block)
+            return std::move(*joined_retracted_block);
+    }
+    else
+    {
+        /// Just erase the existing keys from hash map on left side, no need retract join
```
chenziliang: `/// Since we are not emitting changelog, just erase the existing keys from the left side hash map, no need to do retract join`

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1792,11 +1829,88 @@ Block HashJoin::insertRightBlockAndJoin(Block & right_block)
 
     std::scoped_lock lock(left_data.buffered_data->mutex, right_data.buffered_data->mutex);
 
-    /// If this is a retract block which contains _tp_delta with -1 as it value
-    if (auto joined_retracted_block = eraseExistingKeysAndRetractJoin<false>(right_block); joined_retracted_block)
-        return std::move(*joined_retracted_block);
+    if (emitChangeLog())
+    {
+        /// If this is a retract block which contains _tp_delta with -1 as it value
```
chenziliang: Same comments

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1792,11 +1829,88 @@ Block HashJoin::insertRightBlockAndJoin(Block & right_block)
...
+        if (auto joined_retracted_block = eraseExistingKeysAndRetractJoin<false>(right_block); joined_retracted_block)
+            return std::move(*joined_retracted_block);
+    }
+    else
+    {
+        /// Just erase the existing keys from hash map on right side, no need retract join
```
chenziliang: Same comments

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1792,11 +1829,88 @@ Block HashJoin::insertRightBlockAndJoin(Block & right_block)
...
+        if (isRetractBlock(right_block, *right_data.join_stream_desc))
+        {
+            eraseExistingKeys<false>(right_block, right_data);
+            right_block.clear();
+            return {};
+        }
+    }
 
-    doInsertBlock<false>(right_block, right_data.buffered_data->getCurrentHashBlocksPtr());
+    /// For left join, if right data have new join keys, we need delete prev joined result (join key + left_cols + right_cols<empty>),
```
chenziliang: `/// For left join, if right data have new join keys, we need retract prev left joined result (join key + left_cols + empty right_cols),`

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1792,11 +1829,88 @@ Block HashJoin::insertRightBlockAndJoin(Block & right_block)
...
+    /// for example: `lkv left join rkv on lkv.key = rkv.key`
+    ///         (lkv)           (rkv)               (output)
+    /// (s1)    k1,v1,+1                    >>      k1, v1, (rkv.defaults), +1
```
chenziliang: `(s1) -> t1:`, and same for `s2`

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1792,11 +1829,88 @@ Block HashJoin::insertRightBlockAndJoin(Block & right_block)
...
+            eraseExistingKeys<false>(right_block, right_data);
+            right_block.clear();
+            return {};
+        }
+    }
 
-    doInsertBlock<false>(right_block, right_data.buffered_data->getCurrentHashBlocksPtr());
+    /// For left join, if right data have new join keys, we need delete prev joined result (join key + left_cols + right_cols<empty>),
+    /// for example: `lkv left join rkv on lkv.key = rkv.key`
+    ///         (lkv)           (rkv)               (output)
```
chenziliang: `(output)` -> `(left joined)`

src/Interpreters/Streaming/HashJoin.cpp
chenziliang: `(rkv.defaults)` -> `columns with 'empty' values for rkv`

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1792,11 +1829,88 @@ Block HashJoin::insertRightBlockAndJoin(Block & right_block)
...
+    /// (s1)    k1,v1,+1                    >>      k1, v1, (rkv.defaults), +1
+    /// (s2)                    k1,v2,+1    >>      k1, v1, (rkv.defaults), -1
+    ///                                             k1, v1, rkv.k1, rkv.v2, +1
+    /// So, at (s2): we need to do right join left with `k1,(other defaults),-1` and `k1,v2,+1` 
+    std::unique_ptr<IColumn::Filter> new_keys_filter;
+    if (streaming_kind == Kind::Left && emitChangeLog())
```
chenziliang, line 1858: Let's do

```
if (streaming_kind == Kind::Left && emitChangeLog())
{.   
    auto new_keys_filter = std::make_unique<IColumn::Filter>(right_block.rows(), 0);

    doInsertBlock<false>(right_block, right_data.buffered_data->getCurrentHashBlocksPtr(), new_keys_filter.get());

    assert(right_data.join_stream_desc->hasDeltaColumn());
    ...
}
else
{
    doInsertBlock<false>(right_block, right_data.buffered_data->getCurrentHashBlocksPtr());
}
```

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1792,11 +1829,88 @@ Block HashJoin::insertRightBlockAndJoin(Block & right_block)
...
+        new_keys_filter = std::make_unique<IColumn::Filter>(right_block.rows(), 0);
+
+    doInsertBlock<false>(right_block, right_data.buffered_data->getCurrentHashBlocksPtr(), new_keys_filter.get());
+    if (new_keys_filter)
+    {
+        assert(right_data.join_stream_desc->hasDeltaColumn());
+        /// 1) Generate an block with new keys and default values (such as `k1, (other defaults), -1`)
```
chenziliang:
```
/// 1) Generate a block for the new keys 
```

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1792,11 +1829,88 @@ Block HashJoin::insertRightBlockAndJoin(Block & right_block)
...
+        MutableColumns columns;
+        columns.reserve(right_block.columns());
+        /// FIXME, multiple disjuncts OR clause
+        const auto & key_names = table_join->getClauses().front().key_names_right;
+        size_t new_keys_rows = std::accumulate(new_keys_filter->begin(), new_keys_filter->end(), static_cast<size_t>(0));
+        for (auto & col_with_type_name : right_block)
```
chenziliang: `for (const auto & ...`

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1792,11 +1829,88 @@ Block HashJoin::insertRightBlockAndJoin(Block & right_block)
...
+        {
+            if (std::ranges::any_of(key_names, [&](const auto & name) { return name == col_with_type_name.name; }))
```
chenziliang:
```
auto is_key_column = std::ranges::any_of(key_names, [&](const auto & name) { return name == col_with_type_name.name; });
if (is_key_columns)
{
    ...
}
else if (...)
{
    ...
}
else
{
    ....
}
```

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -1792,11 +1829,88 @@ Block HashJoin::insertRightBlockAndJoin(Block & right_block)
...
+        /// `k1, v1, rkv.k1, (rkv.other_defaults), -1`  =>  `k1, v1, (rkv.defaults), -1`
+        if (retracted_block.rows())
+        {
+            for (auto & col_with_type_name : retracted_block)
+            {
+                if (std::ranges::any_of(key_names, [&](const auto & name) { return name == col_with_type_name.name; }))
+                {
+                    auto key_col = IColumn::mutate(col_with_type_name.column->cloneEmpty());
+                    key_col->reserve(right_block.rows());
+                    key_col->insertManyDefaults(right_block.rows());
```
chenziliang: why we use `right_block.rows()` here ?

src/Interpreters/Streaming/HashJoin.cpp
chenziliang: Look like 1) & 2) can be guarded with `new_keys_rows > 0` ? like 

```
size_t new_keys_rows = std::accumulate(new_keys_filter->begin(), new_keys_filter->end(), static_cast<size_t>(0))

if (new_keys_rows > 0)
{
    /// 1) ...
    ...
    /// 2) ...
    ...
}

/// 3 ...
```

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -2200,6 +2323,16 @@ void HashJoin::transformToOutputBlock(Block & joined_block) const
             /// Then remove right delta column, skip this operation since next reordering will ignore it
             // joined_block.erase(*right_delta_column_position_rlj);
         }
+        else
+        {
+            /// output header:  "c2.i, c2.v, i, v"
+            /// Skip this operation since next reordering will ignore it
+            // if (right_delta_column_position_lrj)
```
chenziliang: shall we remove the code and just leave the comments ?

src/Interpreters/Streaming/HashJoin.cpp
yl-lisen: for an enrichment join (such as `append join changelog`), the joined result has `right_kv._tp_delta`, we still need remove it

src/Interpreters/Streaming/HashJoin.cpp
yl-lisen: it's a mistake

src/Interpreters/Streaming/HashJoin.cpp
yl-lisen: Yes, but now we don't support RIGHT JOIN and FULL JOIN, and `jf.add_missing` is only true in LEFT JOIN.
So it's ok now

src/Interpreters/Streaming/HashJoin.cpp
```diff
@@ -466,7 +468,8 @@ joinColumns(std::vector<KeyGetter> && key_getter_vector, const std::vector<const
             }
         }
 
-        if (!right_row_found)
+        /// Don't add missing for `right_side join left_side` in bidirectional join case
+        if (added_columns.is_left_block && !right_row_found)
```
chenziliang, line 472: I was wondering why not do the following which save a function call.

```
if constexpr (jf.add_missing)
{
    if (added_columns.is_left_block && !right_row_found)
        addNotFoundRow<jf.add_missing, jf.need_replication>(added_columns, current_offset);
}
```

PR #651, opened 2024-04-09
`sudo proton install` to install to /usr/local/bin on Mac instead of the locked /usr/bin

jovezhong:
/usr/bin is locked on Mac. Change it to /usr/local/bin. This should also work on Linux

PR checklist:
- Did you run ClangFormat ?


programs/install/Install.cpp
```diff
@@ -214,7 +214,7 @@ int mainInstall(int argc, char ** argv)
         desc.add_options()
             ("help,h", "produce help message")
             ("prefix", po::value<std::string>()->default_value("/"), "prefix for all paths")
-            ("binary-path", po::value<std::string>()->default_value("usr/bin"), "where to install binaries")
+            ("binary-path", po::value<std::string>()->default_value("usr/local/bin"), "where to install binaries")
```
chenziliang, line 218: Do we like to differentiate platform ? For linux /usr/bin ?

jovezhong: We can, but to make things easier, both mac and linux can install the binary to `/usr/local/bin` should be okay. Not sure how ClickHouse recent code addresses this

chenziliang: 1. There are several `usr/bin` in Install.cpp, you may like to check all of them
2. You may like to use `#if defined(OS_DARWIN)` like in this file to differentiate platforms
3. I didn't recall if /usr/local/bin is in system path or not. Probably not, but you may need validate it in all platforms; otherwise proton won't run.

jovezhong: For point 1, I updated 2 other /usr/bin to /usr/local/bin. For #2, leave this to haohang or others. For #3, at least on mac /usr/local/bin is in $PATH.

For next steps
* one solution is to merge it as verify it in 1.5.5. I don't think it will make things worse
* or haohang or other Proton engineers to handle this, better to enhance homebrew script to setup the folders too
* someone can also generate binary with this branch for linux and mac. With the CI, I can only get a docker image for a branch.

Before this is merged, I will start proton server with a config file

chenziliang: if local bin doesn't work for different linux distributions, it will be a regression (we fix one, the other popped up). 

@yokofly could please pick this up for complete fix
1. For linux (prod distribution), let's keep using first class path /usr/bin/
2. For MacOS, use `/usr/local/bin` moving forward.
3. Fix other installation breakage if there is any

Thanks !

yokofly:
edit: personally I do not want this pr to merged. because you can install proton but cannot access server by proton-client is soo weird.

status: locally verify the `proton --install` work.
**but cannot access proton-client**, proton-local work, because there is no user/group for access proton.
I believe this is why clickhouse take poor support. (`create user for access clickhouse` is different in linux)

```
static void createUser(const String & user_name, [[maybe_unused]] const String & group_name)
{
    if (!user_name.empty())
    {
#if defined(OS_DARWIN)
        // TODO: implement.
        throw Exception(ErrorCodes::NOT_IMPLEMENTED, "Unable to create a user in macOS");
#elif defined(OS_FREEBSD)
        std::string command = group_name.empty()
            ? fmt::format("pw useradd -s /bin/false -d /nonexistent -n {}", user_name)
            : fmt::format("pw useradd -s /bin/false -d /nonexistent -g {} -n {}", group_name, user_name);
        fmt::print(" {}\n", command);
        executeScript(command);
#else
        std::string command = group_name.empty()
            ? fmt::format("useradd -r --shell /bin/false --home-dir /nonexistent --user-group {}", user_name)
            : fmt::format("useradd -r --shell /bin/false --home-dir /nonexistent -g {} {}", group_name, user_name);
        fmt::print(" {}\n", command);
        executeScript(command);
#endif
    }
}

```

yokofly:
I hide the previous comment, seems the problem is my homebrew or old proton-server pollution.

after clean the folder, my local is work and valid. but this is really dangerous, because the default proton search the homebrew first.
I do not take too much time to test but my locally usage is a bit ugly:
```
% sudo /usr/local/bin/proton start
% /usr/local/bin/proton-client -h 127.0.1
% /usr/local/bin/proton-client -h 127.0.1  -u proton --password proton@t+
```

jovezhong:
Thanks for new commits, @yokofly  Yes, it will mess up things if the user homebrew install first then proton install again.

I think it's fair to say
* if the user grabs the binary via curl or download from release page, then `sudo proton install` will put this in /usr/local/bin with config file. The folder of initial proton binary won't be in `PATH`
* if the user `brew install proton`, then they don't need to run `sudo proton install` again. Please use homebrew way to setup the files and other things. https://github.com/timeplus-io/homebrew-timeplus/issues/4 is created to track the enhancement in homebrew.

yokofly:
do u mean make the homebrew same behavior with proton install? 

edit: understand,the rest part shall be homebrew, this pr is enough for local install.

jovezhong:
Feel free to merge if you think it's ready.  

Let me list 3 cases
* if the user grabs the binary from release page, then `sudo proton install` will put this in /usr/local/bin with config file. The folder of initial proton binary won't be in PATH
* if the user `brew install proton`, then they don't need to run sudo proton install again. Please use homebrew way to setup the files and other things. https://github.com/timeplus-io/homebrew-timeplus/issues/4 is created to track the enhancement in homebrew.
* if the user use `curl https://install.timeplus.com | sh`, you can call homebrew to install it. Otherwise it's odd to `curl https://install.timeplus.com | sh` then 'sudo proton install'

This PR only addresses 1. You can keep working in https://github.com/timeplus-io/homebrew-timeplus and https://github.com/timeplus-io/install.timeplus.com

jovezhong:
Since @yokofly said "this pr is enough for local install." So I will merge it and put in 1.5.5

PR #666, opened 2024-04-17
Simplified consumer management for Kafka external stream

zliang-min:
PR checklist:
- Did you run ClangFormat ?

Simplified consumer management for Kafka external stream, this also break the dependency between `Kafka` and `KafkaSource` when Proton tries to shutdown the things. It's observed that it's possible a Proton termination could be blocked due to `Kafka` waiting for everything is shutdown. Since the stop order is not guaranteed, such dependency should not exist.

Also, there was a drawback with the current consumer pool implementation, which is that, pooled consumers are never cleaned up until the external stream is shutdown. Which means, if there are a lot concurrent ad-hoc queries happen at some time, there are a lot consumers will get created and pooled, and they might be never used again in a later time.

### DIFF
--- a/programs/server/config.yaml.example
+++ b/programs/server/config.yaml.example
@@ -960,3 +960,12 @@ send_crash_reports:
     endpoint: 'https://6f33034cfe684dd7a3ab9875e57b1c8d@o388870.ingest.sentry.io/5226277'
     # Uncomment to disable proton internal DNS caching.
     # disable_internal_dns_cache: 1
+
+# External stream related configurations
+external_stream:
+    kafka:
+        # Defines how many consumers can be created for each Kafka external stream.
+        # When a SELECT query on a Kafka external stream is executed, a consumer
+        # will be created for that query. That means, no more than `max_consumers_per_stream`
+        # SELECT queries can be executed in parallel on the same Kafka external stream.
+        max_consumers_per_stream: 50
--- a/src/Storages/ExternalStream/Kafka/Consumer.cpp
+++ b/src/Storages/ExternalStream/Kafka/Consumer.cpp
@@ -24,8 +24,6 @@ Consumer::Consumer(const rd_kafka_conf_t & rk_conf, UInt64 poll_timeout_ms_, con
 
     logger = &Poco::Logger::get(fmt::format("{}.{}", logger_name_prefix, name()));
     LOG_INFO(logger, "Created consumer");
-
-    poller.scheduleOrThrowOnError([this] { backgroundPoll(); });
 }
 
 Consumer::~Consumer()
@@ -51,6 +49,9 @@ std::vector<Int64> Consumer::getOffsetsForTimestamps(const std::string & topic, 
 
 void Consumer::startConsume(Topic & topic, Int32 parition, Int64 offset)
 {
+    if (!started.test_and_set())
+        poller.scheduleOrThrowOnError([this] { backgroundPoll(); });
+
     auto res = rd_kafka_consume_start(topic.getHandle(), parition, offset);
     if (res == -1)
     {
--- a/src/Storages/ExternalStream/Kafka/Consumer.cpp
+++ b/src/Storages/ExternalStream/Kafka/Consumer.cpp
@@ -1,21 +1,15 @@
-#include <Common/logger_useful.h>
 #include <Poco/Logger.h>
 #include <Storages/ExternalStream/Kafka/Consumer.h>
 
 namespace DB
 {
 
-namespace ErrorCodes
-{
-extern const int KAFKA_CONSUMER_STOPPED;
-extern const int RESOURCE_NOT_FOUND;
-}
-
 namespace RdKafka
 {
 
 /// Consumer will take the ownership of `rk_conf`.
-Consumer::Consumer(const rd_kafka_conf_t & rk_conf, UInt64 poll_timeout_ms, const String & logger_name_prefix)
+Consumer::Consumer(const rd_kafka_conf_t & rk_conf, UInt64 poll_timeout_ms_, const String & logger_name_prefix)
+: poll_timeout_ms(poll_timeout_ms_)
 {
     char errstr[512];
     auto * conf = rd_kafka_conf_dup(&rk_conf);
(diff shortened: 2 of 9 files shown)


src/Storages/ExternalStream/Kafka/Consumer.cpp
```diff
@@ -51,6 +49,9 @@ std::vector<Int64> Consumer::getOffsetsForTimestamps(const std::string & topic,
 
 void Consumer::startConsume(Topic & topic, Int32 parition, Int64 offset)
 {
+    if (!started.test_and_set())
```
chenziliang, line 52: are we expecting startConsume  being called multiple times ? may be we will need do

```
if (started.test_and_set())
    return;

...
```

src/Storages/ExternalStream/Kafka/Consumer.h
```diff
@@ -30,17 +31,24 @@ class Consumer : boost::noncopyable
...
 private:
-    void backgroundPoll(UInt64 poll_timeout_ms) const;
+    void backgroundPoll() const;
 
+    UInt64 poll_timeout_ms {0};
     klog::KafkaPtr rk {nullptr, rd_kafka_destroy};
     ThreadPool poller;
     Poco::Logger * logger;
 
+    std::atomic_flag started = ATOMIC_FLAG_INIT;
```
chenziliang: Since c++20, we don't need explicitly init std::atomic_flag with `ATOMIC_FLAG_INIT` anymore since it is the default already and proton is c++20 onward

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -497,6 +494,25 @@ Pipe Kafka::read(
     return pipe;
 }
 
+std::shared_ptr<RdKafka::Consumer> Kafka::getConsumer()
+{
+    auto new_consumer = std::make_shared<RdKafka::Consumer>(*conf, poll_timeout_ms, getLoggerName());
+    std::weak_ptr<RdKafka::Consumer> ref = new_consumer;
+
+    std::lock_guard<std::mutex> lock(consumer_mutex);
```
chenziliang: `std::lock_guard lock{consumer_mutex}` a bit simpiler

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -497,6 +494,25 @@ Pipe Kafka::read(
...
+    for (auto & consumer : consumers)
+    {
+        /// if a consumer is gone, replace it with the new one, so that `consumers` won't keep growing forever.
+        if (consumer.expired())
+        {
+            consumer.swap(ref);
+            return new_consumer;
+        }
+    }
+    consumers.push_back(ref);
```
chenziliang: this seems still an unbounded vector since if no consumer is expired, we keep adding more

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -44,13 +43,21 @@ class Kafka final : public StorageExternalStreamImpl
...
-        consumer_pool->shutdown();
-        if (producer)
-            producer->shutdown();
-
         /// Must release all resources here rather than relying on the deconstructor.
         /// Because the `Kafka` instance will not be destroyed immediately when the external stream gets dropped.
-        consumer_pool.reset();
+        {
+            std::lock_guard<std::mutex> lock(consumer_mutex);
+            for (auto & consumer_ptr : consumers)
```
chenziliang: `for (const auto & ...`

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -44,13 +43,21 @@ class Kafka final : public StorageExternalStreamImpl
...
+                /// if the consumer is still running, mark it stopped
+                if (auto consumer = consumer_ptr.lock())
+                    consumer->setStopped();
+
+            std::vector<std::weak_ptr<RdKafka::Consumer>> empty_consumers;
+            consumers.swap(empty_consumers);
```
chenziliang: `consumers.clear()` seems better and simpler or just leave as it is

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -44,13 +43,21 @@ class Kafka final : public StorageExternalStreamImpl
...
+        }
+
+        if (producer)
+            producer->setStopped();
```
chenziliang, line 54: let's add empty line to introduce thought group

src/Storages/ExternalStream/Kafka/Consumer.cpp
zliang-min: Yes, `startConsume` will be called on each partition, thus, if a topic has multiple partitions, it will be called multiple times.

BTW, just in case you are wondering then why not just change it to `startConsume(Topic & topic, Int64 offset)` and consume all paritions on a topic. The reason is that, one can only select data on particular partitions like `select * from my_stream settings shards='1,3'`.

src/Storages/ExternalStream/Kafka/Consumer.h
zliang-min: Will remove.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: Will fix.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: Yes, in theory that's correct. However, ad-hoc queries will eventually stop at some point, only consumers used in MVs will be long-live. So, in practical, the vector won't keep growing forever. Also, we have already had a setting that limits the number of queries can running in proton at the same time, so, I don't feel like that we need another limit on the number of consumers.

Please let me know if you have different ideas.

src/Storages/ExternalStream/Kafka/Kafka.h
zliang-min: Will fix.

src/Storages/ExternalStream/Kafka/Kafka.h
zliang-min: Will fix.

src/Storages/ExternalStream/Kafka/Kafka.cpp
chenziliang: this dependency / assumption seems too deep. I would like to make sure every component has bounded queue since we have seen things goes wild when the assumption breaks. Also we may not like to creating 1000 consumers when the ad-hoc concurrent query is 1000 but this is a different concern.

zliang-min: Added a limit.

programs/server/config.yaml.example
```diff
@@ -960,3 +960,12 @@ send_crash_reports:
...
     # disable_internal_dns_cache: 1
+
+# External stream related configurations
+external_stream:
+    kafka:
+        # Defines how many consumers can be created for each Kafka external stream.
+        # When a SELECT query on a Kafka external stream is executed, a consumer
+        # will be created for that query. That means, no more than `max_consumers_per_stream`
+        # SELECT queries can be executed in parallel on the same Kafka external stream.
+        max_consumers_per_stream: 50
```
yokofly, line 971: we only edit the config.yaml.example in this pr.
do you use this yaml for testing? I think this yaml is confusing, the proton image internally uses `config.yaml` instead `config.yaml.example`.

jovezhong:

jovezhong:

PR #723, opened 2024-05-14
listen on ipv6 address

### LINKED ISSUE #720: make `proton client` just works

Right now, we may have to do `proton client -h 127.0.0.1`

chenziliang:
PR checklist:
- Did you run ClangFormat ?

Fix #720


jovezhong:
Any chance to merge it this week to improve the CLI UX?

yokofly (approved):
before this pr:
we only listen for 0.0.0.0 (ipv4)

after this pr
listen for 0.0.0.0 and [::]

yokofly:
this pr is ready, i verified in my local it works

jovezhong (approved):

programs/server/config.yaml
```diff
@@ -126,7 +126,7 @@ interserver_http_port: 9009
 # listen_host: '::'
 
 # Same for hosts without support for IPv6:
-listen_host: 0.0.0.0
```
chenziliang, line 129: Is listening only on ipv6 working for both ?

yokofly: well, the current code logic is:
if we do not set the `listen_host` it will connect to both 127.1(localhost) and [::1] (ipv6)
if we set the `listen_host` it will use the setting value.
https://github.com/timeplus-io/proton/blob/8258941c79aad156a4a80b6eda8b3ab7eeba05a8/programs/server/Server.cpp#L413-L423

and before this pr:
in the config we listen to 0.0.0.0, which will make the only listen to this addr
after this pr:
i remove the config value, the default binary shall listen to `127.0.0.1` and `[::1]`
for the docker, i use config to overloaded to make it listen to `0.0.0.0` and `[::]` (this is use for port mapping)

yokofly: basically, the pr changed valid because in our code if not set the `listen_host` we default listen to `127.0.0.1` and `[::1]`

chenziliang: My question is if we only listen on ::1, can it support v4 as well ? I don't want to listen on multiple address and internally multiple servers

yokofly: I tried to get it in my local area, but this has not worked.
I listen for ::1, and connect with 127.0.0.1, it got connection refused

PR #732, opened 2024-05-16
bugfix: table function was not working on kafka external stream

### LINKED ISSUE #592: `table` function does not work with external streams
**Describe what's wrong**

Given an kafka external stream `ex_stream`, running `SELECT * FROM table(ex_stream)` never stops. In fact, it behaves exactly the same as `SELECT * FROM ex_stream`, the `table` function is not working in this case.

**How to reproduce**

**Additional context**

zliang-min:
PR checklist:
- Did you run ClangFormat ?

closes #592

### DIFF
--- a/src/Storages/ExternalStream/Kafka/Kafka.cpp
+++ b/src/Storages/ExternalStream/Kafka/Kafka.cpp
@@ -444,19 +444,6 @@ Pipe Kafka::read(
     Pipes pipes;
     pipes.reserve(shards_to_query.size());
 
-    // const auto & settings_ref = context->getSettingsRef();
-    /*auto share_resource_group = (settings_ref.query_resource_group.value == "shared") && (settings_ref.seek_to.value == "latest");
-    if (share_resource_group)
-    {
-        for (Int32 i = 0; i < shards; ++i)
-        {
-            if (!column_names.empty())
-                pipes.emplace_back(source_multiplexers->createChannel(i, column_names, metadata_snapshot, context));
-            else
-                pipes.emplace_back(source_multiplexers->createChannel(i, {RESERVED_APPEND_TIME}, metadata_snapshot, context));
-        }
-    }
-    else*/
     {
         /// For queries like `SELECT count(*) FROM tumble(table, now(), 5s) GROUP BY window_end` don't have required column from table.
         /// We will need add one
@@ -466,10 +453,21 @@ Pipe Kafka::read(
         else
             header = storage_snapshot->getSampleBlockForColumns({ProtonConsts::RESERVED_EVENT_TIME});
 
-        auto offsets = getOffsets(*consumer, query_info.seek_to_info, shards_to_query);
+        auto streaming = query_info.syntax_analyzer_result->streaming;
+
+        auto seek_to_info = query_info.seek_to_info;
+        /// seek_to defaults to 'latest' for streaming. In non-streaming case, 'earliest' is preferred.
+        if (!streaming && seek_to_info->getSeekTo().empty())
+            seek_to_info = std::make_shared<SeekToInfo>("earliest");
+
+        auto offsets = getOffsets(*consumer, seek_to_info, shards_to_query);
         assert(offsets.size() == shards_to_query.size());
 
         for (auto [shard, offset] : std::ranges::views::zip(shards_to_query, offsets))
+        {
+            Int64 high_watermark = NO_WARTERMARK;
+            if (!streaming)
+                high_watermark = topic_ptr->queryWatermarks(shard).high;
             pipes.emplace_back(
                 std::make_shared<KafkaSource>(
                     *this,
@@ -479,9 +477,11 @@ Pipe Kafka::read(
                     topic_ptr,
                     shard,
                     offset,
+                    high_watermark,
                     max_block_size,
                     external_stream_counter,
                     context));
+        }
     }
 
     LOG_INFO(
--- a/src/Storages/ExternalStream/Kafka/Kafka.cpp
+++ b/src/Storages/ExternalStream/Kafka/Kafka.cpp
@@ -467,7 +467,16 @@ Pipe Kafka::read(
         {
             Int64 high_watermark = NO_WARTERMARK;
             if (!streaming)
-                high_watermark = topic_ptr->queryWatermarks(shard).high;
+            {
+                auto marks = topic_ptr->queryWatermarks(shard);
+                LOG_INFO(logger, "watermarks low={} high={}", marks.low, marks.high);
+                high_watermark = marks.high;
+
+                if (offset == nlog::EARLIEST_SN)
+                    offset = marks.low;
+                else if (offset == nlog::LATEST_SN || offset > marks.high)
+                    offset = marks.high;
+            }
             pipes.emplace_back(
                 std::make_shared<KafkaSource>(
                     *this,
--- a/src/Storages/ExternalStream/Kafka/Kafka.cpp
+++ b/src/Storages/ExternalStream/Kafka/Kafka.cpp
@@ -14,6 +14,7 @@
 #include <Storages/SelectQueryInfo.h>
 #include <Common/ProtonCommon.h>
 #include <Common/logger_useful.h>
+#include <Processors/Sources/NullSource.h>
 #include <Parsers/ASTFunction.h>
 
 #include <boost/algorithm/string/classification.hpp>
@@ -396,10 +397,10 @@ std::vector<int32_t> parseShards(const std::string & shards_setting)
     return specified_shards;
 }
 
-std::vector<Int32> getShardsToQuery(const ContextPtr & context, Int32 parition_count)
+std::vector<Int32> getShardsToQuery(const String & shards_exp, Int32 parition_count)
 {
     std::vector<Int32> ret;
-    if (const auto & shards_exp = context->getSettingsRef().shards.value; !shards_exp.empty())
+    if (!shards_exp.empty())
     {
         ret = parseShards(shards_exp);
         /// Make sure they are valid.
@@ -422,6 +423,23 @@ std::vector<Int32> getShardsToQuery(const ContextPtr & context, Int32 parition_c
 }
 }
 
+std::optional<UInt64> Kafka::totalRows(const Settings & settings_ref)
+{
+    auto consumer = getConsumer();
+    auto topic = RdKafka::Topic(*consumer->getHandle(), topicName());
+    auto shards_to_query = getShardsToQuery(settings_ref.shards.value, topic.getPartitionCount());
+    LOG_INFO(logger, "Counting number of messages in partitions [{}] of topic {}", fmt::join(shards_to_query, ","), topicName());
+
+    UInt64 rows = 0;
+    for (auto shard : shards_to_query)
+    {
+            auto marks = topic.queryWatermarks(shard);
+            LOG_INFO(logger, "watermarks on partition {} low={} high={}", shard, marks.low, marks.high);
+            rows += marks.high - marks.low;
+    }
+    return rows;
+}
+
 Pipe Kafka::read(
     const Names & column_names,
     const StorageSnapshotPtr & storage_snapshot,
@@ -437,9 +455,12 @@ Pipe Kafka::read(
 
     /// User can explicitly consume specific kafka partitions by specifying `shards=` setting
     /// `SELECT * FROM kafka_stream SETTINGS shards=0,3`
-    auto shards_to_query = getShardsToQuery(context, topic_ptr->getPartitionCount());
+    auto shards_to_query = getShardsToQuery(context->getSettingsRef().shards.value, topic_ptr->getPartitionCount());
     assert(!shards_to_query.empty());
-    LOG_INFO(logger, "Reading from partitions [{}] of topic {}", fmt::join(shards_to_query, ","), topicName());
+
+    auto streaming = query_info.syntax_analyzer_result->streaming;
+
+    LOG_INFO(logger, "Reading from partitions [{}] of topic {} streaming={}", fmt::join(shards_to_query, ","), topicName(), streaming);
 
     Pipes pipes;
     pipes.reserve(shards_to_query.size());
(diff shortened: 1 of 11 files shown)


src/Storages/Streaming/SeekToInfo.cpp
```diff
@@ -174,9 +174,6 @@ std::pair<int64_t, bool> tryParseAbsoluteTimeSeek(const String & seek_to)
 
 SeekToInfo::SeekToInfo(String seek_to_) : seek_to(std::move(seek_to_))
 {
-    if (seek_to == "latest")
```
chenziliang, line 177: @yl-lisen could you please check if this breaks things ?

src/Storages/ExternalStream/Kafka/Topic.cpp
```diff
@@ -47,6 +47,17 @@ int Topic::getPartitionCount() const
         throw Exception(ErrorCodes::RESOURCE_NOT_FOUND, "Describe topic of {} returned 0 partitions", name());
 }
 
+WatermarkOffsets Topic::queryWatermarks(Int32 partition) const
+{
+    int64_t low, high;
+    auto err = rd_kafka_query_watermark_offsets(&rdk, name().data(), partition, &low, &high, /*timeout_ms=*/1000);
```
chenziliang, line 53: Is this something cached in the client ? If yes, when the offsets get updated at what frequency / interval ?

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
```diff
@@ -90,6 +92,19 @@ Chunk KafkaSource::generate()
     if (isCancelled())
         return {};
 
+    /// For non-streaming queries
+    if (high_watermark != Kafka::NO_WARTERMARK)
```
chenziliang: do we need this check ? I think checking `if (offset == high_watermark)` is good to go

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
```diff
@@ -90,6 +92,19 @@ Chunk KafkaSource::generate()
...
+    {
+        /// The topic is empty.
+        if (offset == high_watermark) [[unlikely]]
+            return {};
+
+        /// All available messages up to the moment when the query was executed have been consumed, no need to read the messages beyond that point.
+        /// `high_watermark` is the next available offset, i.e. the offset that will be assigned to the next message, thus need to use `high_watermark - 1`.
+        if (lastProcessedSN() >= high_watermark - 1)
```
chenziliang: do we need this check ?

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -466,10 +453,30 @@ Pipe Kafka::read(
...
+
+        auto offsets = getOffsets(*consumer, seek_to_info, shards_to_query);
         assert(offsets.size() == shards_to_query.size());
 
         for (auto [shard, offset] : std::ranges::views::zip(shards_to_query, offsets))
+        {
+            Int64 high_watermark = NO_WARTERMARK;
+            if (!streaming)
+            {
+                auto marks = topic_ptr->queryWatermarks(shard);
```
chenziliang, line 490: is this accurate ?

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -466,10 +453,30 @@ Pipe Kafka::read(
...
+                LOG_INFO(logger, "watermarks low={} high={}", marks.low, marks.high);
+                high_watermark = marks.high;
+
+                if (offset == nlog::EARLIEST_SN)
+                    offset = marks.low;
```
chenziliang, line 501: if it is not accurate or it is dynamic, we can't do this ?

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -466,10 +453,30 @@ Pipe Kafka::read(
...
+                else if (offset == nlog::LATEST_SN || offset > marks.high)
+                    offset = marks.high;
```
chenziliang, line 503: Same

chenziliang:
We shall override `virtual std::optional<UInt64> totalRows(const Settings &) const` in external stream storage, which means `SELECT count() table(external_stream)` shall returned very fast without scanning from beginning to end.

src/Storages/ExternalStream/Kafka/Topic.cpp
chenziliang: If not, what's the underlying Kafka API it is using ? Like to confirm it is a good one

zliang-min: No, it's not cached. Everytime it queries the broker for the offsets. This is the underlying implementation: https://github.com/ClickHouse/librdkafka/blob/6f3b483426a8c8ec950e27e446bec175cf8b553f/src/rdkafka_request.c#L552. It looks it calls the kafka offset API.

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
zliang-min: Yes, a special case is `offset` and `high_watermark` both equal to `-1`. I think we should better check `high_watermark != Kafka::NO_WARTERMARK` to make sure that we won't run into some conner cases.

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
zliang-min: Yes. That's the point to use `high_watermark`, so we know when to stop. The idea here is that, when `table()` is used on the external stream, we first find out what's the maximum offset, i.e. the high watermark, currently in the topic partition, and we make sure that we only read messages up to that offset.

For a less busy Kafka topic, without this check could be fine, but for a busy topic, without this check, it will never stop.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: Yes.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: This is safe. For example, even in the case that, when it queries Kafka, it gets the low watermark `10`. And when it actually starts reading messages, the low watermark changes to `15` ( because of retention happens ), Start reading from offset `10` is still valid and the first message it gets will be the message with offset `15`.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: Answered above.

src/Storages/ExternalStream/Kafka/Kafka.cpp
chenziliang: i think it will report error ? no ? otherwise it seems pretty odd. I like to read 10, but you give me 15

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
chenziliang: NO_WARTERMARK can be -99999 to simplify logic since `-1` is used already ?

src/Storages/ExternalStream/Kafka/Topic.cpp
chenziliang: This is the API i was expecting.

src/Storages/ExternalStream/Kafka/Kafka.cpp
chenziliang: and i was thinking do we care about low watermark ? when topic is empty, what is the watermark returned ?

zliang-min: > i think it will report error ? no ?

No, it won't. But what I said was not correct either. If the low offset is `15` and it starts reading from `10`, it won't report error, but it will get stuck. So I reverted that change, and do not set `offset` to `marks.low`.

> and i was thinking do we care about low watermark ? when topic is empty, what is the watermark returned ?

Both the low and high watermarks are the same number. For example, a new topic that has no message return `(0, 0)`.

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
zliang-min: We can't remove the check of `if (offset == high_watermark)` no matter what value `Kafka::NO_WATERMARK` is. Because we need to check `lastProcessedSN() >= high_watermark - 1`.

In the latest commit, I have introduced a new flag `reached_the_end`. And now the check is simpler, and more efficient.

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -422,6 +423,23 @@ std::vector<Int32> getShardsToQuery(const ContextPtr & context, Int32 parition_c
...
+{
+    auto consumer = getConsumer();
+    auto topic = RdKafka::Topic(*consumer->getHandle(), topicName());
+    auto shards_to_query = getShardsToQuery(settings_ref.shards.value, topic.getPartitionCount());
+    LOG_INFO(logger, "Counting number of messages in partitions [{}] of topic {}", fmt::join(shards_to_query, ","), topicName());
+
+    UInt64 rows = 0;
+    for (auto shard : shards_to_query)
+    {
+            auto marks = topic.queryWatermarks(shard);
```
chenziliang, line 436: The indent seems not good

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -422,6 +423,23 @@ std::vector<Int32> getShardsToQuery(const ContextPtr & context, Int32 parition_c
...
+            LOG_INFO(logger, "watermarks on partition {} low={} high={}", shard, marks.low, marks.high);
```
chenziliang, line 437: You might want add topic name in the logging

src/Storages/ExternalStream/Kafka/Kafka.cpp
chenziliang: I thought if you requested 10 but it has only 15, broker reported error and librdkafka probably captured the error but we didn't consume it ?

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -466,10 +474,34 @@ Pipe Kafka::read(
...
+            if (!streaming)
+            {
+                auto marks = topic_ptr->queryWatermarks(shard);
+                LOG_INFO(logger, "watermarks on parition {} low={} high={}", shard, marks.low, marks.high);
+                high_watermark = marks.high;
+
+                if (marks.low == marks.high) /// there are no messages in the topic
+                {
+                    /// As there are no messages, no need to create a KafkaSource instance at all.
+                    pipes.emplace_back(std::make_shared<NullSource>(header));
```
chenziliang, line 497: Can we just `continue` without NullSource ?

src/Storages/ExternalStream/Kafka/Kafka.h
```diff
@@ -19,6 +19,8 @@ class Kafka final : public StorageExternalStreamImpl
 public:
     using ConfPtr = std::unique_ptr<rd_kafka_conf_t, decltype(rd_kafka_conf_destroy) *>;
 
+    static const Int64 NO_WARTERMARK = -1;
```
chenziliang, line 22: Can we use another constant like -13 which is not overloaded ?

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
```diff
@@ -44,6 +45,9 @@ KafkaSource::KafkaSource(
     , topic(topic_)
     , shard(shard_)
     , offset(offset_)
+    , high_watermark(high_watermark_)
+    /// if offset == high_watermark, it means there is no message to read, so it already reaches the end
+    , reached_the_end(high_watermark != Kafka::NO_WARTERMARK && offset == high_watermark)
```
chenziliang, line 50: If we reach to end initially, can we avoid creating such source in the first place ?

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
```diff
@@ -129,12 +136,27 @@ void KafkaSource::readAndProcess()
...
 
     Int64 current_batch_last_sn = -1;
-    auto callback = [&current_batch_last_sn, this](void * rkmessage, size_t total_count, void * data)
-    {
-        auto current_offset = parseMessage(rkmessage, total_count, data);
-        if (current_offset.has_value()) [[likely]]
-            current_batch_last_sn = *current_offset;
-    };
+    RdKafka::Consumer::Callback callback;
+    if (high_watermark == Kafka::NO_WARTERMARK)
```
chenziliang, line 140: Nit: If we set NO_WATERMARK = std::numeric_limits<int64>::max()- 1; seems we can simply things even more.

We don't need branch to have 2 callbacks.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: No, I confirmed there is no error, but it will just keep wait for that message. At least, this is the behavior the version of librdkafka used in proton.

src/Storages/ExternalStream/Kafka/Kafka.cpp
zliang-min: No. If we even remove NullSource, a simple `select * from table(ex_stream)` will report missing column error like this

```
<Error> executeQuery: Code: 8. DB::Exception: Cannot find column `message` in source stream, there are only columns: []. (THERE_IS_NO_COLUMN) (version 1.5.8) (from [::1]:44578) (in query: select * from table(foo_raw);)
```

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
zliang-min: There is already a special handling for this scenario, which is that, when low watermark == high water mark, we use `NullSource`, which is more light-weight, instead of `KafkaSource`. Here, we still have this check in KafkaSource is just to let `KafkaSource` to handle this situation properly, so that logic inside `KafkaSource` is complete.

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
zliang-min: Yes. However, that means it has to check high_wartermark against the current offset for each message, even in streaming mode. Having two callbacks is more code, but for streaming queries, it only check high_watermark for every message batch instead of every message.

jovezhong:

jovezhong:

PR #386, opened 2023-12-06
add external stream test

lizhou1111:
PR checklist:
- Did you run ClangFormat ?
- Did you separate headers to a different section in existing community code base ?
- Did you surround `proton: starts/ends` for new code in existing community code base ?

Please write user-readable short description of the changes:

this closes #311

PR #744, opened 2024-05-24
bugfix: window functions didn't work with Kafka external streams

### LINKED ISSUE #631: Cannot run `tumble(ext_stream, 1d)` on external stream, DB::Exception: This input format is only suitable for streams with a single column of type String but the number of columns is 3. (BAD_ARGUMENTS)
**Describe what's wrong**

Use case, I want to check the number of messages every day in the Kafka topic. I created an external stream for the topic, then want to run the tumble window with time rewind
```sql
SELECT window_start, count() FROM tumble(ext_stream,1d) where _tp_time>earliest_ts() group by window_start
```
Got error

> Received exception from server (version 1.5.4):
> Code: 36. DB::Exception: Received from localhost:8463. DB::Exception: This input format is only suitable for streams with a single column of type String but the number of columns is 3. (BAD_ARGUMENTS)

Using the virtual column in the tumble window has the same error
```sql
SELECT window_start, count() FROM tumble(ext_stream,_tp_time,1d) where _tp_time>earliest_ts() group by window_start
```

One workaround is to create a subquery
```sql
with cte as (select _tp_time,raw from ext_stream settings seek_to='earliest')
                SELECT window_start, count() FROM tumble(cte,1d) group by window_start
```

But this is anti-intuition. 

The issue for the error message
* What is "This input format"? Why applying `tumble` function will create a input format?
* "a single column of type String", should be `string`, not `String`
* "only suitable for streams with a single column of type String but the number of columns is 3" I think the message swapped the object. Could be the opposite: "This input format is only suitable for streams with the number of columns is 3 but we got a single column of type String"

This can be reproduced in latest Proton 1.5.4.

zliang-min:
PR checklist:
- Did you run ClangFormat ?

closes #631

### DIFF
--- a/src/Storages/ExternalStream/Kafka/Kafka.cpp
+++ b/src/Storages/ExternalStream/Kafka/Kafka.cpp
@@ -488,13 +488,12 @@ Pipe Kafka::read(
     pipes.reserve(shards_to_query.size());
 
     {
-        /// For queries like `SELECT count(*) FROM tumble(table, now(), 5s) GROUP BY window_end` don't have required column from table.
-        /// We will need add one
-        Block header;
-        if (!column_names.empty())
-            header = storage_snapshot->getSampleBlockForColumns(column_names);
-        else
-            header = storage_snapshot->getSampleBlockForColumns({ProtonConsts::RESERVED_EVENT_TIME});
+        /// Make _tp_time always available, otherwise window functions won't work with Kafka external streams.
+        Names cns = column_names;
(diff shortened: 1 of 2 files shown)


yokofly:
I noticed we already support external streams in our smoke testing. Could we possibly add a new smoke test for this?
https://github.com/timeplus-io/proton/pull/386/files/

jovezhong:

jovezhong:

PR #761, opened 2024-06-05
feat: break down big messages for Kafka sink

zliang-min:
PR checklist:
- Did you run ClangFormat ?

There is a limit on how big a message can be in Kafka. Currently, with `IRowOutputFormat`s, like `CSV`, `JSONEachRow`, etc. When there are a lot of data in a `Chunk`, it's possible that the generated message can be too big to send to Kafka.

With this PR, the `KafkaSink` will respect `max_insert_block_size` setting, and try to generate Kafka messages not exceed the value of that setting ( if one single row is too big, it still can't break that row into mutiple messages, other clients won't be able to handle such messages ). Which means users can now use `max_insert_block_size` the message size:

```sql
INSERT INTO my_ex_stream ( ... ) SETTINGS max_insert_block_size=1048576 VALUES (...)(...)(...);
```

The above example set the message size to 1MiB.


src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -135,10 +134,16 @@ KafkaSink::KafkaSink(
     , partition_cnt(topic->getPartitionCount())
     , one_message_per_row(kafka.produceOneMessagePerRow())
     , topic_refresh_interval_ms(kafka.topicRefreshIntervalMs())
+    , max_message_batch_size(context->getSettingsRef().max_insert_block_size.value)
```
chenziliang: `max_insert_block_size` is not for bytes, but for row count

chenziliang: Let's meter the row count instead of the bytes which could be more efficient and less computation check ?

chenziliang: The obvious shorting coming will be row count doesn't mean the total bytes, but this is fine for now since max_block_size etc is also handling the block / chunk like this way.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -210,26 +216,70 @@ KafkaSink::KafkaSink(
...
+/// 1) len == total_len. That means the whole Chunk (or the whole row, when one_message_per_row is true) can fit into the buffer.
+///    In this case, we just need to create a rd_kafka_message_t, and push it to current_batch.
+///
+/// 2) len < total_len. That means the Chunk (or a row) is too big to fit into the buffer. In this case, the data between
+///    [pos + len, pos + total_len] will be copied into pending_data so that the bufer can accept more data until it gets
+///    complete data. Then it can create a rd_kafka_message_t.
+///
+/// In this way, we limit the size of one single Kafka message, so that it can avoid hitting the max.message.size.
+/// However, it's still possible that, one single row is still too big and it exceeds that limit. There is nothing we can do about it for now.
+void KafkaSink::addMessageToBatch(char * pos, size_t len, size_t total_len)
```
chenziliang, line 236: if `max_ingest_block_size` means row count in the same kafka message, then i think things may be easier here

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -210,26 +216,70 @@ KafkaSink::KafkaSink(
...
+        pending_size = 0;
+    }
+
+    if (len == total_len)
+        /// Nothing left
+        return;
+
+    /// There are some remaining incomplete data, copy them to pending_data.
+    auto remaining = total_len - len;
+    LOG_INFO(logger, "Writing to pending_data, size={}", remaining);
```
chenziliang: will this be too verbose ?

src/Storages/ExternalStream/Kafka/KafkaSink.h
```diff
@@ -100,6 +101,7 @@ class KafkaSink final : public SinkToStorage
     Int32 partition_cnt {0};
     bool one_message_per_row {false};
     Int32 topic_refresh_interval_ms = 0;
+    UInt64 max_message_batch_size = 0;
```
chenziliang: we probably don't need save this as a data member since it got passed to `WriteBufferFromKafkaSink` and `pending_data` ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Okay, I will add a new setting instead. Using number of rows could be easier to code, however, it's not user-friendly. It's very hard to control the size by number of row.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Yeah, it could be. Will remove it.

src/Storages/ExternalStream/Kafka/KafkaSink.h
zliang-min: Yes, it does not need to be. Will remove it.

src/Storages/ExternalStream/Kafka/WriteBufferFromKafkaSink.h
```diff
@@ -10,25 +10,45 @@ namespace DB
...
 {
 public:
-    explicit WriteBufferFromKafkaSink(std::function<void(char * pos, size_t len)> on_next_, size_t buffer_size = DBMS_DEFAULT_BUFFER_SIZE)
+    explicit WriteBufferFromKafkaSink(
+        std::function<void(char * pos, size_t len, size_t total_len)> on_next_,
+        std::function<void()> after_next_,
+        size_t buffer_size = DBMS_DEFAULT_BUFFER_SIZE)
     : BufferWithOwnMemory<WriteBuffer>(buffer_size)
     , on_next(on_next_)
+    , after_next(after_next_)
```
chenziliang: `after_next(std::move(after_next_))`, same for `on_next(...)`

src/Storages/ExternalStream/Kafka/WriteBufferFromKafkaSink.h
```diff
@@ -10,25 +10,45 @@ namespace DB
...
+
+    void afterNext() override
+    {
+        after_next();
+        marked_offset = 0;
     }
 
-    std::function<void(char *, size_t)> on_next;
+    /// Callback for handling the next chunk of data.
+    /// @param pos the begin position of the data.
```
chenziliang: `\param` this is the convention and IDE likes this as well

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
chenziliang: from my view, honoring  ingest row count is good enough and honoring bytes size introduces copies which is not ideal

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: See above comments.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Even we change the code to limit by row, the logic actually won't change much. Because, the buffer still has its limit, it's still possible that a row or a few rows, after encoded, it will exceed that limit. Thus, we will still need that `markOffset` logic to tell it if it gets a complete data or not. So in fact, before this PR, there is actually a potential bug that it will produce weird messages if the above case happens.

zliang-min: We can do both count and size limits at the same time, just like a normal batching strategy does. With the size limit, it ensures that a message won't be oversized, while with the row limit, one can tune it to reduce the memcpy operations.

chenziliang: I like the row based since it is more intuitive than size

src/Core/Settings.h
```diff
@@ -800,6 +800,7 @@ static constexpr UInt64 operator""_GiB(unsigned long long value)
     M(Int64, record_consume_timeout_ms, 100, "Timeout of consuming record", 0) \
     M(Int64, kafka_fetch_wait_max_ms, 100, "When reading from Kafka, max wait time", 0) \
     M(Int64, kafka_fetch_max_bytes, 52428800, "When reading from Kafka, max bytes to fetch per read", 0) \
+    M(Int64, kafka_max_message_size, 1024 * 1024, "When writing to Kafka, the max size of a message in bytes", 0) \
```
chenziliang: i was thinking if we have general way doing this like the existing `max_ingest_block_size` or `max_block_size`

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: Updated. I also added two metrics to external streams: `MessagesBySize` and `MessagesByRow`, to help users to observe the behavior, so that they can adjust when it needs to.

src/Core/Settings.h
```diff
@@ -800,6 +800,8 @@ static constexpr UInt64 operator""_GiB(unsigned long long value)
     M(Int64, record_consume_timeout_ms, 100, "Timeout of consuming record", 0) \
     M(Int64, kafka_fetch_wait_max_ms, 100, "When reading from Kafka, max wait time", 0) \
     M(Int64, kafka_fetch_max_bytes, 52428800, "When reading from Kafka, max bytes to fetch per read", 0) \
+    M(UInt64, kafka_max_message_size, 1024 * 1024, "When writing to Kafka, the max size of a message in bytes. This setting only works when a row based output format is used. It works together with `kafka_max_message_rows`, when one of these limits is reached, a message will be created.", 0) \
+    M(UInt64, kafka_max_message_rows, DEFAULT_INSERT_BLOCK_SIZE, "When writing to Kafka, the max number of rows of data can be written into one message. This setting only works when a row based output format is used. It works together with `kafka_max_message_size`, when one of these limits is reached, a message will be created.", 0) \
```
chenziliang: Can we just use the max_insert_block_size and treat kafka_max_message_size an additional enhancement ?

zliang-min: We can. The reason why I introduced this `kafka_max_message_rows` is to keep the settings consistent. Compare `settings kafka_max_message_size = 100, kafka_max_message_rows = 2` and `settings kafka_max_message_size = 100, max_insert_block_size = 2`, the former one looks more reasonable to me.

Let me know what you think, I don't have huge concern on changing it to `max_insert_block_size`.

src/Storages/ExternalStream/ExternalStreamCounter.h
```diff
@@ -31,6 +35,8 @@ class ExternalStreamCounter
             {"WriteBytes", write_bytes.load()},
             {"WriteCounts", write_counts.load()},
             {"WriteFailed", write_failed.load()},
+            {"MessagesBySize", messages_by_size.load()},
```
chenziliang, line 38: we probably don't need this in counters since they are dupliated to WriteCounts / WriteBytes ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -155,13 +159,22 @@ KafkaSink::KafkaSink(
...
+            LOG_WARNING(logger, "Data format `{}` is not a row-based format, `one_message_per_row` setting will not be applied", data_format);
     }
     else
     {
-        writer = FormatFactory::instance().getOutputFormat(data_format, *wb, header, context, {}, kafka.getFormatSettings(context));
+        auto max_rows_per_message = context->getSettingsRef().kafka_max_message_rows.value;
+        writer = FormatFactory::instance().getOutputFormat(
+            data_format, *wb, header, context, [this, max_rows_per_message](auto & /*column*/, auto /*row*/) {
+                wb->markOffset();
+                if (++rows_in_current_message % max_rows_per_message == 0)
```
chenziliang: nit : 
```
if (rows_in_current_message >= max_rows_per_message)
{
    rows_in_current_message = 0;
    ...
}
```

chenziliang: Why only `++rows_in_current_message % max_rows_per_message == 0`, we do `wb->next()` ?

zliang-min: This has already been handled in `addMessageToBatch` method. This counter needs to be reset when max size has reached as well.

chenziliang: OutputFormat writer 
  -> writes to wb WriteBufferFromKafkaSink 
  -> for every write call back to WriteCallback which is the lambda function passed to `getOutputFormat`,
      the lamba function checks if we reach the `max_row_per_message` limit, if it does, call `wb->next()` which 
      in turn calls WriteBufferFromKafkaSink::nextImpl to addMessageToBatch(...), 
      look like the logic is broken here since we will need addMessageToBatch until we reach the max limit ?

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
```diff
@@ -210,26 +223,72 @@ KafkaSink::KafkaSink(
...
+}
+
+void KafkaSink::tryCarryOverPendingData()
+{
+    /// If there are pending data and it can be fit into the buffer, then write the data back to the buffer,
+    /// so that we can use the buffer to limit the message size.
+    /// If the pending data are too big, that means we get a over-size row.
+    if (!pending_data.empty() && pending_data.size() < wb->available())
+    {
+        wb->write(pending_data.data(), pending_data.size());
```
chenziliang, line 289: This is something i don't quite like to have for now. 

When data get serialized to WriteBuffer, does the format guarantee row based boundary ? or for every column and every row, the addMessageToBatch will be called ? 

If row based boundary is there, we can keep tracking every row and the size so far, if the current row reaches max row count or max message size, we will need keep the data for the current row around (pending_data) for the next start.

I am thinking : it is possible to avoid this : moving to pending area and then write it back to write buffer. Instead start a new batch and write the current row directly to the new batch.

chenziliang: In general, this is just a potential optimization which we can do in next PR.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
chenziliang: > This has already been handled in `addMessageToBatch` method. This counter needs to be reset when max size has reached as well.

Yes, then let's use `>=` which is a bit more economic than '%`

chenziliang: an ascii flow could be very helpful here btw. but in next pr

src/Storages/ExternalStream/ExternalStreamCounter.h
zliang-min: >  they are dupliated to WriteCounts / WriteBytes

`MessagesBySize` indicates that the number of messages generated because of it reached the max message size limit, and `MessagesByRow`indicates that the number of messages generated because of it reached the max messager row limit. So, in a normal case that when the whole Chunk can fit into a message without hitting those limits, then these two counters won't increase.

So unless you have other concerns, I will keep these. Since we don't print logs for generating messages, these will be useful for monitoring how the sinks work.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: `addMessageToBatch` is where it move the data out from the buffer (i.e. `WriteBufferFromKafkaSink`) to and create the Kafka message.

The output format serializes the rows in the chunk and write it to the buffer. The output format accepts a callback which will be called everything a row is written into the buffer. We use this callback to mark the offset and check if it reaches the row limit. If it reaches the row limit, it flush the buffer (by calling `wb->next()`).

Except reaching the row limit, there are two more cases that the buffer will flush:
1. All rows in the chunk have been written to the buffer.
2. The buffer is full.

So `addMessageToBatch` is the central place to handle all these flush logic. It creates the Kafka messages, and handle in-complete data (pending data), and reset the batch states. I believe there is nothing broken here.

Definitely worth improving the documentation to describe this logic better.

src/Storages/ExternalStream/Kafka/KafkaSink.cpp
zliang-min: I have considered using another buffer, for example, the `pending_data` for batching data for a message. It doesn't look it's a better solution. Because it has to copy the data from buffer to pending_data for every single row.

Another way to optimize this is to swap pending_data and the buffer of `wb`. I will try this in next PR.

src/Core/Settings.h
chenziliang: let's use `max_insert_block_size` because users can use this settings for all scenarios and also for kafka ingest. If max_insert_block_size doesn't work for kafka ingest, users got surprised as well.

chenziliang (approved):

jovezhong:

jovezhong:

PR #762, opened 2024-06-05
[Kafka External Stream] bugfix: repect data format for count optimization

zliang-min:
PR checklist:
- Did you run ClangFormat ?

### DIFF
--- a/src/Storages/ExternalStream/Kafka/Kafka.cpp
+++ b/src/Storages/ExternalStream/Kafka/Kafka.cpp
@@ -259,6 +259,15 @@ Kafka::Kafka(
 
     calculateDataFormat(storage);
 
+    /// For raw format, one_message_per_row default to true. Because if multiple rows are combined together,
+    /// it could be hard to separate them again.
+    if (data_format == "RawBLOB" && !settings->isChanged("one_message_per_row"))
+        settings->set("one_message_per_row", true);
+
+    support_count_optimization = data_format == "RawBLOB" ||
+        data_format == "ProtobufSingle" ||
+        (data_format == "Avro" && (!settings->format_schema.value.empty() || !settings->kafka_schema_registry_url.value.empty()));
+
     cacheVirtualColumnNamesAndTypes();
 
     rd_kafka_conf_set_log_cb(conf.get(), &Kafka::onLog);
@@ -447,6 +456,9 @@ std::vector<Int32> getShardsToQuery(const String & shards_exp, Int32 parition_co
 
 std::optional<UInt64> Kafka::totalRows(const Settings & settings_ref)
 {
+    if (!support_count_optimization)
+        return {};
+
     auto consumer = getConsumer();
     auto topic = RdKafka::Topic(*consumer->getHandle(), topicName());
     auto shards_to_query = getShardsToQuery(settings_ref.shards.value, topic.getPartitionCount());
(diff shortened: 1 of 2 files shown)


zliang-min:
fixes: timeplus-io/proton-enterprise#4829

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -259,6 +259,15 @@ Kafka::Kafka(
 
     calculateDataFormat(storage);
 
+    /// For raw format, one_message_per_row default to true. Because if multiple rows are combined together,
+    /// it could be hard to separate them again.
+    if (data_format == "RawBLOB" && !settings->isChanged("one_message_per_row"))
+        settings->set("one_message_per_row", true);
+
+    support_count_optimization = data_format == "RawBLOB" ||
```
chenziliang, line 268: 1. If `one_message_per_row` is on at external stream storage (definition) layer, we support count optimization ?
2. JSON etc also supports count optimization ?

I was thinking the other direction : what prevents this optimization

1) CSV / TSV with one_message_per_row=false
2) JSONEachRow with one_message_per_row=false

And for others, by default, count optimization shall be fine ?

zliang-min: In fact, we can't do the optimization for most formats, because most formats actually decode multiple rows from one single message. You can take a look at this page for the details: https://clickhouse.com/docs/en/interfaces/formats.

`one_message_per_row` is for sink, it only control how the external stream writes to kafka, but there are many other ways to write to a Kafka topic. So we can't decide if the optimization is doable by checking `one_message_per_row`. Because we don't have control on what messages are in a Kafka topic. For example, one can use benthos (now redpanda connect) to generate Kafka messages that contains multiple JSON object (just like the `JSONEachRow` format does).

chenziliang: if user produce `multiple JSON object` in one kafka message, and if users doesn't setup the JSONEachRow on our end, we will still treat them as one row instead of multiple since we can't decode it anyway.

yes, `one_message_per_row` is for sink and it is tricky but i was thinking when we query it back by using the same external stream (after write to it), we still have the insights it is one message per row, although there may be others writing multiple rows in to the same kafka topic for which we can say it is users setup problem not us.

chenziliang: Or we have a setting which allows user to use the optimization always. Like in kv stream, we have a `use_approximate_count`, when users query like `SELECT count() FROM table(kafka_topic) SETTINGS use_approximate_count=true`, we always fallback to this optimization which could be very helpful for users who really just like to know how many messages there in the topic for now.

chenziliang: For this case, we can choose to use the same setting or choose another one like `count_kafka_message=true` which use this optimization always.

zliang-min: I am thinking maybe we are overthinking for this issue.

There is already a setting called `optimize_trivial_count_query`. I suggest, let's just close that PR w/o merging it. And let the users set `optimize_trivial_count_query` to `false` if they find the optimisation does not work for them?

zliang-min: After thinking about it a bit, I added `one_message_per_row` check and didn't add any new settings. I can't think of any other cases that people want to use message count as the count value for now. Let's keep it simple, and only add new settings when needed.

chenziliang (approved):

jovezhong:

jovezhong:

PR #781, opened 2024-06-25
bugfix: format schema cache is not cleared after the schema is dropped

zliang-min:
PR checklist:
- Did you run ClangFormat ?

In this PR, I have ported a few changes related to `ProtobufSchemas` from CH repo for cleaning up the Protobuf schema cache (please check the `Commits` tag for what are ported). And based on those changes, I added the logic to `DROP FORMAT SCHEMA` and `CREATE OR REPLACE FORMAT SCHEMA` to clear theProtobuf schema cache.

Here is a use case that is blocked by this bug:

1. Create a protobuf schema, for example
    ```sql
    CREATE FORMAT SCHEMA foo as '
    syntax = "proto3";
    
    message Foo {
      string a = 1;
      int64 b = 2;
    }
    ' TYPE Protobuf;
    ```
2. Use the schema with a Kafka external stream:
    ```sql
    CREATE EXTERNAL STREAM foo (
    a string,
    b int64
    ) SETTINGS type='kafka',...,data_format='ProtobufSingle',format_schema='foo:Foo';
    
    SELECT * FROM foo;
    ```
    This works fine.
3. Replace the schema with a new one ( like the schema is evolved ):
    ```sql
    CREATE FORMAT SCHEMA foo as '
    syntax = "proto3";
    
    message Foo {
      string a = 1;
      int64 b = 2;
      string c = 3;
    }
    ' TYPE Protobuf;
    ```
    In this example, we added a new field `c`.
4. Use the new schema with an evolved external stream
    ```sql
    CREATE EXTERNAL STREAM foo (
    a string,
    b int64,
    c string
    ) SETTINGS type='kafka',...,data_format='ProtobufSingle',format_schema='foo:Foo';
    
    SELECT * FROM foo;
    ```
    This time, the values of column `c` will always be empty, because the schema is cached, and even after we re-create it, it still uses the old cached schema.


src/Formats/FormatSchemaFactory.h
```diff
@@ -22,7 +20,7 @@ class FormatSchemaFactory final : private boost::noncopyable
 
     void registerSchema(const String & schema_name, const String & format, std::string_view schema_body, ExistsOP exists_op, ContextPtr & context);
 
-    void unregisterSchema(const String & schema_name, const String & format, bool throw_if_not_exists, const ContextPtr & context);
+    void unregisterSchema(const String & schema_name, const String & format, bool throw_if_not_exists, ContextPtr & context);
```
chenziliang: it is interesting we pass `ContextPtr &` around, shall it be `const ContextPtr & ` ?

chenziliang:
LGTM in general

src/Formats/FormatSchemaFactory.h
zliang-min: Fixed.

src/Formats/FormatFactory.cpp
```diff
@@ -582,16 +581,15 @@ bool FormatFactory::checkIfFormatHasExternalSchemaWriter(const String & name)
 
 ExternalSchemaWriterPtr FormatFactory::getExternalSchemaWriter(
     const String & name,
-    std::string_view body,
-    ContextPtr & context,
-    const std::optional<FormatSettings> & _format_settings) const
+    const ContextPtr & context,
+    std::optional<FormatSettings> _format_settings) const
```
chenziliang: `format_settings_`

src/Formats/ProtobufSchemas.cpp
```diff
@@ -44,30 +49,41 @@ class ProtobufSchemas::ImporterWithSourceTree : public google::protobuf::compile
...
-        // isn't executed.
+        if (error)
+        {
+            auto info = error.value();
+            error.reset();
+            throw Exception(
+                ErrorCodes::CANNOT_PARSE_PROTOBUF_SCHEMA,
+                "Cannot parse '{}' file, found an error at line {}, column {}, {}",
+                info.filename,
+                std::to_string(info.line),
```
chenziliang: remove `std::to_string`

chenziliang (approved):

jovezhong:

jovezhong:

PR #788, opened 2024-06-30
Update docker-compose.yml version pin typo fix

mathislindner:
I fixed a typo where there was a :: instead of a :


jovezhong (approved):
Thanks for fixing this. Certainly will merge it. I will also check other docker compose files

PR #831, opened 2024-09-02
fix the problem of multishard incorrect result on distinct

### LINKED ISSUE #830: Remain problem existed in multi shards distinct
**Describe what's wrong**
```
client@1:
create stream if not exists test13_multishard_7(id int, val int) settings shards=3, sharding_expr='to_int(id)'
client@1:
select sum_distinct(val), sum(val) from test13_multishard_7 emit periodic 1ms
client@2:
insert into test13_multishard_7(id, val) values(1, 30), (2, 30)

```

```
c1 :) select sum_distinct(val), sum(val) from v2

SELECT
  sum_distinct(val), sum(val)
FROM
  test13_multishard_7
EMIT STREAM PERIODIC 1ms

Query id: 01f530a6-e045-481f-b7d2-8f8518e5e886

┌─sum_distinct(val)─┬─sum(val)─┐
│                 0 │       60 │
└───────────────────┴──────────┘

```

Jasmine-ge:
close #830

### DIFF
--- a/src/AggregateFunctions/Streaming/AggregateFunctionDistinct.h
+++ b/src/AggregateFunctions/Streaming/AggregateFunctionDistinct.h
@@ -107,9 +107,21 @@ struct AggregateFunctionDistinctSingleNumericData
         for (auto it = extra_data_since_last_finalize.begin(); it != extra_data_since_last_finalize.end();)
         {
             if (rhs.set.find(*it) != rhs.set.end())
-                it = extra_data_since_last_finalize.erase(it);
+            {
+                auto found = std::find(rhs.extra_data_since_last_finalize.begin(), rhs.extra_data_since_last_finalize.end(), *it);
+                if (found == rhs.extra_data_since_last_finalize.end())
+                {
+                    it = extra_data_since_last_finalize.erase(it);
+                }
+                else
+                {
+                    ++it;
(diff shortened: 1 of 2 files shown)


src/AggregateFunctions/Streaming/AggregateFunctionDistinct.h
```diff
@@ -107,9 +107,21 @@ struct AggregateFunctionDistinctSingleNumericData
         for (auto it = extra_data_since_last_finalize.begin(); it != extra_data_since_last_finalize.end();)
         {
             if (rhs.set.find(*it) != rhs.set.end())
-                it = extra_data_since_last_finalize.erase(it);
+            {
+                auto found = std::find(rhs.extra_data_since_last_finalize.begin(), rhs.extra_data_since_last_finalize.end(), *it);
```
yl-lisen:
```c++
bool is_new_data = std::find(rhs.extra_data_since_last_finalize.begin(), rhs.extra_data_since_last_finalize.end(), *it) != rhs.extra_data_since_last_finalize.end();
if (is_new_data)
    ++it;
else
    it = extra_data_since_last_finalize.erase(it);
```

yl-lisen (approved):

Issue #276, opened 2023-11-14
Support for Apache Arrow and ADBC (Arrow Database Connectivity)


zeroshade:
From what I can tell, if Proton is based on / extends ClickHouse, it would make a lot of sense to add support for retrieving data in the [Apache Arrow](https://arrow.apache.org) format as ClickHouse already supports using Arrow for performance reasons. This could also extend to adding support for Arrow Flight SQL (which would provide ADBC support for free).

Alongside generalized Arrow support, and the recent announcements of ODBC and JDBC drivers, it would be extremely beneficial and performant to also add an [ADBC](https://arrow.apache.org/adbc/) driver. To be fair, as I mentioned above, if Arrow FlightSQL support is provided, then the existing ADBC driver for Arrow Flight SQL could just be used directly (and would also allow using the existing Arrow FlightSQL JDBC/ODBC drivers.

The generalized reason for supporting ADBC is that it allows bypassing the transposition of column-oriented data to and from row-orientations just to pass across ODBC/JDBC interfaces (which are inherently row-oriented).

This could be provided either by contributing a driver to the https://github.com/apache/arrow-adbc repository, or by simply maintaining a driver here (if it is decided to not go the route of using Arrow FlightSQL). I can absolutely assist with any work involved here, though don't have the bandwidth to do it myself. I just thought it would be a useful suggestion to make here.

chenziliang:
Thanks for the input. May I ask if you are using ADBC etc in your current project / prod env ? I am trying to understand how mature the ADBC framework etc is and how  does the ecosystem look like.

zeroshade:
@chenziliang I'm one of the developers of ADBC :smile: and wrote the Snowflake ADBC driver https://medium.com/snowflake/arrow-database-connectivity-adbc-support-for-snowflake-7bfb3a2d9074

To my knowledge, dbt is adding support for ADBC and there are a few microsoft employees who are frequently contributing to the C# ADBC libraries. Currently we have drivers for sqlite, postgres, Arrow Flight SQL (which also covers Dremio), Snowflake, and DuckDB added an implementation of the ADBC interface to it's shared object. I'll be soon working on a BigQuery driver myself.

I'd be happy to answer any questions you have regarding ADBC and Arrow in general. The ecosystem is continuing to grow rapidly (consider that the underlying memory representation for Polars is Arrow, along with pandas), and we're picking up support for more systems as we go.

ranjanmangla1:
[...]

PR #856, opened 2024-11-19
Add ADBC driver support for Arrow Flight SQL

vishwamartur:
Related to #276

Add support for ADBC (Arrow Database Connectivity) driver for Arrow Flight SQL.

* **ADBC Driver Implementation**
  - Add `src/Processors/Formats/Impl/ADBCDriver.cpp` to implement the ADBC driver for Arrow Flight SQL.
  - Add `src/Processors/Formats/Impl/ADBCDriver.h` to declare the ADBC driver class and its methods.

* **Configuration**
  - Modify `src/configure_config.cmake` to include ADBC driver support and necessary libraries.

* **Documentation**
  - Update `README.md` to include information about ADBC driver support, examples, and usage instructions.

* **Testing**
  - Add `tests/ADBCDriverTest.cpp` to implement tests for the ADBC driver, including connection and query execution tests.

/claim #276

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,31 @@ Integrations with other systems:
 * Homebrew https://github.com/timeplus-io/homebrew-timeplus
 * dbt https://github.com/timeplus-io/dbt-proton
 
+* ADBC Driver for Arrow Flight SQL: Timeplus Proton now supports ADBC (Arrow Database Connectivity) driver for Arrow Flight SQL. This allows for efficient data retrieval and manipulation using the Arrow format. The ADBC driver can be used to connect to Arrow Flight SQL servers and execute queries.
+
+### Example usage of ADBC driver:
+
+```cpp
+#include <iostream>
+#include "ADBCDriver.h"
+
+int main() {
+    try {
+        DB::ADBCDriver driver("grpc://localhost:8815");
+        auto table = driver.ExecuteQuery("SELECT * FROM my_table");
+
+        // Process the retrieved data
+        for (const auto& column : table->columns()) {
+            std::cout << column->ToString() << std::endl;
+        }
+    } catch (const std::exception& e) {
+        std::cerr << "Error: " << e.what() << std::endl;
+    }
+
+    return 0;
+}
+```
+
 ## Documentation
 
 We publish full documentation for Timeplus Proton at [docs.timeplus.com](https://docs.timeplus.com/proton) alongside documentation for Timeplus Enterprise.
--- a/src/Processors/Formats/Impl/ADBCDriver.cpp
+++ b/src/Processors/Formats/Impl/ADBCDriver.cpp
@@ -1,0 +1,56 @@
+#include "ADBCDriver.h"
+#include <adbc.h>
+#include <arrow/flight/sql/client.h>
+#include <arrow/flight/sql/types.h>
+#include <arrow/flight/client.h>
+#include <arrow/flight/types.h>
+#include <arrow/io/memory.h>
+#include <arrow/result.h>
+#include <arrow/status.h>
+#include <arrow/table.h>
+#include <arrow/util/logging.h>
+#include <iostream>
+#include <memory>
+#include <string>
+
+namespace DB
+{
+
+ADBCDriver::ADBCDriver(const std::string & uri)
+{
+    arrow::flight::Location location;
+    arrow::Status status = arrow::flight::Location::Parse(uri, &location);
+    if (!status.ok())
+    {
+        throw std::runtime_error("Failed to parse URI: " + status.ToString());
+    }
+
+    arrow::flight::FlightClientOptions options;
+    status = arrow::flight::FlightClient::Connect(location, options, &client_);
+    if (!status.ok())
+    {
+        throw std::runtime_error("Failed to connect to Flight server: " + status.ToString());
+    }
+}
+
+std::shared_ptr<arrow::Table> ADBCDriver::ExecuteQuery(const std::string & query)
+{
+    arrow::flight::sql::FlightSqlClient sql_client(client_);
+    arrow::flight::sql::StatementQueryResult result;
+    arrow::Status status = sql_client.Execute(query, &result);
+    if (!status.ok())
+    {
+        throw std::runtime_error("Failed to execute query: " + status.ToString());
+    }
+
+    std::shared_ptr<arrow::Table> table;
+    status = result.reader->ReadAll(&table);
+    if (!status.ok())
+    {
+        throw std::runtime_error("Failed to read query result: " + status.ToString());
+    }
+
+    return table;
+}
+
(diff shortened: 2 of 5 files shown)


jovezhong:
Thanks for the PR. We will be reviewing it shortly.

jovezhong:
Hi @vishwamartur,

Thanks for the PR. I am checking with our engineering team to see who will be the best person to look into the implementation details. What I am expecting
* Able to send SQL query/DDL to Timeplus via all supported ADBC languages. According to https://arrow.apache.org/adbc/current/index.html, they are C, C++, Go, Java, Python, R. You updated the README for a C++ example, with grpc to Arrow Flight SQL. Can you also work on the ADBC drivers in other langauges?
* For large resultset, the ADBC should perform same or better than JDBC(row based)
* It'll be great to support streaming SQL with ADBC

I will arrange some blog/video around the ADBC/Arrow support when the PR is merged.

Hope it makes sense, and feel free to let us know your thoughts.

vishwamartur:
Hi @jovezhong,

Thank you for the detailed feedback and suggestions!

To start, we’d like to focus on fully implementing and stabilizing the ADBC driver support in C++. Once the C++ implementation is complete and meets the required performance and functionality benchmarks (e.g., large result set handling, streaming SQL), we can then plan to extend support to other languages like Go, Java, Python, and R.

This phased approach will allow us to ensure a solid foundation before expanding to other ecosystems. Let me know if this sounds good, or if you have any immediate priorities that require parallel development in other languages.

Thanks!
@vishwamartur

jovezhong:
Sounds good. Let's have the C++ driver has the 1st feature-complete ADBC driver, then expand to more languages. From high priority to lower: C++ > Java > Python > Go. You don't need to work on R adapter.  Ideally we contribute the ADBC driver for Timeplus, similar to https://arrow.apache.org/adbc/current/driver/postgresql.html

zeroshade:
Looking at this, this doesn't appear to actually have much to do with ADBC in anything but name. Does Timeplus already support Arrow FlightSQL? If so, then there's nothing that needs to be done as all of the ADBC bindings would be able to use the FlightSQL driver to connect query data from any one of multiple languages (Go, C++, C, Python, R, Rust, Java, etc.)

If Timeplus doesn't already support FlightSQL, then you need to implement the ADBC C interface to create a driver, ideally as a shared object library that can be separately distributed as a client rather than built into Timeplus directly. I can help with that if needed.

jovezhong:
Thanks Matt for the comment. Today in Timeplus Proton server we don't have FlightSQL built-in. I leave more discussions between you and @vishwamartur 

To be clear, we want ADBC support more than FlightSQL.

zeroshade:
I just want to clarify: @vishwamartur is the goal here to have an ADBC driver to connect to Time plus with? Or for Time plus to connect to other sources via ADBC? That will affect what is expected to be implemented here.

zeroshade:
@jovezhong i just to be clear, if Timeplus exposes a Flight SQL server for connectivity, you would get ADBC support for free via the flight SQL ADBC (and ODBC/JDBC) driver that already exists. 

That said, I believe you already are built on ClickHouse, so it shouldn't be too difficult to create an ADBC driver which can use the ClickHouse protocol for connecting and retrieving Arrow formatted data, right?

vishwamartur:
Hi @zeroshade,  

Thanks for the clarification! The goal is to create an ADBC driver for clients to connect **to Timeplus**. Leveraging the ClickHouse protocol to retrieve Arrow-formatted data makes sense, given our architecture.  

If you have any specific suggestions for implementing the ADBC C interface or designing the driver as a shared library, I’d greatly appreciate it.  

Looking forward to your thoughts!  

Best,  
Vishwa

zliang-min:
@vishwamartur I might have missed something, but looking at the PR, I don't see how this can let someone create a ADBC driver to connect to timeplus proton. Could you help me to understand how this works, please?

zliang-min:
> @jovezhong i just to be clear, if Timeplus exposes a Flight SQL server for connectivity, you would get ADBC support for free via the flight SQL ADBC (and ODBC/JDBC) driver that already exists.
> 
> That said, I believe you already are built on ClickHouse, so it shouldn't be too difficult to create an ADBC driver which can use the ClickHouse protocol for connecting and retrieving Arrow formatted data, right?

@zeroshade yes, proton also has the arrow format support as ClickHouse does, but there are gaps as the implementations are not up-to-date with the ClickHouse repo at the moment. This might or might not have impact on implementing an ADBC driver ( I don't now much about implementing an ADBC driver ). I don't know if ADBC interface supports streaming already, since proton is a streaming data engine, this is one thing to pay attention to when implementing a database driver for it.

vishwamartur:
image


@zeroshade, could you please suggest any changes?

@zliang-min, if I’m mistaken, I would appreciate your guidance and suggestions for improvements. I’ll do my best to implement them.

zliang-min:
@vishwamartur to achieve the goal of being able to connect to timeplus proton via an ADBC driver, there are two options:
* create an timeplus proton specific ADBC driver that understands how to talk to proton to execute queries. Such a driver can live in its own repo, and it does not have to be implemented in C++ ( for example, you can use Go and [the proton go driver](https://github.com/timeplus-io/proton-go-driver) to implement a GO ADBC driver for proton ).
* or, like @zeroshade mentioned, you can implement a Flight SQL server in proton, so that people can use any existing ADBC driver to connect to proton and run queries.

The second option allows the maximum availability and makes it easier to integrate with the existing ecosystem. The first option is probably easier, but it has big limitations ( it limits what languages can be used, and it's hard to utilize what are already there in the ecosystem ).

Hopefully this makes sense.

zeroshade:
> The second option allows the maximum availability and makes it easier to integrate with the existing ecosystem. The first option is probably easier, but it has big limitations ( it limits what languages can be used, and it's hard to utilize what are already there in the ecosystem ).

It actually doesn't limit the languages as much as you'd expect. For example, the current ADBC FlightSQL driver is implemented in Go and distributed as a C shared object that can be loaded by ADBC driver managers. If you implement the Go ADBC Interface, then it's a simple case to use the [existing SDK](https://github.com/apache/arrow-adbc/tree/main/go/adbc/pkg) to create a distributable driver that can be easily loaded by any ADBC driver manager.

> @zeroshade, could you please suggest any changes?

I would argue that `ADBC Driver` belongs in the same box as `SDK, JDBC/ODBC` and `Data/BI Connectors`. An ADBC driver is just another driver, similar in concept to a JDBC or ODBC driver (but columnar and Arrow-native instead of row-oriented).

vishwamartur:
I’ve made the changes in [this pull request](https://github.com/timeplus-io/proton/pull/865). Could you please review them and share your suggestions? I’m happy to make any necessary updates.

PR #898, opened 2025-02-12
Support group concat function

### LINKED ISSUE #889: Support group_concat function
Reference https://clickhouse.com/docs/en/sql-reference/aggregate-functions/reference/groupconcat

The use case is to follow the ClickHouse docs [script](https://github.com/ClickHouse/clickhouse-docs/blob/main/scripts/settings/autogenerate-settings.sh) to generate the markdown based on settings source code.
```sql
SELECT prefix || (SELECT groupConcat(*) FROM main_content)
INTO OUTFILE 'docs/en/operations/settings/settings.md' TRUNCATE FORMAT LineAsString
```
There is only 1 column in CTE main_content, so it uses `groupConcat(*) `

I can use group_array(col) to create an array but there is no function to turn this array of string to a single string. concat function takes multiple string params, but not an array.

Jasmine-ge:
Calculates a concatenated string from a group of strings, optionally separated by a delimiter, and optionally limited by a maximum number of elements.

**Syntax**
group_concat (expression [, delimiter [, limit ] ])

**Arguments**

- expression — The expression or column name that outputs strings to be concatenated.
- delimiter — A string that will be used to separate concatenated values. This parameter is optional and defaults to an empty string or delimiter from parameters if not specified.

**Parameters**

- delimiter — A string that will be used to separate concatenated values. This parameter is optional and defaults to an empty string if not specified.
- limit — A positive integer specifying the maximum number of elements to concatenate. If more elements are present, excess elements are ignored. This parameter is optional.

**Returned value**

- Returns a string consisting of the concatenated values of the column or expression. If the group has no elements or only null elements, and the function does not specify a handling for only null values, the result is a nullable string with a null value.

Port CH PRs and update the obvious refactor (utils, format, etc):
https://github.com/ClickHouse/ClickHouse/pull/65009
https://github.com/ClickHouse/ClickHouse/pull/69455
https://github.com/ClickHouse/ClickHouse/pull/72540

close #889


src/Interpreters/ExpressionAnalyzer.cpp
```diff
@@ -277,6 +277,30 @@ void tryTranslateToParametricAggregateFunction(
...
         types.pop_back();
     }
+    else if (lower_name == "group_concat")
+    {
+        if (arguments.size() > 3 || arguments.size() < 1)
+        {
+            throw Exception(ErrorCodes::NUMBER_OF_ARGUMENTS_DOESNT_MATCH,
+                "Incorrect number of parameters for aggregate function {}, should be 0, 1 or 2, got: {}", node->name, parameters.size());
+        }
+        else if (arguments.size() == 3)
```
yl-lisen:
```suggestion
        if (arguments.size() > 3 || arguments.size() < 1)
            throw Exception(ErrorCodes::NUMBER_OF_ARGUMENTS_DOESNT_MATCH,
                "Incorrect number of parameters for aggregate function {}, should be 0, 1 or 2, got: {}", node->name, parameters.size());

        /// ...
```

yl-lisen (approved):

Issue #918, opened 2025-03-06
fail to build: "error: no member named 'headObject' in namespace 'DB::S3'"


zyxdenny:
**Describe what's wrong**

Fail to build with `ninja`.

**How to reproduce**
The error happens after running `ninja`. I used clang 19.1.7 to build on my Linux laptop (OpenSUSE).

**Error message and/or stacktrace**

```
proton/src/IO/WriteBufferFromS3.cpp:183:29: error: no member named 'headObject' in namespace 'DB::S3'
  183 |         auto response = S3::headObject(*client_ptr, bucket, key, "", write_settings.for_object_storage);
      |                         ~~~~^
1 error generated.
ninja: build stopped: subcommand failed.
```

**Additional context**

zliang-min:
Hi @zyxdenny thank you for reporting the issue. Do you mind sharing all the steps that you took to compile the projects?

zyxdenny:
Hi! I followed the steps in BUILD.md which is basically
```
$ git clone --recurse-submodules --depth=1 git@github.com:timeplus-io/proton.git
$ cd proton
$ mkdir -p build && cd build && cmake ..
$ ninja
```
The only difference is I made a depth=1 clone. Then the error happened when building with ninja

zliang-min:
Thank you for sharing, we will look into it. Meanwhile, could you please try replacing `cmake ..` with `cmake .. -DENABLE_AWS_S3=OFF` and try run `ninja` again?

zyxdenny:
Thank! That seems to be working.
As for the error, I found this in `proton/src/IO/S3/getObjectInfo.cpp`:
```
namespace DB::S3
{

namespace
{
Aws::S3::Model::HeadObjectOutcome headObject(
    const S3::Client & client, const String & bucket, const String & key, const String & version_id, bool for_disk_s3)
{
    ProfileEvents::increment(ProfileEvents::S3HeadObject);
    if (for_disk_s3)
        ProfileEvents::increment(ProfileEvents::DiskS3HeadObject);

    S3::HeadObjectRequest req;
    req.SetBucket(bucket);
    req.SetKey(key);

    if (!version_id.empty())
        req.SetVersionId(version_id);

    return client.HeadObject(req);
}
```
It seems that `headerObject` is defined in an unnamed namespace. I'm not familiar with the code-base, but I guess this makes the function inaccessible from other translation units.

zliang-min:
[...]

PR #922, opened 2025-03-07
`cmake ..` will not activate `ENABLE_AWS_S3` by default.

yokofly:
we still have some work to find missing code about aws_s3.
the aws_msk is good for now.
by default: `cmake ..` now will not try build aws_s3.  but -DENABLE_AWS_S3=ON still valid

follow up #921


yokofly:
more in #918

PR #930, opened 2025-03-22
chore: update FixedSizeDequeWithGaps.h

eltociear:
minor fix


jovezhong:
Looks good. Thanks for fixing the typo. Let us know if you find other issues like this, we can put them in the same PR. Otherwise I will merge it in next couple days

PR #1004, opened 2025-10-29
More enhancements and bugfixes

yl-lisen:


yokofly:
Since this PR contains many commits, we would like to preserve each commit instead of squashing them.
I have enabled the setting, so please choose **[Create a merge commit]** instead of Squash

Also, the stateless test case failure appears to be related — please verify it before merging.

cmake/autogenerated_versions.txt
```diff
@@ -2,7 +2,7 @@
 
 # NOTE: has nothing common with DBMS_TCP_PROTOCOL_VERSION,
 # only DBMS_TCP_PROTOCOL_VERSION should be incremented on protocol changes.
-SET(VERSION_REVISION 136)
+SET(VERSION_REVISION 140)
```
yokofly, line 5: intentional?

yokofly (approved):

cmake/autogenerated_versions.txt
yl-lisen: Yeah, no effects.

PR #1077, opened 2025-12-27
Docs/iceberg example

gangtao:
This PR provide a sample a read/write iceberg table using local minio as iceberg storage backend


examples/iceberg/script/proton.sql
```diff
@@ -0,0 +1,30 @@
+CREATE DATABASE demo
+SETTINGS
+    type = 'iceberg',
+    catalog_uri = 'http://iceberg-rest:8181',
+    catalog_type = 'rest',
+    warehouse = 's3://warehouse/',
+    storage_endpoint = 'http://minio:9000',
```
yokofly:
```suggestion
    storage_endpoint = 's3://warehouse/',
```


 `storage_endpoint = 'http://minio:9000'` This works for the current demo because the table is created by pyiceberg and Proton reads the table location from Iceberg metadata, so it never exercises Proton’s `CREATE STREAM` path.

  When Proton creates a table, HTTP endpoints are converted to `s3://<host>/...`, so `http://minio:9000` resolves to bucket `minio` instead of the `warehouse` bucket created by the
  stack. Suggest switching `storage_endpoint` to `s3://warehouse/` and adding an optional Proton‑created table example in `examples/iceberg/script/proton.sql`, e.g.:

```
-- Optional: create and write an Iceberg table from Proton (tests storage_endpoint path)
CREATE STREAM IF NOT EXISTS demo.proton_events
(
    id int64,
    timestamp datetime64(6, 'UTC'),
    user_id string,
    event_type string,
    value nullable(float64)
);

INSERT INTO demo.proton_events (id, timestamp, user_id, event_type, value)
VALUES (101, now64(6), 'user_9', 'login', NULL);

SELECT id, timestamp, user_id, event_type, value FROM demo.proton_events;

```

examples/iceberg/docker-compose.yaml
```diff
@@ -0,0 +1,83 @@
+version: '3.8'
+
+services:
+  # Iceberg REST Catalog
+  iceberg-rest:
+    image: tabulario/iceberg-rest:latest
```
yokofly: Prefer pin a fixed version instead pulling latest

examples/iceberg/script/proton.sql
gangtao: Thanks @yokofly I think this is an improvment proton should consider
in my example, an extra configuration is required for local minio support

```
url_scheme_mappers:
  s3:
    to: http://minio:9000/{bucket}
```

I think we should use storage_endpoint to get the real s3 url instead of using extra settings.

I understand this is how proton works for now, please discuss with @yuzifeng1984 what is the best way to make proton support s3 url.

examples/iceberg/docker-compose.yaml
gangtao: make sense

yokofly (approved):

PR #1108, opened 2026-02-09
Port CH #82137: settings over URI is broken

yokofly:
PR checklist:
- Did you run ClangFormat ?

clickhouse/clickhouse#82137

### DIFF
--- a/src/Server/HTTPHandler.cpp
+++ b/src/Server/HTTPHandler.cpp
@@ -1,4 +1,4 @@
-#include <Server/HTTPHandler.h>
+﻿#include <Server/HTTPHandler.h>
 
 #include <Access/Authentication.h>
 #include <Access/Credentials.h>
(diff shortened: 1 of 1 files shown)


src/Server/HTTPHandler.cpp
```diff
@@ -1,4 +1,4 @@
-#include <Server/HTTPHandler.h>
+﻿#include <Server/HTTPHandler.h>
```
yokofly, line 1: That include-line diff is from a hidden UTF-8 BOM (U+FEFF) added at the start of src/Server/HTTPHandler.cpp, not from a real include change.

  GPT verified:

  - HEAD: starts with bytes ef bb bf 23 (BOM + #)
  - HEAD~1: starts with 23 (# only)

  So Git renders it as:

  - old: #include ...
  - new: ﻿#include ... (same text, but with invisible BOM before #)

PR #1141, opened 2026-03-23
feat: implement PostgreSQL extended query protocol (Parse/Bind/Execute/Sync)

### LINKED ISSUE #225: PostgreSQL protocol support
**Use case**

As a OLAP user, i would like to connect proton via postgresql protocol. psql / DBeaver are initially integration / testing points 

**Additional context**

SBALAVIGNESH123:
This PR adds support for the PostgreSQL extended query protocol, which is needed for tools like DBeaver and pgAdmin to connect to Proton. Currently, the handler rejects all extended query messages (Parse, Bind, Describe, Execute, Sync, Close, Flush) with an error saying "proton doesn't support extended query mechanism". This means most modern PostgreSQL drivers and GUI tools fail on connect since they default to the extended query protocol for prepared statements and parameter binding.
The approach here is server-side parameter substitution — the client sends Parse/Bind with $1, $2 placeholders and parameter values, and we substitute them into the SQL before passing it to Proton's existing query engine. This way we get full client compatibility without any changes to the parser or interpreter. A new PreparedStatementManager handles the lifecycle of named and unnamed statements and portals.I also added interception for the various system queries that PG clients send on connect (pg_catalog, SET, SHOW server_version, SELECT version(), SELECT current_schema(), transaction commands, etc.) so that clients don't crash when they can't find system tables. The type mapping was expanded to cover Bool, DateTime, DateTime64, Enum, Map, Array, and large integer types.The parameter substitution is string-literal-aware (won't replace $1 inside single quotes) and escapes single quotes and backslashes to prevent SQL injection. Error handling follows the PG spec — after an error during extended query, all subsequent messages are skipped until a Sync message resets the pipeline.
"This is intentionally a compatibility layer — same approach CockroachDB used in their early PG wire support. Server-side substitution gives us 95% client compatibility with zero query engine changes. True prepared execution is a future enhancement."

Closes #225


SBALAVIGNESH123:
Hi @yokofly,

Just following up on this PR. Could you please let me know if there is any feedback, review, or additional work needed from my side? I'd be happy to make any required changes or improvements. Thanks for your time and consideration.

src/Server/PostgreSQLHandler.cpp
```diff
@@ -333,4 +360,384 @@ bool PostgreSQLHandler::isEmptyQuery(const String & query)
...
+    std::transform(upper_query.begin(), upper_query.end(), upper_query.begin(),
+                   [](unsigned char c) { return std::toupper(c); });
+
+    if (upper_query.find("SET ") == 0)
+    {
+        message_transport->send(
+            PostgreSQLProtocol::Messaging::CommandComplete(
+                PostgreSQLProtocol::Messaging::CommandComplete::Command::SELECT, 0));
+        return true;
+    }
```
yokofly, line 582: Does `SET unknown = 1` currently return fake success?

Since PG and Proton have different setting namespaces, no-oping every `SET` seems confusing.

Should we only no-op known PG bootstrap settings and let Proton settings like `max_threads` reach `executeQuery()`?

src/Core/PostgreSQLProtocol.cpp
```diff
@@ -32,16 +39,35 @@ ColumnTypeSpec convertTypeIndexToPostgresColumnTypeSpec(TypeIndex type_index)
...
 
         case TypeIndex::UUID:
             return {ColumnType::UUID, 16};
 
+        case TypeIndex::Enum8:
+        case TypeIndex::Enum16:
+            return {ColumnType::VARCHAR, -1};
+
+        case TypeIndex::Map:
+            return {ColumnType::JSONB, -1};
```
yokofly, line 65: have you verified the map<-> jsonB is work? can u add test cover?

yokofly (commented):
I feel the implementation is too broad — there's a lot of behavior to get right across the full Parse → Bind → Execute → Sync flow, and without a concrete requirement driving the scope, it's hard for us to review and maintain. So let's hold on this one for now.

SBALAVIGNESH123:
Hi @yokofly,

Thank you for the review and detailed feedback.

Given that the scope is currently too broad and I am unable to set up a Linux environment to build, test, and address the merge conflicts at this time, I am closing this PR.

If there is a concrete requirement or interest in revisiting PostgreSQL extended query protocol support in the future, we can use this work as a baseline and narrow down the scope as needed.

Thanks again for your time!

PR #1156, opened 2026-03-31
ci: add trusted mirror workflow for fork PRs

### LINKED ISSUE #1154: ci: enable CI for fork PRs via label-gated pull_request_target
## Problem

When an external contributor opens a PR from a fork repository, CI fails immediately with:

```
Input required and not supplied: aws-region
```

**Root cause:** GitHub Actions does not pass repository secrets to workflows triggered by `pull_request` events from forks. This is a security feature and cannot be bypassed. Since all Proton CI jobs require AWS credentials (EC2 self-hosted runners), DockerHub credentials (docker push/pull), and GH_PERSONAL_ACCESS_TOKEN, no meaningful CI work can run without secrets.

Reference failure: https://github.com/timeplus-io/proton/actions/runs/23742524909

## Solution

Adopt the **label-gated `pull_request_target`** pattern (same approach used by ClickHouse with their `can be tested` label):

1. **Internal PRs** (same repo): continue using `pull_request` trigger — auto CI, no change.
2. **Fork PRs**: use `pull_request_target` trigger with `labeled` type. A maintainer reviews the code, then adds the existing `pr-test` label. This triggers CI in the base-repo context where secrets are available.

### Changes required

**`proton_ci.yml`:**
- Add `pull_request_target` trigger (type: `labeled`, branches: develop/2.9/3.0)
- Update concurrency group to use PR number (consistent across both trigger types)
- Add fork/label gate to `build_address_x64` job condition:
  - `pull_request` + same repo → run (internal PR)
  - `pull_request_target` + fork repo + `pr-test` label → run (approved fork PR)
- Pass `ref: refs/pull/N/head` for `pull_request_target` events so fork code is checked out
- Update test job conditions to check `needs.build_address_x64.result == 'success'`
- Update `clean_passed_test_image` to also run for approved fork PRs
- Remove the `SkipExternalFork` placeholder job (no longer needed)

**`run_command.yml`:**
- Revert fork-skip guards added on current branch (no longer needed; fork PRs now use `pull_request_target` which has secrets)
- Add a step after checkout to align `GITHUB_SHA` with the actual checked-out HEAD when a custom `ref` is passed (prevents docker image tag collisions for `pull_request_target` runs)

### Security model

- `pull_request_target` always uses the workflow file from the **base branch** — a fork cannot modify the CI pipeline
- The `pr-test` label is the approval gate — only maintainers can add labels
- After new pushes, maintainer must re-review and re-apply the label
- No fork code executes until a maintainer explicitly approves via label

### Testing

- Internal PR: CI should auto-trigger as before
- Fork PR without label: CI should not run (jobs skipped)
- Fork PR with `pr-test` label: CI should run with full secrets
- `GITHUB_SHA` in docker image tags should be the PR head SHA (not base branch SHA)

yl-lisen:
## Summary

- skip secret-backed `ProtonCI` jobs on external fork `pull_request` events
- add a maintainer-only `trusted_fork_pr_ci` workflow that dispatches `ProtonCI` manually for a fork PR
- dispatch `ProtonCI` on the target base branch, but checkout the fork PR code directly via `refs/pull/<pr>/head`
- fail fast if the fork PR moves after dispatch so the checked-out HEAD cannot drift away from the dispatched `head_sha`
- keep the fork-PR SHA override local to the trusted `ProtonCI` path so image tags and cleanup still use the fork head SHA
- do not use `pull_request_target`, mirror branches, or PR comment syncing

## Security model

- does not use `pull_request_target`
- fork code only runs with secrets after a maintainer manually triggers `trusted_fork_pr_ci`
- the workflow definition still comes from the base repository branch used for `workflow_dispatch`
- trusted CI checks out the fork PR head directly through `refs/pull/<pr>/head`, so no extra trusted branch needs to be created in the base repository
- if the PR head changes after dispatch, the trusted jobs fail instead of silently building a different commit
- internal PRs keep the existing `pull_request` flow unchanged

## Feasibility validation

- validated against real cross-repo PR `#1140`
- `gh pr view 1140` reports it as `isCrossRepository=true`
- `git fetch origin refs/pull/1140/head && git rev-parse FETCH_HEAD` resolves to the same head SHA shown by the PR, so the base repository can read the fork PR head directly without a mirror branch

## Test plan

- [x] YAML parse for modified workflows
- [x] `bash -n` validation for the updated workflow `run:` scripts
- [x] verified that `refs/pull/1140/head` resolves to the fork PR head SHA
- [ ] end-to-end run via `trusted_fork_pr_ci` on a real fork PR

Closes #1154


.github/workflows/trusted_fork_pr_ci.yml
```diff
@@ -0,0 +1,93 @@
...
+          if [[ "$is_draft" == "true" ]]; then
+            echo "::error::PR #${PR_NUMBER} is still a draft. Mark it ready for review before running trusted CI."
+            exit 1
+          fi
+
+          tmp_dir="${GITHUB_WORKSPACE:-$PWD}/tmp"
+          mkdir -p "$tmp_dir"
+          dispatch_file="$(mktemp -p "$tmp_dir")"
+          trap 'rm -f "$dispatch_file"' EXIT
+
```
yokofly, line 68: **P2: CI results not attached to fork PR.**

`workflow_dispatch` records the run against `base_ref` (e.g. `develop`), not the fork PR's `head_sha`. The PR page won't show check status from the trusted run.

Low priority for a small team — maintainer knows they triggered it and can check Actions tab. If it becomes annoying, add a `gh api repos/.../statuses/${head_sha}` call after dispatch to post a pending status linking to the dispatched run.

.github/workflows/trusted_fork_pr_ci.yml
```diff
@@ -0,0 +1,93 @@
...
+        run: |
+          pr_json="$(gh api "repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}")"
+
+          pr_state="$(jq -r '.state' <<<"$pr_json")"
+          is_cross_repo="$(jq -r '.head.repo.full_name != .base.repo.full_name' <<<"$pr_json")"
+          is_draft="$(jq -r '.draft' <<<"$pr_json")"
+          base_ref="$(jq -r '.base.ref' <<<"$pr_json")"
+          head_sha="$(jq -r '.head.sha' <<<"$pr_json")"
+          labels_csv="$(jq -r '[.labels[].name] | join(",")' <<<"$pr_json")"
+          pr_ref="refs/pull/${PR_NUMBER}/head"
```
yokofly, line 38: **P2: Floating ref — race between dispatch and checkout.**

`pr_ref` is `refs/pull/N/head` (mutable). If the contributor pushes after dispatch but before EC2 checkout, the job builds the new commit while tagging images with the old `head_sha`.

Practically rare for your fork PR volume. One-line fix in each job's command block after checkout:
```bash
if [[ -n "$EXPECTED_SHA" ]] && [[ "$(git rev-parse HEAD)" != "$EXPECTED_SHA" ]]; then
  echo "::error::HEAD doesn't match dispatched SHA. PR was updated after dispatch."
  exit 1
fi
```

yokofly (commented):
Looks good overall — no `pull_request_target`, clean security model. Two low-priority issues worth a follow-up fix.

yokofly (approved):

PR #1175, opened 2026-05-15
Enable shard pruning for IN-subquery and literal tuple-IN

sofiane-soufi:
## Summary

Two related planner improvements that close shard-pruning gaps for any query shaped like

```sql
WHERE (k1, k2, ...) IN (literal-list)
-- or
WHERE k IN (SELECT k FROM small_lookup WHERE pk = ?)
```

against a sharded Stream. Before this PR, both shapes silently fell back  to fanning out to **every shard** once they crossed about five keys, making the natural query form unusable on multi-TB streams.

Typical workloads that hit this:

- User-/tenant-/account-keyed analytics where the lookup table maps an owner to its N child keys, and the events stream is sharded by those child keys.
- Multi-entity aggregations: "show me everything that happened across this list of N entities."
- Composite-key point-in-time queries where the planner has the full key set in hand but the IN-list crosses the bloom-filter compounding threshold.

### 1. `optimize_skip_unused_shards_with_subqueries` (off by default)

When the setting is on, bounded IN-subqueries are materialized via `PreparedSets` and the AST is rewritten to a literal IN **before** `PruneShards.cpp` evaluates the sharding expression. Existing literal-IN pruning then applies unchanged.

Guards:

- Respects `optimize_skip_unused_shards_limit`.
- Conservative fallback (no rewrite) for: `NOT IN`, OR/NOT context, empty subquery (under unsafe logic), `NOT_IMPLEMENTED` from `buildOrderedSetInplace`, and **any NULL element** — dropping NULLs is unsafe for nullable sharding keys / `transform_null_in = 1`.
- Off by default, so no behavior change for existing deployments.

### 2. Literal tuple-IN pushed to shard pruning

`analyzeFunction` in `evaluateConstantExpression.cpp` previously matched only single-`ASTIdentifier` LHS for `IN`. For `(c1, c2) IN ((v11, v12), (v21, v22), ...)` the LHS is `tuple_cast(...)`, so the analyzer returned an empty disjunction and shard pruning gave up.

The new branch decomposes tuple-IN into a disjunction of conjunctions (one conjunction per row, one equality per column). Handles both AST shapes the parser produces:

- Single `ASTLiteral` whose `Field` is a `Tuple` of inner `Tuple`s.
- `tuple_cast(tuple_cast(...), tuple_cast(...))` with optional `cast(literal, type)` wrappers — unwrapped via the shared `isFunctionCast` helper.

Single-column IN (existing path) is untouched, and non-literal values trigger a safe fallback (return empty disjunction → no rewrite).

## Repro

Generic schema with a composite sharding key:

```sql
CREATE STREAM events (
  k1   uint32,
  k2   string,
  kind low_cardinality(string),
  ...
)
ENGINE = Stream(N, city_hash64((k1, k2)))
ORDER BY (k1, k2, kind, _index);
```

Behavior on an N-shard cluster, with a moderately large IN-list:

| Form                                                                 | Before                            | After                          |
| -------------------------------------------------------------------- | --------------------------------- | ------------------------------ |
| `k1 = X AND k2 IN ('a', 'b', ...)` (large literal list)              | fans out to all shards            | prunes to matching shards      |
| `k2 IN (SELECT k2 FROM lookup WHERE owner = ?)`                      | timeout (subquery skips pruning)  | prunes to matching shards      |
| `(k1, k2) IN ((X, 'a'), (X, 'b'), ...)` (literal tuple-IN)           | timeout (tuple LHS skips pruning) | prunes to matching shards      |

## Test plan

- [x] New `03000_subquery_in_shard_pruning` exercises setting OFF/ON, empty subquery, NULL handling (`transform_null_in = 1`), `optimize_skip_unused_shards_limit` clamp, `NOT IN` fallback, multiple IN-subqueries in one predicate.
- [x] New `03001_tuple_in_shard_pruning` exercises bare tuple literals, explicit `::T` casts, mixed-`k1` rows, non-literal value fallback, and limit clamp.
- [x] Existing `02374_in_tuple_index` (granule-level tuple-IN) still passes — proves no regression to the `KeyCondition` path.
- [x] Verified end-to-end on a multi-shard repro: tuple-IN with N keys now hits only the shards that actually contain matching rows instead of all of them.

## Out of scope

- `JOIN` keys → multi-key PK seek (separate change, requires JOIN planner work).
- `join_algorithm = 'direct'` against regular Streams (separate ask).
- Tuple-IN at the **partition** key level — partition pruning uses `KeyCondition::tryPrepareSetIndex`, which already handles tuple-IN correctly; this PR doesn't touch that path.


yokofly:
@codex review

sofiane-soufi:
Ready for another review.

yokofly:
@codex review

src/Storages/tests/gtest_prune_shards_rewrite_in_subqueries.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include <Parsers/ASTFunction.h>
+#include <Parsers/ASTIdentifier.h>
+#include <Parsers/ASTLiteral.h>
+#include <Parsers/IAST.h>
+#include <QueryPipeline/SizeLimits.h>
+#include <Storages/PruneShardsInternal.h>
+
+#include <Common/tests/gtest_global_context.h>
+
+namespace DB::tests
```
yokofly:
```suggestion
namespace DB
```

tests/queries_ported/0_stateless/03000_subquery_in_shard_pruning.sql
yl-lisen, line 1: Current cases are only for the correctness, we need also check if the pruned operation is working by explain its pipeline/actions etc.

tests/queries_ported/0_stateless/03001_tuple_in_shard_pruning.sql
yl-lisen, line 1: Same with above

sofiane-soufi: Fixed in commit [82c63e0](https://github.com/timeplus-io/proton/pull/1175/commits/82c63e0b57767764f1ee90d319f63b12f0e5fb77)

tests/queries_ported/0_stateless/03000_subquery_in_shard_pruning.sql
sofiane-soufi: Fixed in commit [82c63e0](https://github.com/timeplus-io/proton/pull/1175/commits/82c63e0b57767764f1ee90d319f63b12f0e5fb77)

src/Storages/PruneShards.cpp
```diff
@@ -94,6 +106,186 @@ void replaceConstantExpressions(
...
+                return changed;
+
+            future_set = prepared_sets->addFromSubquery(
+                right_hash,
+                std::move(subquery_plan),
+                /*external_table=*/nullptr,
+                /*external_table_set=*/nullptr,
+                context->getSettingsRef());
+        }
+        catch (...)
```
yl-lisen: **`catch (...)` is too broad.**

This swallows everything, including `std::bad_alloc` (real OOM) and `std::logic_error` from broken AST invariants — both of which should propagate immediately rather than silently degrade to "no shard pruning".

Suggest narrowing to `catch (const Exception &)`. The legitimate targets here — `NOT_IMPLEMENTED` from streaming subqueries, parser/resolver errors from unsupported shapes — are all `DB::Exception`. Anything outside that hierarchy is a programmer/runtime bug worth surfacing.

src/Storages/PruneShards.cpp
```diff
@@ -94,6 +106,186 @@ void replaceConstantExpressions(
...
+    {
+        /// An empty IN-subquery is always false. Only flag it for shard short-circuiting when
+        /// it sits in a pure conjunctive position -- otherwise an enclosing operator may still
+        /// evaluate the whole predicate to true for some rows (e.g. `if(flag, id IN (empty), 1)`).
+        if (in_conjunctive_position)
+            result.has_empty_subquery_in_conjunctive_position = true;
+        return changed;
+    }
+
+    if (total_rows > limit)
```
yl-lisen, line 274: **Bound check happens after materialization, not before.**

`set->getTotalRowCount()` is compared against `limit` on line 238, but the set itself was already fully materialized by `buildOrderedSetInplace` on line 215. That materialization is bounded only by `use_index_for_in_with_subqueries_max_values` (default much larger than `optimize_skip_unused_shards_limit`, which defaults to 1000), so a million-row subquery allocates a million `Field`s before being rejected as too large for shard pruning.

Suggest either passing `limit + 1` as a `max_elements_to_fill` cap when constructing the `Set`, or short-circuiting on the rough row-count estimate (e.g. the subquery's `LIMIT` or storage stats) before calling `buildOrderedSetInplace`.

Minor follow-up at line 268 (`tuple.push_back(value)`): each `Field` is copied. For heap-typed sharding keys (string, array), moving from a non-const accessor would avoid the per-row copy.

tests/queries_ported/0_stateless/03000_subquery_in_shard_pruning.sql
```diff
@@ -0,0 +1,83 @@
...
+select count() from table(03000_subquery_in_shard_pruning) where if(value = 10, 1, id in (select 1 where 0));
+select count() from table(03000_subquery_in_shard_pruning) where multi_if(value = 10, 1, id in (select 1 where 0), 99, 1);
+-- Empty IN inside a `tuple` element must not zero out the shard set either.
+select count() from table(03000_subquery_in_shard_pruning) where tuple_element((1, id in (select 1 where 0)), 1) = 1;
+
+-- Historical IN-subquery against the same stream should match every row and read every shard.
+select count() from table(03000_subquery_in_shard_pruning) where id in (select id from table(03000_subquery_in_shard_pruning));
+
+-- Early shard-pruning materialization must preserve normal subquery-depth validation.
+select count() from (select * from (select * from table(03000_subquery_in_shard_pruning) where id in (select 1))) settings max_subquery_depth = 2; -- { serverError 162 }
```
yl-lisen: **This regression test doesn't isolate the `subquery_depth` plumbing fix.**

The query wraps the SELECT in two `SELECT * FROM (...)` levels, so the outer nesting alone already trips `serverError 162` via `InterpreterSelectQuery`'s normal depth check — independent of whether `query_info.subquery_depth` is forwarded into the early materialization path. If a future refactor reverted to the previously-hardcoded `interpretSubquery(right, context, 1, {})`, this test would still pass.

Suggest nesting *inside* the IN-subquery body so the limit can only fire if the early path forwards depth correctly:

```sql
SELECT * FROM table(03000_subquery_in_shard_pruning)
WHERE id IN (SELECT id FROM (SELECT id FROM (SELECT 1 AS id)))
SETTINGS max_subquery_depth = 2; -- { serverError 162 }
```

src/Storages/tests/gtest_prune_shards_rewrite_in_subqueries.cpp
```diff
@@ -0,0 +1,181 @@
...
+{
+
+namespace
+{
+
+/// Registers a single empty Set under the tree-hash of an `ASTTableIdentifier` so that the
+/// rewrite helper's `findStorage()` path resolves to it. This lets us deterministically trip
+/// the empty-IN branch without spinning up a real subquery execution.
+class EmptyInFixture
```
yl-lisen, line 43: **`EmptyInFixture` only covers the `findStorage` / `ASTTableIdentifier` branch.**

All 12 test cases register the empty set via `prepared_sets->addFromStorage(...)` (line 44) with an `ASTTableIdentifier`. The central new branch added by this PR — the `addFromSubquery` materialization at `PruneShards.cpp:177-206`, the `isStreaming()` bail-out, and the broad `catch (...)` fallback — has no unit coverage; deleting that entire block would not flip a single gtest. The SQL tests cover it end-to-end but cannot isolate streaming-plan detection.

Suggest adding one fixture that hands the helper an `ASTSubquery` (so `addFromSubquery` actually fires), or splitting the streaming check into a small helper that can be unit-tested with a stub `QueryPlan`.

yl-lisen (commented):
A few inline suggestions on the new shard-pruning rewrite — none are blockers; the change looks well-engineered and the commit history walks through the trade-offs explicitly. The comments below focus on planning-stage cost, a too-broad catch, layer coupling on `SelectQueryInfo`, an over-permissive empty-shards branch in `StorageStream`, and two test-coverage gaps that would let the corresponding fixes silently regress.

yl-lisen:
CI build failed https://github.com/timeplus-io/proton/actions/runs/26011178388/job/76452101061

yokofly:
> CI build failed https://github.com/timeplus-io/proton/actions/runs/26011178388/job/76452101061

I tried to follow the previous work in [https://github.com/timeplus-io/proton/pull/1156](https://github.com/timeplus-io/proton/pull/1156) to support running CI on external contributors’ PRs, but it seems the correct actions were not triggered. As a workaround, I manually forked the branch and pushed two new commits:

 7eef46f91c77833279e2f57984151ab54747d547
f34912c254a2611c44cc21ef08e3320a3c9797ff

See:

* [https://github.com/timeplus-io/proton/actions/runs/26009829215](https://github.com/timeplus-io/proton/actions/runs/26009829215)
  The action output looks correct, but it uses the `develop` HEAD commit instead of the PR branch:

* [https://github.com/timeplus-io/proton/actions/runs/26009832845](https://github.com/timeplus-io/proton/actions/runs/26009832845)

yl-lisen:
https://github.com/timeplus-io/proton/actions/runs/26016349848

src/Storages/PruneShards.cpp
sofiane-soufi: Addressed in commit [4414e88050](https://github.com/timeplus-io/proton/pull/1175/commits/4414e880505d72030ec245f0c3a8fc7af7ea03f0)

src/Storages/PruneShards.cpp
sofiane-soufi: Addressed in commit [4414e88050](https://github.com/timeplus-io/proton/pull/1175/commits/4414e880505d72030ec245f0c3a8fc7af7ea03f0)

tests/queries_ported/0_stateless/03000_subquery_in_shard_pruning.sql
sofiane-soufi: Addressed in commit [4414e88050](https://github.com/timeplus-io/proton/pull/1175/commits/4414e880505d72030ec245f0c3a8fc7af7ea03f0)

src/Storages/tests/gtest_prune_shards_rewrite_in_subqueries.cpp
sofiane-soufi: Addressed in commit [4414e88050](https://github.com/timeplus-io/proton/pull/1175/commits/4414e880505d72030ec245f0c3a8fc7af7ea03f0)

src/Storages/PruneShards.cpp
```diff
@@ -94,6 +108,219 @@ void replaceConstantExpressions(
...
+
+        tuple.push_back(std::move(value));
+    }
+
+    if (tuple.empty())
+    {
+        /// All rows were either NULL (handled above) or filtered out by validation —
+        /// keep the original subquery to preserve correctness.
+        return changed;
+    }
```
yokofly: I removed this branch from my forked one 

https://github.com/timeplus-io/proton/pull/1177/changes/7eef46f91c77833279e2f57984151ab54747d547#diff-dfdaf3f59489c60e4379b4b75296ae6bf58dd9ec9da3a345ee22303a906f6185L271-L277


after
```
   if (!values || values->size() == 0)
    {
        if (in_conjunctive_position)
            result.has_empty_subquery_in_conjunctive_position = true;
        return changed;
    }

```

Inside the loop above, we always perform at least one tuple.push_back(), so this branch `if (tuple.empty())` is never reachable.

Alternatively, we could add:`chassert(!tuple.empty());`

src/Storages/tests/gtest_prune_shards_rewrite_in_subqueries.cpp
yokofly: I'm not clear on why we need this. Could we align the style with the refinement in:
https://github.com/timeplus-io/proton/pull/1177/changes/7eef46f91c77833279e2f57984151ab54747d547

src/Storages/PruneShards.cpp
sofiane-soufi: Sorry, I missed that you'd pushed #1177 with the cleanups, I was working off the inline review comments and didn't realize the fork branch already had concrete patches.

I picked up the substantive changes in commit 34f7f6c1c7:

- Removed the unreachable if (tuple.empty()) block, replaced with chassert(!tuple.empty()) per your alternative suggestion.
- Renamed DB::Internal → DB::detail and PruneShardsInternal.h → PruneShardsDetail.h to match the codebase convention.
- Dropped the namespace DB::tests { … } wrapper in the gtest in favor of using namespace DB; at file scope, matching the style of the other Storages gtests.
- Tidied the PruneShards.cpp includes and added the missing ARGUMENT_OUT_OF_BOUND extern.
- Cherry-picked your three additional gtests (non-empty rewrite, NULL element, limit-exceeded short-circuit),  they cover ground my new tests didn't touch.

One note: inside the gtest the detail references are qualified as DB::detail::… to disambiguate from ::detail in base/find_symbols.h, which  becomes visible once using namespace DB; is in scope.

yl-lisen:
https://github.com/timeplus-io/proton/actions/runs/26076287482 , CI PASS - ✅

src/Interpreters/evaluateConstantExpression.cpp
```diff
@@ -224,6 +224,109 @@ namespace
                 return true;
             };
 
+            /// Multi-column tuple-IN: (c1, c2, ...) IN ((v11, v12, ...), (v21, v22, ...), ...).
+            /// Each inner tuple yields a conjunction "c1=v1i AND c2=v2i AND ..." and the rows are ORed.
```
chenziliang, line 229: Have you covered duplicated tuples on the right ? like `IN ((v11, v12), (v11, v12))` ? If we don't do dedup, simply ORed conjunction produce duplicate rows.

sofiane-soufi: Good catch, addressed in [57dc4fd95f]. Two line fix inside add_tuple_row: track seen std::vector<Field> row keys in a std::set and skip duplicates before they reach the result.

Added regression coverage in 03001_tuple_in_shard_pruning.sql: the EXPLAIN-pipeline case (IN ((1, 'addr_42'), (1, 'addr_42'), (1, 'addr_88')) with optimize_skip_unused_shards_limit = 2) shows 2 MergeTreeSelect steps with the fix, 8 without, verified by reverting just the dedup guard locally.

PR #1192, opened 2026-06-10
fix: throw bad_alloc on failed ByteVector allocation instead of NULL deref

yokofly:
Check the `malloc` result in `ByteVector(size_t)` and throw `std::bad_alloc` on failure instead of leaving a null buffer for a later NULL-deref (e.g. on a torn-log read of a garbage size). Adds a regression test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yokofly:
@codex review

PR #1193, opened 2026-06-10
ci: declare AWS_ACCESS_KEY_ID in compatibility_test workflow_call

### LINKED ISSUE #1191: NightlyTest fails at startup: AWS_ACCESS_KEY_ID passed to compatibility_test.yml, which no longer declares it
## Summary

The `NightlyTest` workflow fails immediately with **`startup_failure`** — GitHub rejects the workflow file before any job runs. `proton_ci` is unaffected.

## Symptom

- Latest `workflow_dispatch` run ends in `startup_failure` with `jobs: []` and GitHub's generic "This run likely failed because of a workflow file issue."
- `NightlyTest` last succeeded on 2026-03-31; every run since has been `startup_failure`.
- `proton_ci` passes on the same commits, so this is a workflow-definition problem, not a test/code failure.

## Root cause

`nightly_test.yml` is the only workflow that calls the reusable `compatibility_test.yml`. That call passes `AWS_ACCESS_KEY_ID` in its `secrets:` block:

```yaml
uses: ./.github/workflows/compatibility_test.yml@develop
secrets:
  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
  ...
```

But `compatibility_test.yml`'s `on.workflow_call.secrets` no longer **declares** `AWS_ACCESS_KEY_ID`. Passing a secret that the called workflow doesn't declare is a startup-time validation error in GitHub Actions — and it fires even though the compatibility job is guarded by `if: false`, because the reusable-workflow interface is validated when the workflow is compiled, before any `if` is evaluated. That aborts the entire run before any job starts.

`proton_ci` only calls `run_command.yml`, which **does** declare `AWS_ACCESS_KEY_ID`, so it never hits the mismatch.

## How it got here

- One CI cleanup PR removed the `AWS_ACCESS_KEY_ID` declaration from `compatibility_test.yml`'s `workflow_call.secrets` (and removed the matching call-site usages).
- A follow-up PR re-added the call-site usages (including the nightly → compatibility call) and restored the declaration in `run_command.yml`, but did **not** restore the declaration in `compatibility_test.yml`.

Net result is an asymmetry: `nightly_test.yml` passes a secret that `compatibility_test.yml` no longer accepts.

## Fix

Restore the declaration in `.github/workflows/compatibility_test.yml` under `on.workflow_call.secrets` (insert above `AWS_SECRET_ACCESS_KEY`):

```yaml
    secrets:
      AWS_ACCESS_KEY_ID:
        required: true
      AWS_SECRET_ACCESS_KEY:
        required: true
      ...
```

This realigns the callee interface with what `nightly_test.yml` already passes and matches the last known-good state.

## Notes / verification

- Other usual startup-failure causes were ruled out: the file parses cleanly (no duplicate job keys), no dangling `needs:`, and every `run_command.yml` call carries all required secrets. The only net structural change to `nightly_test.yml` since the last green run is this secret churn plus a clang toolchain version bump inside shell `command:` strings (harmless).
- After the fix, re-run `NightlyTest` via `workflow_dispatch` and confirm jobs start.

yokofly:
Fixes #1191.

## Problem

`NightlyTest` fails immediately with `startup_failure` — GitHub rejects the workflow file before any job runs. `proton_ci` is unaffected.

`nightly_test.yml` is the only workflow that calls the reusable `compatibility_test.yml`, and its call passes `AWS_ACCESS_KEY_ID` in the `secrets:` block. But `compatibility_test.yml`'s `on.workflow_call.secrets` no longer **declared** that secret. Passing a secret the callee doesn't declare is a startup-time validation error in GitHub Actions, and it fires even though the compatibility job is guarded by `if: false` (interface validation happens at workflow-compile time, before `if` is evaluated). That aborts the whole run.

`proton_ci` only calls `run_command.yml`, which already declares `AWS_ACCESS_KEY_ID`, so it never hits the mismatch.

## Fix

Restore the `AWS_ACCESS_KEY_ID` declaration in `compatibility_test.yml`'s `workflow_call.secrets`, realigning the callee interface with what `nightly_test.yml` already passes (matches the last known-good state).

```diff
     secrets:
+      AWS_ACCESS_KEY_ID:
+        required: true
       AWS_SECRET_ACCESS_KEY:
         required: true
```

## Verification

- YAML parses; `workflow_call.secrets` now includes `AWS_ACCESS_KEY_ID`.
- After merge, re-run `NightlyTest` via `workflow_dispatch` and confirm jobs start.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/compatibility_test.yml
+++ b/.github/workflows/compatibility_test.yml
@@ -71,6 +71,8 @@ on:
         type: string
         default: "3"
     secrets:
+      AWS_ACCESS_KEY_ID:
+        required: true
       AWS_SECRET_ACCESS_KEY:
         required: true
       AWS_REGION:


yokofly:
@codex review

PR #1197, opened 2026-06-11
fix(postgres): only enable SSL when cert and key files actually exist on disk

### LINKED ISSUE #824: Cannot connect to postgres port 5432
**Describe what's wrong**

After starting Timeplus Proton with 5432 port exposed from docker or using bare metal installation, I failed to connect the 5432 Postgres port
```
psql -h 127.0.0.1 -p 5432
psql: error: connection to server at "127.0.0.1", port 5432 failed: SSL SYSCALL error: EOF detected
connection to server at "127.0.0.1", port 5432 failed: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```
The port 5432 is certainly opening for pgsql. Otherwise if I set a different port, the error message is different
```
psql -h 127.0.0.1 -p 5431
psql: error: connection to server at "127.0.0.1", port 5431 failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?
```

**Impact**
Some users want to use AWS QuickSight to build dashboard for Timeplus. It supports Postgres/MySQL/ClickHouse, but not Timeplus. Considering the long process of adding Timeplus data source to AWS product, we should allow users to use Postgres or MySQL port. This worked before but is broken for a while.

burnerlee:
PR checklist:
- Did you run ClangFormat ? Yes (no formatting changes in modified lines)
- Did you separate headers to a different section in existing community code base ? N/A (no new headers in inherited code)
- Did you surround `proton: starts/ends` for new code in existing community code base ? N/A (PostgreSQLHandlerFactory is not ClickHouse-inherited code)

## Problem

Connecting to Proton's PostgreSQL port (5432) with `psql` fails immediately:

```
psql: error: connection to server at "127.0.0.1", port 5432 failed: SSL SYSCALL error: EOF detected
        connection to server at "127.0.0.1", port 5432 failed: server closed the connection unexpectedly
```

This happens on any deployment where SSL certificate files are not present (the default).

## Root Cause

`PostgreSQLHandlerFactory` hardcoded `ssl_enabled = true` at compile time whenever `USE_SSL` was set. The `psql` client always sends an `SSL_REQUEST` first before the startup message. The handler entered `makeSecureConnectionSSL()`, which called `Poco::Net::SSLManager::instance().defaultServerContext()` — this throws/fails when no cert/key files are configured, and the connection drops with EOF before any error can be sent back to the client.

## Fix

Default `ssl_enabled` to `false`. In the factory constructor, read `openSSL.server.certificateFile` and `openSSL.server.privateKeyFile` from the server config at runtime. Only set `ssl_enabled = true` when both keys are non-empty **and** both files exist on disk.

When `ssl_enabled = false`, the handler correctly responds `'N'` to `SSL_REQUEST`, and `psql` falls back to plaintext — the expected behavior for a server without SSL configured.

Fixes #824


chenziliang (approved):

yokofly:
edit v1: assign->sign
please sign CLA https://github.com/timeplus-io/proton/pull/1197#issuecomment-4683069106 @burnerlee

burnerlee:
@yokofly done!

PR #1201, opened 2026-06-21
feat(kafka): support Avro-encoded message key via schema registry

### LINKED ISSUE #915: Kafka External Stream to support read the Avro-encoded message key
**Use case**

Today the message key is read as a string. If the producer create the message with Avro encoded message key, cannot read the content. This is also brought up in https://github.com/ClickHouse/ClickHouse/issues/57664

Add new settings in the kafka externals stream to set the schema name for the message key, then decode the message key as a JSON string (otherwise the type will be too complex)

**Additional context**

burnerlee:
PR checklist:
- [x] Did you run ClangFormat? — new code follows surrounding style; all edited files are in Proton-specific paths
- [x] Did you separate headers to a different section in existing community code base? — new includes added in alphabetical order within existing include blocks
- [x] Did you surround `proton: starts/ends` for new code in existing community code base? — not applicable, all changes are in `src/Storages/ExternalStream/Kafka/` which is Proton-specific code

Adds a new `message_key_schema_name` setting to Kafka external streams. When set, Proton decodes the Kafka message key using the Confluent Avro wire format (5-byte wire prefix: magic byte + 4-byte schema ID) and emits the decoded value as a JSON string into `_tp_message_key`.

**Motivation**: Producers that use Confluent Schema Registry commonly Avro-encode message keys, not just values. Reading `_tp_message_key` on such topics previously returned raw bytes.

**Behaviour**:
- `message_key_schema_name` names an Avro schema subject in the schema registry
- Requires `kafka_schema_registry_url` (validated at `CREATE EXTERNAL STREAM` time)
- `_tp_message_key` column must be `String` or `Nullable(String)` in Avro key mode (validated at `CREATE` time)
- Schema lookup is cached via the existing `KafkaSchemaRegistryForAvro` mechanism
- On read: Avro binary key is decoded into a `GenericDatum`, then re-encoded as Avro JSON (not plain JSON — union types use Avro union encoding, bytes fields are base64). The result is stored as a string in `_tp_message_key`
- On write: throws `NOT_IMPLEMENTED` if `message_key_schema_name` is set — Avro-encoding keys on write is not yet supported. Plain string keys via `_tp_message_key` without `message_key_schema_name` continue to work
- Empty key (`key_len == 0`) returns `NULL` for `Nullable(String)` columns, empty string for `String` columns

**Example**:

```sql
CREATE EXTERNAL STREAM orders (
    _tp_message_key String,
    payload         String
) SETTINGS
    type                       = 'kafka',
    brokers                    = 'localhost:9092',
    topic                      = 'orders',
    data_format                = 'JSONEachRow',
    kafka_schema_registry_url  = 'http://localhost:8081',
    message_key_schema_name    = 'orders-key';

SELECT _tp_message_key FROM orders;
-- Returns Avro JSON string, e.g. {"user_id": 42, "tenant": "acme"}
```

Closes #915


src/Storages/ExternalStream/Kafka/KafkaSource.cpp
```diff
@@ -641,11 +667,23 @@ void KafkaSource::getPhysicalHeader()
...
-                        if (inside_nullable && kmessage->key_len == 0)
-                            return Null{};
-                        return {static_cast<char *>(kmessage->key), kmessage->key_len};
-                    };
+                    if (avro_key_schema_registry)
+                    {
+                        virtual_col_value_functions[pos] = [this, inside_nullable](const rd_kafka_message_t * kmessage) -> Field
+                        {
+                            if (inside_nullable && kmessage->key_len == 0)
+                                return Null{};
```
yuzifeng1984: Use empty string instead of decode when key_len == 0 and type is not nullable.

src/Storages/ExternalStream/Kafka/KafkaSource.h
```diff
@@ -98,6 +104,9 @@ class KafkaSource final : public Streaming::ISource, public ExternalStreamSource
     bool ignore_format_errors = false;
     bool request_virtual_columns = false;
 
+    std::shared_ptr<KafkaSchemaRegistryForAvro> avro_key_schema_registry;
+    String avro_key_schema_subject;
```
yuzifeng1984: schema name is not needed in read as schema id is from the key string

src/Storages/ExternalStream/Kafka/KafkaSource.cpp
```diff
@@ -436,6 +445,23 @@ void KafkaSource::parseFormat(const rd_kafka_message_t * kmessage)
...
 
+Field KafkaSource::decodeAvroKey(const rd_kafka_message_t * kmessage) const
+{
+    ReadBufferFromMemory key_buf(static_cast<const char *>(kmessage->key), kmessage->key_len);
+    UInt32 schema_id = KafkaSchemaRegistry::readSchemaId(key_buf);
+    auto schema = avro_key_schema_registry->getSchema(schema_id);
+    auto avro_stream = std::make_unique<AvroInputStreamReadBufferAdapter>(key_buf);
+    auto avro_decoder = avro::binaryDecoder();
+    avro_decoder->init(*avro_stream);
+    Block key_header{ColumnWithTypeAndName(std::make_shared<DataTypeString>(), "_avro_key")};
```
yuzifeng1984: Can you explain how Kafka key (user defined Avro schema) is mapped to Proton string columns? 

Key column content might be a JSON string of decoded Kafka key.

yuzifeng1984:
Need implement KafkaSink related things to encode message key or throw an exception on write if currently not supported.

src/Storages/ExternalStream/Kafka/Kafka.cpp
```diff
@@ -21,6 +21,8 @@
 #include <Storages/parseShards.h>
 #include <Common/ProtonCommon.h>
 #include <Common/logger_useful.h>
+#include "Formats/FormatSettings.h"
```
yuzifeng1984: pls follow the include style: angle brackets & with the order of proton header, 3rd libs, std libs

burnerlee:
@yuzifeng1984 how can I test my changes in local - can you share your testing setup?

yuzifeng1984:
Hi @burnerlee, I use Redpanda for local testing. You may use its web UI to configure Kafka things including schema 
https://docs.redpanda.com/labs/docker-compose/single-broker/

burnerlee:
> Hi @burnerlee, I use Redpanda for local testing. You may use its web UI to configure Kafka things including schema 
> 
> https://docs.redpanda.com/labs/docker-compose/single-broker/

Thanks

yuzifeng1984:
@burnerlee The PR looks good. 

Can you also update the the kafka_schema_registry_url setting? Just in case it is only used for key but not message body. 

>  Blocker — Format validation conflict makes the documented use case impossible

>  File: src/Storages/ExternalStream/Kafka/Kafka.cpp:264-279 (verifySettings)

 > The pre-existing check unconditionally rejects any kafka_schema_registry_url value when data_format is not 'Avro' or 'ProtobufSingle':

  > if (!new_settings->kafka_schema_registry_url.value.empty())
  > {
  >    const auto & format = new_settings->data_format.value;
   >   const bool format_supported = format == "ProtobufSingle" || format == "Avro";
   >   if (!format_supported)
   >       throw Exception(...);
  >}

  > The PR's own example in the description uses data_format = 'JSONEachRow' with kafka_schema_registry_url set — exactly the case where a user wants Avro-encoded keys but JSON-encoded values. This CREATE EXTERNAL STREAM would throw at creation time, before reaching any of the new code.

> Fix: Condition the format check on whether the registry is used for value decoding. When message_key_schema_name is non-empty and data_format is not itself Avro/Protobuf, the registry is used solely for key decoding and the format
  constraint should not apply. For example:

 > const bool using_registry_for_value = new_settings->message_key_schema_name.value.empty();
  if (!new_settings->kafka_schema_registry_url.value.empty() && using_registry_for_value)
  {
      // existing format check ...
  }

burnerlee:
> @burnerlee The PR looks good. 
> 
> 
> 
> Can you also update the the kafka_schema_registry_url setting? Just in case it is only used for key but not message body. 
> 
> 
> 
> >  Blocker — Format validation conflict makes the documented use case impossible
> 
> 
> 
> >  File: src/Storages/ExternalStream/Kafka/Kafka.cpp:264-279 (verifySettings)
> 
> 
> 
>  > The pre-existing check unconditionally rejects any kafka_schema_registry_url value when data_format is not 'Avro' or 'ProtobufSingle':
> 
> 
> 
>   > if (!new_settings->kafka_schema_registry_url.value.empty())
> 
>   > {
> 
>   >    const auto & format = new_settings->data_format.value;
> 
>    >   const bool format_supported = format == "ProtobufSingle" || format == "Avro";
> 
>    >   if (!format_supported)
> 
>    >       throw Exception(...);
> 
>   >}
> 
> 
> 
>   > The PR's own example in the description uses data_format = 'JSONEachRow' with kafka_schema_registry_url set — exactly the case where a user wants Avro-encoded keys but JSON-encoded values. This CREATE EXTERNAL STREAM would throw at creation time, before reaching any of the new code.
> 
> 
> 
> > Fix: Condition the format check on whether the registry is used for value decoding. When message_key_schema_name is non-empty and data_format is not itself Avro/Protobuf, the registry is used solely for key decoding and the format
> 
>   constraint should not apply. For example:
> 
> 
> 
>  > const bool using_registry_for_value = new_settings->message_key_schema_name.value.empty();
> 
>   if (!new_settings->kafka_schema_registry_url.value.empty() && using_registry_for_value)
> 
>   {
> 
>       // existing format check ...
> 
>   }

Sure let me update

burnerlee:
@yuzifeng1984 updated. Could you check?

Issue #1221, opened 2026-09-08
Iceberg sink cannot write to AWS S3 Tables: four defects (Keep-Alive header, duplicate Host on commit, // in metadata paths, missing Parquet field-ids)


timotejkralik-oddin:
## Summary

The Iceberg database engine (`CREATE DATABASE ... SETTINGS type='iceberg'`, #928) cannot write to **AWS S3 Tables**. Four independent defects; each was isolated, root-caused in source, and its fix validated by an external bypass. With all four bypassed, the full pipeline (Kafka external stream → `ProtobufSingle` decode → `INSERT` → Iceberg commit) works end to end and the table is readable by PyIceberg and DuckDB.

Tested on `timeplus/proton:3.0.29` (both `latest` and `develop` tags resolved to 3.0.29, aarch64) against the S3 Tables REST catalog `https://s3tables.eu-west-1.amazonaws.com/iceberg`, table pre-created with PyIceberg. Source references are `develop` as of 2026-09-08.

| # | Defect | Effect | Fix size |
|---|--------|--------|----------|
| 1 | `Keep-Alive` request header on S3 data-plane PUTs | every `INSERT` fails on the first Parquet upload | small, port of an upstream ClickHouse commit |
| 2 | catalog `commitTable` POST sends the `Host` header twice | commit fails with HTTP 400 after all data files are written | 1 word |
| 3 | `//` written into committed metadata paths | committed table unreadable by PyArrow and DuckDB | 1 line |
| 4 | Parquet files carry no Iceberg field-ids | PyIceberg refuses the table, DuckDB reads every column as NULL | moderate |

What already works and is worth keeping: SigV4 against the REST catalog (`rest_catalog_sigv4_enabled`, `rest_catalog_signing_region`, `rest_catalog_signing_name`) authenticated every read first try, and the protobuf decode of deeply nested `map<uint64, Message>` payloads was the best of every engine we compared.

## Reproduction

```sql
CREATE DATABASE lake
SETTINGS type='iceberg',
  warehouse='arn:aws:s3tables:eu-west-1:<account>:bucket/<table-bucket>',
  catalog_type='rest',
  catalog_uri='https://s3tables.eu-west-1.amazonaws.com/iceberg',
  rest_catalog_sigv4_enabled=true,
  rest_catalog_signing_region='eu-west-1',
  rest_catalog_signing_name='s3tables';
-- the database name must equal an existing catalog namespace

CREATE EXTERNAL STREAM src (...)
[...]

PR #1225, opened 2026-09-08
fix(iceberg): write Iceberg field ids into Parquet data files

timotejkralik-oddin:
> **Updated 2026-09-11** — rewritten for the current head (`c94387e2`): the change now ports ClickHouse's own field-id code instead of a Proton-specific variant, and it has been built and verified end to end. The previous description applied to the original commits only.

PR checklist:
- Did you run ClangFormat? Yes — on the changed lines of every touched file (the community files are not clang-format clean at HEAD, so whole-file runs would rewrite unrelated code). The one formatting change not taken: in `IcebergSink.cpp` clang-format wanted to re-sort a pre-existing include, which is left as it was.
- Did you separate headers to a different section in existing community code base? Yes — `<DataTypes/NestedUtils.h>` in `PrepareForWrite.cpp`, `<optional>`/`<unordered_map>` in `Write.h` and `FormatSettings.h`, each in their own `/// proton: starts` … `/// proton: ends` section.
- Did you surround `proton: starts/ends` for new code in existing community code base? Yes — new blocks in `Parquet/{PrepareForWrite.cpp,Write.h}` and `FormatSettings.h`; modified lines carry `/// proton: updates`. Files under `src/Storages/Iceberg` and `src/Storages/ExternalStream/Iceberg` are Proton's own and unmarked.

Part 4 of 4 for #1221 (defect 4).

**What users saw.** Data written through the Iceberg sink cannot be resolved by readers that follow the Iceberg spec: the Parquet files carry no `field_id` on any schema element, and Iceberg readers resolve columns by id, not by name. PyIceberg 0.12 refuses the file (`Parquet file does not have field-ids …`); DuckDB, Trino and Daft return the right row count with **every column NULL**, silently. Reproduced on `timeplus/proton:3.0.30` against AWS S3 Tables; the defect is catalog-independent.

**Root cause.** `IcebergSink::consume()` creates the Parquet writer with plain format settings, so the encoder never sees the Iceberg schema the sink holds as `metadata.getSchemaJSON()`. The custom encoder's `Parquet::convertSchema()` builds the footer `SchemaElement`s and never calls `__set_field_id()`.

**What the change does** — a port of ClickHouse's Iceberg writer, adapted to Proton's plumbing:
- `IcebergSchemaProcessor::traverseSchema()` (`src/Storages/Iceberg/SchemaProcessor.{h,cpp}`, same as upstream) walks the table's Iceberg schema JSON into `dotted path → field id` for every field, nested included: struct members as `t.x`, list elements as `t.element`, map entries as `m.key`/`m.value`.
- `FormatSettings::parquet.field_ids` carries that map to `ParquetBlockOutputFormat`, which hands it to `Parquet::convertSchema()`.
- `PrepareForWrite.cpp` gets upstream's helpers, most lines verbatim: `buildSubFieldIds()` narrows the map to a node's children at each level, `lookupLeafFieldId()` resolves leaves, `validateIcebergFieldIds()` checks every logical field of the block has an id and throws `INCORRECT_DATA` otherwise, so a schema mismatch fails the INSERT loudly. Ids land on leaves and on list/map/struct groups; the Parquet-only `list` and `key_value` wrapper groups stay id-less — for maps, key and value are prepared directly rather than through `prepareColumnTuple("key_value")`, as upstream does.
- Without a map the encoder's output is unchanged (verified byte-identical, below), and the new parameters are defaulted, so existing call sites are untouched.
- The sink computes the map when it opens each new data file (so a schema refreshed after a commit conflict is honoured) and pins `use_custom_encoder = true`, since only the custom encoder writes the ids.
- `unit_tests_dbms` now links `ch_contrib::parquet` (`src/CMakeLists.txt`), which the gtest needs for the Thrift schema types.

**Relation to upstream.** Producer and encoder are ClickHouse master's code (`SchemaProcessor.cpp` `traverseSchema`; `PrepareForWrite.cpp` `buildSubFieldIds` / `lookupLeafFieldId` / `validateIcebergFieldIds`), so future merges from ClickHouse line up textually. The one deliberate difference is the carrier: upstream delivers the map through `FormatFilterInfo::column_mapper`. Proton's `FormatFactory::OutputCreator` takes three arguments, so carrying `FormatFilterInfo` the upstream way means changing all 43 `registerOutputFormat` lambdas across 36 files, and `FormatFilterInfo` is reader-side machinery (`ActionsDAG`, `KeyCondition`, `PrewhereInfo`) of which the writer uses only `column_mapper`. `FormatSettings::parquet.field_ids` (10 lines) carries the same `unordered_map<String, Int64>`; if Proton ports the Iceberg reader later, `ColumnMapper` slots in as a drop-in. Upstream's `IcebergOptionality` (REQUIRED vs OPTIONAL for nested containers) is a separate concern and not ported.

**Verification** — built and run, not just reviewed:
- Full Release build (arm64, `proton-binary-builder:clang-21`, CI cmake flags) of `proton` and `unit_tests_dbms` at this head's code.
- `unit_tests_dbms --gtest_filter=IcebergParquetFieldIds*`: 4/4 pass (`SchemaJSONToPaths`, `ParquetSchemaCarriesIds`, `NoIdsWithoutIcebergSchema`, `MissingIdIsAnError`).
- No regression for non-Iceberg Parquet: the same query (Map, Array, Tuple, Nullable, LowCardinality, array-of-tuple, map-of-array, map-of-tuple; `output_format_parquet_use_custom_encoder=1`) on `timeplus/proton:3.0.30` and on this build produces **byte-identical files** (`cmp`); `PrepareForWrite.cpp` is unchanged between v3.0.30 and `develop`, so 3.0.30 is a valid baseline.
- End to end against AWS S3 Tables: `INSERT` from this build into a table with long, string, `list<string>`, `map<string,int>`, `struct<x,y>` and `list<struct<ts>>`; PyIceberg 0.12 and DuckDB (`ENDPOINT_TYPE s3_tables`) both read the rows back correctly; all 16 `SchemaElement`s carry the id the table schema assigns, none on the `list`/`key_value` wrappers; #1222–#1224 hold in the same run (no `S3TablesUnsupportedHeader`, no duplicate `Host`, clean object keys).
- Not run here: the x86-64 build and the full test suite — ProtonCI has not been triggered on this fork PR yet.

**Workaround for files already written.** Existing data files stay id-less; readers can resolve them by name if the table property `schema.name-mapping.default` is set from the table schema (PyIceberg: `table.transaction().set_properties({"schema.name-mapping.default": table.name_mapping().model_dump_json()}).commit_transaction()`). Files written after this fix need no mapping.

**Known limitations, not addressed here.** Proton writes `REQUIRED` for outer list/map/struct groups where the spec allows `OPTIONAL`; PyIceberg and DuckDB accept the stricter file. Unnamed tuples produce paths like `point.1` that no Iceberg struct has and fail validation. Separately observed on the same files: `Date`/`DateTime` are written as `UINT_16`/`UINT_32` rather than Parquet logical types, and the sink's default `lz4` pages are unreadable by Trino 483 while `zstd` is fine.

### DIFF
--- a/base/poco/Net/include/Poco/Net/HTTPRequest.h
+++ b/base/poco/Net/include/Poco/Net/HTTPRequest.h
@@ -18,130 +18,128 @@
 #define Net_HTTPRequest_INCLUDED
 
 
-#include "Poco/Net/Net.h"
 #include "Poco/Net/HTTPMessage.h"
+#include "Poco/Net/Net.h"
 
 
-namespace Poco {
-namespace Net {
-
-
-class Net_API HTTPRequest: public HTTPMessage
-	/// This class encapsulates an HTTP request
-	/// message.
-	///
-	/// In addition to the properties common to
-	/// all HTTP messages, a HTTP request has
-	/// a method (e.g. GET, HEAD, POST, etc.) and
-	/// a request URI.
+namespace Poco
 {
-public:
-	HTTPRequest();
-		/// Creates a GET / HTTP/1.0 HTTP request.
-		
-	HTTPRequest(const std::string& version);
-		/// Creates a GET / HTTP/1.x request with
-		/// the given version (HTTP/1.0 or HTTP/1.1).
-		
-	HTTPRequest(const std::string& method, const std::string& uri);
-		/// Creates a HTTP/1.0 request with the given method and URI.
+namespace Net
+{
 
-	HTTPRequest(const std::string& method, const std::string& uri, const std::string& version);
-		/// Creates a HTTP request with the given method, URI and version.
 
-	virtual ~HTTPRequest();
-		/// Destroys the HTTPRequest.
+    class Net_API HTTPRequest : public HTTPMessage
+    /// This class encapsulates an HTTP request
+    /// message.
+    ///
+    /// In addition to the properties common to
+    /// all HTTP messages, a HTTP request has
+    /// a method (e.g. GET, HEAD, POST, etc.) and
+    /// a request URI.
+    {
+    public:
+        HTTPRequest();
+        /// Creates a GET / HTTP/1.0 HTTP request.
 
-	void setMethod(const std::string& method);
-		/// Sets the method.
+        HTTPRequest(const std::string & version);
+        /// Creates a GET / HTTP/1.x request with
+        /// the given version (HTTP/1.0 or HTTP/1.1).
 
-	const std::string& getMethod() const;
-		/// Returns the method.
+        HTTPRequest(const std::string & method, const std::string & uri);
+        /// Creates a HTTP/1.0 request with the given method and URI.
 
-	void setURI(const std::string& uri);
-		/// Sets the request URI.
-		
-	const std::string& getURI() const;
-		/// Returns the request URI.
-		
-	void setHost(const std::string& host);
-		/// Sets the value of the Host header field.
-		
-	void setHost(const std::string& host, Poco::UInt16 port);
-		/// Sets the value of the Host header field.
-		///
-		/// If the given port number is a non-standard
-		/// port number (other than 80 or 443), it is
-		/// included in the Host header field.
-		
-	const std::string& getHost() const;
-		/// Returns the value of the Host header field.
-		///
-		/// Throws a NotFoundException if the request
-		/// does not have a Host header field.
+        HTTPRequest(const std::string & method, const std::string & uri, const std::string & version);
+        /// Creates a HTTP request with the given method, URI and version.
 
-	void setCookies(const NameValueCollection& cookies);
-		/// Adds a Cookie header with the names and
-		/// values from cookies.
-		
-	void getCookies(NameValueCollection& cookies) const;
-		/// Fills cookies with the cookies extracted
-		/// from the Cookie headers in the request.
+        virtual ~HTTPRequest();
+        /// Destroys the HTTPRequest.
 
-	bool hasCredentials() const;
-		/// Returns true iff the request contains authentication
-		/// information in the form of an Authorization header.
-		
-	void getCredentials(std::string& scheme, std::string& authInfo) const;
-		/// Returns the authentication scheme and additional authentication
-		/// information contained in this request.
-		///
-		/// Throws a NotAuthenticatedException if no authentication information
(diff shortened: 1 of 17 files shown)


yuzifeng1984:
Hi @timotejkralik-oddin, thanks for try proton and submitting the fixes.

I am still doing some more testing for this change. Meanwhile, is it possible to include FormatFilterInfo from CH and have the field ids there? Better if we could leverage upstream codes and only make the necessary changes; thus reduce regression risk and the future maintenance/extension effort.

timotejkralik-oddin:
> Hi @timotejkralik-oddin, thanks for try proton and submitting the fixes.
> 
> I am still doing some more testing for this change. Meanwhile, is it possible to include FormatFilterInfo from CH and have the field ids there? Better if we could leverage upstream codes and only make the necessary changes; thus reduce regression risk and the future maintenance/extension effort.

Hello, thanks for solving my issue, I am glad it is moving so fast.

Thanks for suggestion I dig to it and converged to slightly better version.

I looked at doing it with FormatFilterInfo and the cost turned out to be the problem. Wiring it in the upstream way means adding a fourth argument to FormatFactory::OutputCreator, and since that's one shared std::function type, every registered output format has to change with it — 43 registerOutputFormat lambdas across 36 files, none of them related to Iceberg. Plus FormatFilterInfo brings reader-side machinery (ActionsDAG, KeyCondition, PrewhereInfo) into the write path, where the writer only ever uses its column_mapper. That's a lot more regression surface than the fix itself, which is the opposite of what we both want here.

So I went the other way and ported the upstream code itself: IcebergSchemaProcessor::traverseSchema for the schema walk and ClickHouse's field-id helpers in PrepareForWrite.cpp (buildSubFieldIds, lookupLeafFieldId, validateIcebergFieldIds), most of it line-for-line. The only Proton-specific piece left is carrying the id map through the existing FormatSettings::parquet — 10 lines. If Proton later ports the Iceberg reader, ColumnMapper slots in there as a drop-in.

I also fixed the unit test, which wasn't compiling before — the parquet headers weren't linked into unit_tests_dbms.

This time I built and tested whole my change on EC2 + S3 tables bucket and everything worked, test passed and Iceberg data was readable

Thanks

PR #1226, opened 2026-09-10
refine format

yuzifeng1984:
follow up #1221 and run CI test


yuzifeng1984:
The aws account is suspended; thus failed 30_udf_create_remote

PR #1227, opened 2026-09-16
use iceberg database storage_endpoint for s3 url

yuzifeng1984:
fixing `fail to write to iceberg`
 
```
2026.06.17 14:36:17.656115 <Error> default.mv_to_iceberg: Background runtime error:
  Code: 499. DB::Exception: Failed to check existence of key
  /tp_iceberg_test/data/00000-1-...-00001.parquet
  in bucket icebergtest: No response body. (S3_ERROR)
```


yuzifeng1984:
30_udf_create_remote failed because of aws lambda removed.

PR #1229, opened 2026-09-20
fix: roll back torn WAL tail on partial write under ENOSPC

yl-lisen:
## Problem

When the Raft WAL append path hits a write error mid-record (typically `ENOSPC` on a full disk), `AppendOnlyFile::append` returned `{written_bytes, errno}` with the partially written bytes already on disk. That leaves a torn record at the log tail, and on reboot recovery observes `committed_sn > next_log_sn` and crash-loops.

## Fix

Make a torn tail impossible by construction at the WAL level: on a write error after a partial write, the append truncates the bytes it just wrote so the file tail stays on a clean record boundary.

- `AppendOnlyFile::rollbackTornTail(partial_bytes, error_code)` is called from the error-return path of both `append` overloads (the `writev` batch and the `write` single-buffer one). It `ftruncate`s the partial tail and reports **0** durable bytes, which keeps `FileLogEntries`' cached size counter (`bytes += written.result`, applied unconditionally) consistent with the on-disk file.
- If the `ftruncate` (or the preceding `size()`) itself fails, it falls back to reporting `partial_bytes`, so the cached accounting still matches the un-truncated file; recovery trims the torn tail on restart.
- No happy-path cost: the `fstat` + `ftruncate` only run on the error path.

## Non-goals

- No disk-utilization pre-check or throttling on the append path — the truncate guarantee is sufficient and keeps the hot path free of extra syscalls.
- No change to the fail-stop (`LOG_PANIC`) semantics when the disk is full.


yl-lisen:
failed case `30_udf_create_remote` is not related

PR #1230, opened 2026-09-20
test: skip remote UDF cases whose endpoint is retired

yl-lisen:
## Why

The remote UDF cases call `https://<...>.lambda-url.us-west-2.on.aws/`, which no longer serves the function — it answers `403 {"message":"This account is suspended"}`. Every run of these cases fails on that response, so they report a broken endpoint rather than a product regression.

## What

Tag the four affected cases with `skip`, which is in probe's default blacklist (`bug,todo,skip`), so they stay in the tree but stop running:

- `tests/cluster/smoke/0022_udf2/30_udf_create_remote.yaml`
- `tests/cluster/compatibility/04_udf/06_udf_remote_prepare_data.yaml`
- `tests/cluster/compatibility/04_udf/07_udf_remote_basic_tests.yaml`
- `tests/cluster/compatibility/04_udf/08_udf_remote_extra_tests.yaml`

Each tag carries an inline note about why. Drop it once a replacement endpoint exists.

`tests/queries_ported/0_stateless/99013_remote_udf.sql` still covers the remote UDF DDL path and never calls the URL, so it is untouched.


yl-lisen:
no code changes, skipping ci

